PR #13, opened 2016-08-04
Shell out to pip instead of using the pip module

### LINKED ISSUE #7: Manage logging properly
When a function calls `subprocess.Popen()` the logging seems to reset to default and print to `stdout` and `stderror`. 

This is probably because logging hasn't been configured properly. The `opsdroid` object should probably handle this as it is accessible almost everywhere.

jacobtomlinson:
The logging does seem to be fine. The problem is actually in the pip module where it is messing with the root logger (pypa/pip#3043).

This PR stops using the `pip` module and instead uses `subprocess` to shell out to pip.

Closes #7

### DIFF
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -5,7 +5,6 @@ import os
 import shutil
 import subprocess
 import importlib
-import pip
 import yaml
 from opsdroid.const import (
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
Slight decrease in coverage due to me being unable to work out how to mock `subprocess` efficiently. The `dev` branch will still be up over `master` though.

PR #34, opened 2016-08-19
Test for class inheritance instead of class name

### LINKED ISSUE #33: Check for base classes instead of name when importing connectors and databases
Currently you must give your connector or database class a specific name for them to be imported.

Now that these classes should inherit from the base classes we can test for `isinstance(baseclass)` instead.

jacobtomlinson:
For connectors and databases.

Resolves #33


jacobtomlinson:
All connectors and databases need to be updated to work with this.

PR #46, opened 2016-08-31
Improved connector import debugging

jacobtomlinson:
Logs very verbosely


jacobtomlinson:
This might be too verbose

PR #47, opened 2016-09-02
Refactor the start connectors function to remove duplicate code

jacobtomlinson:


jacobtomlinson:
Coverage of `core.py` remains the same, it is just shorter. When coverage is less than 100% removing lines will always result in a decrease on average.

PR #107, opened 2017-02-16
Streamline json API response

jacobtomlinson:
Remove data which is already included in the headers.

- [x] Code
- [ ] Tests
- [x] Documentation


jacobtomlinson:
Decrease due to removal of one line.

PR #179, opened 2017-06-08
Add restart and stop methods to opsdroid.

### LINKED ISSUE #157: Reload skills
It should be possible for opsdroid to flush out all the loaded skills and other modules and reload them without restarting the application.

jacobtomlinson:
Allow opsdroid to be restarted without exiting.

Fixes #157.


jacobtomlinson:
Allow decrease in coverage as the lines are in `__main__.py`.

PR #213, opened 2017-09-05
Updated example_configuration.yaml

FabioRosado:
Added more configurations from the skills. 
Added mongo database settings.

Quick question:
If a skills has "None" in its README. Does it need to updated to the following?

```
skills:
  - name: (skillname)
```

If not, do we need to update the field on the example_configuration.yaml anyway for it?

### DIFF
--- a/opsdroid/configuration/example_configuration.yaml
+++ b/opsdroid/configuration/example_configuration.yaml
@@ -49,6 +49,10 @@ connectors:
 
 ## Database modules (optional)
 # databases:
+#   mongo:
+#     host:       "my host"     # (Optional) default "localhost"
+#     port:       "12345"       # (Optional) default "27017"
+#     database:   "mydatabase"  # (Optional) default "opsdroid"
 
 ## Skill modules
 skills:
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Thanks for this!

A couple of notes:

- Yes you're correct, you only need to specify the name if there are no config options.
- This config would result in all new users having all skills installed by default. This is probably not ideal so we should probably comment out most of these. But I like the idea of having lots of config examples in the file, users can just uncomment them to enable them.
- There a couple of typos (the vault skill [still says Devtools](https://github.com/FabioRosado/opsdroid/blob/c0ca01dc2ad877f214c119c153ccd5e7a13c0968/opsdroid/configuration/example_configuration.yaml#L105) in the comment, the [indentation is off](https://github.com/FabioRosado/opsdroid/blob/c0ca01dc2ad877f214c119c153ccd5e7a13c0968/opsdroid/configuration/example_configuration.yaml#L100) on the api.ai config, etc). I would appreciate if you could give it another pass.
- It would be awesome to autogenerate this when new skills are added. This would probably be best done by creating a scripts directory in the root of the project and add a python script in there to update the `example_configuration.yaml` file. This script could be called in the `Makefile`. This last one is probably out of scope for this PR so if you could [raise an issue](https://github.com/opsdroid/opsdroid/issues/new) for it that would be great.

jacobtomlinson:
Nice!

PR #220, opened 2017-09-09
Added more connectors and homeassistant skill

FabioRosado:
Added the twitter and facebook connectors into the example and the homeassistant skill


jacobtomlinson:
Nice thanks

PR #230, opened 2017-09-23
Allows to include a yaml file inside another

FabioRosado:
I've tested my configuration.yaml file with !include and everything seems to work fine.
I've also attempted to create a test file to see if the !include is working properly, since this was my first attempt, please let me know if this should be done differently.

### DIFF
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -199,4 +199,4 @@ skills: !include skills.yaml
 
 ```
 
-_Note: The file.yaml that you wish to include in the config must be located in the /.opsdroid folder_
+_Note: The file.yaml that you wish to include in the config must be in the same directory as your configuration.yaml (e.g ~/.opsdroid)_
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -188,3 +188,15 @@ skills:
 ```
 
 _Note: Your environment variable names must consist of uppercase characters and underscores only. The value must also be just the environment variable, you cannot currently mix env vars inside strings._
+
+## Include additional yaml files
+
+You can split the config into smaller modules by using the value `!include file.yaml` to import the contents of a yaml file into the main config. 
+
+
+```yaml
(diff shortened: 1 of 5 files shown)


jacobtomlinson:
Good work on this! And well done sticking with the tests!

Sadly it looks like your function isn't actually being tested. You can see in the [coverage](https://coveralls.io/builds/13407963/source?filename=opsdroid%2Floader.py#L150) that the lines are not executed. This means the test needs tweaking.

jacobtomlinson:
If the line is too long you can always break it onto a new line instead of creating a variable.

```python
path = "tests/configs/minimal_with_include.yaml"
config = loader.load_config_file(path)
```

```python
config = loader.load_config_file(
                   "tests/configs/minimal_with_include.yaml")
```

**This is just an example, the indentation may be wrong.**

jacobtomlinson:
Also I really like that you've created a config which simply includes the minimal config! It could be nice to have a test which loads both and ensures they are equal (which they chould be).

jacobtomlinson:
Yeah that test is nice! Weird that the coverage is still failing...

FabioRosado:
I'm giving it another go to see if the coverage will remain the same with the  linebreak on the loader.
I've also added a test to check if the config_with_include is the same as minimal one.

Any suggestion on how to test the:

```
154 | with open(included_yaml, 'r') as included: | !
155 | return yaml.load(included)
```

jacobtomlinson:
It must be running those lines in order for the test to pass! I'm happy to merge this despite coverage complaining.

Can you just add a section to the docs to show people how to use this.

FabioRosado:
Perhaps the coverage is failing because I'm not actually testing if the path is /.opsdroid or tests/configs/ so it complains that this bit is not good enough?
Yeah I'll update the docs and push the changes in a bit

jacobtomlinson:
No the issue is it's saying that it literally isn't executing the lines of code in that function. But it must be for the test to be equal.

Issue #215, opened 2017-09-06
Log some getting started guidance on first run


jacobtomlinson:
If a users starts opsdroid for the first time they will get a [default config file created](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/loader.py#L109) for them.

At the same time it might be nice to log some getting started information. Perhaps a pointer on how to install opsdroid desktop and a link to the getting started videos or the config reference.

FabioRosado:
Would you like this to be added in the **main** and to be logged every time opsdroid is started or only when a default config file is created?

PR #231, opened 2017-09-26
Added Getting Started Info (Issue #215)

FabioRosado:
Logged some important links and added a quick text about customising opsdroid instance in configuration.yaml. I didn't add too much text because the user probably doesn't want to be spammed with a wall of text.

Let me know if this is okay or if you'd rather have something else logged in the console.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,3 +23,10 @@ deploy:
     tags: true
     repo: opsdroid/opsdroid
     condition: "$TOXENV = py35"
+notifications:
+  webhooks:
+    urls:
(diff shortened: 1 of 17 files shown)


jacobtomlinson:
This is great thanks! 

I like the idea of showing this on every run, but it might be nice to allow people to hide it. Could you add an option to the config along the lines of `show_welcome_message` which defaults to true, but if it's set to false then it needs to be hidden.

This probably affects the place where you are logging this, as it will need to be done after the config is loaded.

jacobtomlinson:
Awesome work!

PR #239, opened 2017-09-30
Adding try/except to welcome_message

accessdenied12:
This will keep the key error from breaking the start process.

### DIFF
--- a/opsdroid/__main__.py
+++ b/opsdroid/__main__.py
@@ -102,7 +102,7 @@ def welcome_message(config):
             _LOGGER.info("Install Opsdroid Desktop at: "
                          "https://github.com/opsdroid/opsdroid-desktop/releases")
             _LOGGER.info("=" * 40)
-    except KeyError as e:
+    except KeyError:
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
Nice work thanks for this! Could you please add a test which follows the except block and asserts that it doesn't raise.

accessdenied12:
@jacobtomlinson - Thanks.  I've updated the unit test so that it checks to make sure nothing is sent back from the function if welcome-message is not present.  Let me know if you want me to change anything.

jacobtomlinson:
This is perfect. It looks like the linter is failing you because line 103 is longer than 80 characters. Can you either break it over two lines or add a pylint disable comment so it's ignored.

accessdenied12:
@jacobtomlinson - Sure thing.  Looks like the build is passing now in Travis 👍

jacobtomlinson:
Awesome! Ping me your address via DM on twitter or Gitter and I'll post you some opsdroid stickers!

PR #249, opened 2017-09-30
Fix #248: Update documentation in skills.md

### LINKED ISSUE #248: Update skills.md to reflect the changes in adding new skills
After going through the [`skills.md`](https://github.com/opsdroid/opsdroid/blob/master/docs/extending/skills.md) documentation file and comparing it with a sample skill ([skill-hello](https://github.com/opsdroid/skill-google-it/blob/master/__init__.py)), there is a couple of things that may be outdated:
1. `__init__.py` should exist for a skill instead of `<skill_name>.py`
2. `match_regex` now exists in `opsdroid.matchers` instead of `opsdroid.skills`

gtseres:
Fix #248. Some updates in the `skills.md` documentation to reflect the current status. 
@jacobtomlinson I will need some help with the parsers dead links, where should they point to?


jacobtomlinson:
Great thanks!

I think those lines should be going somewhere in the parsers directory within docs. Probably the [overview](https://github.com/opsdroid/opsdroid/blob/master/docs/parsers/overview.md). The parsers section should probably be renamed to matchers though.

jacobtomlinson:
A little history:

The decorators which a skill author uses to match the string from the chat client used to be called parsers. But there is an internal set of classes, also called parsers, which actually parses the string and matches it against the skills. So I decided to rename the decorators to matchers, as they are describing to the parsers how to match against a skill. 

However the docs have remained a little out of date.

jacobtomlinson:
Thanks for the further changes. I'm afraid the links you thought were broken actually aren't. They are for the [readthedocs](http://opsdroid.readthedocs.io/en/latest/?badge=latest) documentation rather than navigating around GitHub, so adding `.md` to them will actually break the main docs. Sorry!

PR #259, opened 2017-10-04
Minor tweak to note how to run opsdroid under pdb.

go8ose:
This is the first thing I wanted to do, and I couldn't see it in the documentation, so I thought I'd give you a tiny PR to add it.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -25,8 +25,9 @@ pip install -r requirements.txt
 # run opsdroid
 python -m opsdroid
 
-# or run opsdroid under the debugger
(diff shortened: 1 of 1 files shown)


go8ose:
Actually, I'm not sure this works.  I think pdb was running the opsdroid I'd installed with pip, rather than the github clone that I'd made.  I'll update later if I work this out.

jacobtomlinson:
Cool thanks. I'll close this then if you're not sure. Feel free to raise it again when you're happy with it.

go8ose:
@jacobtomlinson I did 617d516 to get this better.  Can this PR be re-opened and re-considered?

PR #267, opened 2017-10-14
adding Lex matcher/parser

slapula:
I am submitting this PR to satisfy #251.  This is still a bit of a work in progress at the moment while I continue testing.  I will update this PR when it is ready for final review.  Comments or feedback are welcome though at any point.

PR #263, opened 2017-10-06
Add LUIS parser/matcher

pohzipohzi:
fix for #252 

Some notes: 

I used the sample Calendar pre-built entity provided by microsoft, and tested for the action `Calendar.Add`. `LUIS` doesn't offer the same `speech` return as `api.ai` does (at least I don't think I can find it), so I tried to parse the message returned in a similar way as `api.ai`

example `__init__.py`
```
from opsdroid.matchers import match_luisai_intent

@match_luisai_intent('Calendar.Add')
async def passthrough(opsdroid, config, message):
    if message.luisai["topScoringIntent"]["score"]:
        await message.respond(str(message.luisai["topScoringIntent"]["score"]))

```

example configuration.yaml
```
##                      _           _     _
##   ___  _ __  ___  __| |_ __ ___ (_) __| |
##  / _ \| '_ \/ __|/ _` | '__/ _ \| |/ _` |
## | (_) | |_) \__ \ (_| | | | (_) | | (_| |
##  \___/| .__/|___/\__,_|_|  \___/|_|\__,_|
##       |_|
##                   __ _
##   ___ ___  _ __  / _(_) __ _
##  / __/ _ \| '_ \| |_| |/ _` |
## | (_| (_) | | | |  _| | (_| |
##  \___\___/|_| |_|_| |_|\__, |
##                        |___/
##
## A default config file to use with opsdroid

## Set the logging level
# logging:
#   level: info
#   path: opsdroid.log
#   console: true

## Set the location for opsdroid to install modules
# module-path: "."

## Show welcome message
welcome-message: true

## Configure the web server
# web:
#   host: '127.0.0.1'
#   port: 8080
#   ssl:
#     cert: /path/to/cert.pem
#     key: /path/to/key.pem

## Parsers
parsers:
#   - name: regex
#     enabled: true
#
#   - name: crontab
#     enabled: true
#
   - name: apiai
     access-token: "<access-token>"
     min-score: 0.6

   - name: luisai
     appid: "<application-id>"
     appkey: "<subscription-key>"
     verbose: True
     min-score: 0.6


## Connector modules
connectors:
  - name: shell
  - name: websocket

  # Uncomment the connector(s) that you wish opsdroid to work on
#  ## Slack (https://github.com/opsdroid/connector-slack)
#  - name: slack
#    # Required
#    api-token: "zyxw-abdcefghi-12345"
#    # optional
#    bot-name: "mybot"       # default "opsdroid"
#    default-room: "#random" # default "#general"
[...]

PR #268, opened 2017-10-16
Add witai parser

FabioRosado:
Still haven't attempted to create tests for it.

edit: The bad auth code that I was getting was because I placed the access-token key between  "", the access token needs to be inserted without them.
 
edit2: match_witai now works if imported in a skill

### DIFF
--- a/opsdroid/parsers/witai.py
+++ b/opsdroid/parsers/witai.py
@@ -1,0 +1,79 @@
+"""A helper function for parsing and executing wit.ai skills."""
+
+import logging
+import json
+
+import aiohttp
+
+_LOGGER = logging.getLogger(__name__)
+
+
+async def call_witai(message, config):
+    """Call the wit.ai api and return the response."""
+    async with aiohttp.ClientSession() as session:
+        headers = {
+            "Authorization": "Bearer " + config['access-token']
+        }
+        payload = {
+            "v": "20170307",
+            "q": message.text
+        }
+        resp = await session.get("https://api.wit.ai/message?v={}&q={}".format(
+            payload['v'], payload['q']), headers=headers)
+        result = await resp.json()
+        _LOGGER.debug("wit.ai response - " + json.dumps(result))
+        _LOGGER.info(result)
+        return result
+
+
+async def parse_witai(opsdroid, message, config):
+    """Parse a message against all witai skills."""
+    # pylint: disable=broad-except
+    # We want to catch all exceptions coming from a skill module and not
+    # halt the application. If a skill throws an exception it just doesn't
+    # give a response to the user, so an error response should be given.
+    if 'access-token' in config:
+        try:
+            result = await call_witai(message, config)
+        except aiohttp.ClientOSError:
+            _LOGGER.error("No response from wit.ai, check your network.")
+            return
+
+        if 'code' in result:
+            _LOGGER.error("wit.ai error - " + str(result['code'])
+                          + " " + str(result['error']))
+            return
+        elif result['entities'] == {}:
+            _LOGGER.error("wit.ai error - No intent found. Did you "
+                          "forget to create one?")
+            return
+
+        try:
+            confidence = result['entities']['intent'][0]['confidence']
+        except KeyError:
+            confidence = 0.0
+        if "min-score" in config and confidence < config['min-score']:
+            _LOGGER.debug("wit.ai score lower than min-score")
+            return
+
+        if result:
+            for skill in opsdroid.skills:
+                if "witai_intent" in skill:
+                    if (skill['witai_intent'] in
+                            [i['value'] for i in
+                             result['entities']['intent']]):
+                        message.witai = result
+                        try:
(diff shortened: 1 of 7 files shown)


jacobtomlinson:
Nice! Have a look at #263 and #267 which add LUIS and Lex support respectively. You should see what they've done and hopefully anything you've missed.

With the matcher issue is that because you named the decorator `match_witai_intent`?

FabioRosado:
Should I name `match_witai_intent` differently?

jacobtomlinson:
It's up to you really. The only reason the apiai one is called `match_apiai_intent` is because there is also `match_apiai_action`, if that didn't exist I would probably have just called it `match_apiai`.

I'm happy either way, just ensure you are consistent and that it is documented.

opsdroid/parsers/witai.py
```diff
@@ -0,0 +1,68 @@
...
+                                                 message)
+                        except Exception:
+                            await message.respond(
+                                "Whoops there has been an error")
+                            await message.respond(
+                                "Check the log for details")
+                            _LOGGER.exception("Exception when parsing '" +
+                                              message.text +
+                                              "' against skill '" +
+                                              result + "'")
```
FabioRosado: I'm writing the tests and for this bit to work I have to do one of the following:
`result['entities']['intent'][0]['value']` which makes the line too long or simply do `str(result)` and pass the whole json received, which one would you rather use?

jacobtomlinson: I would say the first probably.

FabioRosado:
I changed the name of the matcher as suggested, I've also added the tests for wit.ai based on api.ai ones. I've added the documentation for wit.ai but I want to play around with `message.witai` and check how to use a skill other information returned by witai.

I've also tested the wit.ai matcher/parser and it seems to work nicely with the weather skill so I guess it should work fine with other skills as well.

I've finally understood why one of the docstring tests where failing, sorry for the testing spam. 
Let me know what you think and if I need to change anything.

jacobtomlinson:
This is really good work!

Here's a few points to get the last couple of steps so it can be merged:

- Travis is failing because the linter is complaining about a broad exception. Check out the lint exception comments in the [other parsers](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/parsers/luisai.py#L32).
- Can you throw in a test for a response with no entities to cover [43-45](https://coveralls.io/builds/13739572/source?filename=opsdroid%2Fparsers%2Fwitai.py#L43) in `witai.py`. The other missed lines are fine, they are missing in the other parsers too.
- Can you update the example in the docs to use an entity like a location from the response. For example you could ask wit.ai "what's the weather in London" and then get the location entity `London` and then use it in your example skill.

FabioRosado:
I've followed your suggestions and updated the commit.
Let me know if the docs make sense or if they are too confusing. I'm going to start reading about tests to try and figure out how to test for `ClientOSError`

jacobtomlinson:
Great I'll take a look at this tomorrow!

Why don't you have a look at those tests under a new PR. I'm happy to merge this one once I've tested it locally.

opsdroid/parsers/witai.py
```diff
@@ -0,0 +1,76 @@
...
+        if 'code' in result:
+            _LOGGER.error("wit.ai error - " + str(result['code'])
+                          + " " + str(result['error']))
+            return
+        elif result['entities'] == {}:
+            _LOGGER.error("wit.ai error - No intent found. Did you "
+                          "forget to create one?")
+            return
+
+        confidence = result['entities']['intent'][0]['confidence']
```
jacobtomlinson: If the string doesn't match an intent then there will be no intent paramater in the response resulting in an error. This needs to be caught.

```python
  File "/opt/boxen/homebrew/Cellar/python3/3.5.2_3/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "/Users/jacob/Projects/opsdroid/opsdroid/opsdroid/parsers/witai.py", line 51, in parse_witai
    confidence = result['entities']['intent'][0]['confidence']
KeyError: 'intent'
```

FabioRosado:
I've added a try/except on that piece of code that you mentioned @jacobtomlinson, I've also added a test for it. At the moment I'm setting confidence to 0.0 because if there is no intent then probably there  is no action to follow up.  Please let me know if this is okay or if you rather have it done in a different way.

I've also added the `test_parse_witai_clientOSError` to see if I could pass the ClientOSError, the coverage seem to increase 2%, does this means that this test worked? If so I'll update the other parsers to include it 👍

jacobtomlinson:
Yeah I like that! Annoyingly the linter is now complaining about some whitespace issues. Can you quickly tweak those.

FabioRosado:
Sorry for that I was giving it a go at the ClientOSError issue and messed up the spaces.
This attempt should pass the tests, I'll keep trying to play around with the ClientOSError to see if I can crack the issue 👍

jacobtomlinson:
Awesome good work!

PR #271, opened 2017-10-17
Handle missing section details in load_modules_from_config

go8ose:
If the user has a 'databases:' section in their configuration, but nothing
listed as part of it, the opsdroid.loader.load_modules_from_config() would
call _load_modules with the modules parameter set to None, which would cause
a problem when trying to iterate it on line 223.


jacobtomlinson:
Nice one thanks for this!

go8ose:
I've done 5eba9335 to add some tests. I found that my ca9858a change does fix the tests that I've added. However other tests in tests/test_loader.py are failing, and I don't know why.  Also, I'm developing on a macOS with conda, so I can't run tox to exercise all the tests.

jacobtomlinson:
The diffs show you've added three tests, two of which are failing. Is that right?

Potentially these tests aren't cleaning up after themselves and that needs correcting.

FabioRosado:
I tried to play around a bit and travis might be failing because on your minimal_config_file2 you added databases but didn't added names under it. 
When I tried to run opsdroid without the "name" param it all died.

I'd also like to point that you can refactor your tests into one, you can add the mininal_config_file and minimal_config_file2 on the same test by using a different var name and test if both are not none.

Hope this helps a bit

jacobtomlinson:
@FabioRosado This PR is adding support for an empty `databases` section, so it is expected to be empty in `minimal_config_file2.yaml`.

jacobtomlinson:
@go8ose any update on this?

go8ose:
Hi @jacobtomlinson.  I'm planning on having another go tomorrow morning.  My thoughts are to do some mocking/stubbing in my test so that I can still exercise the load_modules_from_config() function (in order to test it correctly parses the configuration), but doesn't actually step down into the _load_modules() code.

jacobtomlinson:
Ok great. Let me know when you are happy with it.

go8ose:
All the tests are passing, can you please merge it now?

jacobtomlinson:
Ah! I thought you wanted to do more work on it when you said:

> I'm planning on having another go tomorrow morning.

PR #275, opened 2017-10-19
Add tests for ClientOSError in apiai, witai and luisai parsers

FabioRosado:
Testing to see if using side_effect would cause the ClientOSError exception to increase the coveralls for parsers.


jacobtomlinson:
The failure is just because of linter issues (unused variables and your comment). However there is no increase in coverage so it hasn't covered the lines you're trying to test.

jacobtomlinson:
Awesome work!

Issue #224, opened 2017-09-13
Connector disconnect method


jacobtomlinson:
When opsdroid is exiting it should call a `disconnect` method on each connector. This would allow connectors to clean up before the application exits.

 - [x] Implement dummy method in the `BaseConnector`
 - [x] Call the method on each connector at exit

PR #276, opened 2017-10-19
Add a disconnect method

Cadair:
This is to address #224 

I couldn't work out the best way of calling the disconnect coroutine from the main class. My asyncio knowledge gets poor around calling coroutines from functions.

### DIFF
--- a/opsdroid/connector.py
+++ b/opsdroid/connector.py
@@ -82,3 +82,4 @@ class Connector():
             opsdroid (OpsDroid): An instance of the opsdroid core.
 
         """
+        pass
--- a/opsdroid/connector.py
+++ b/opsdroid/connector.py
@@ -71,3 +71,14 @@ class Connector():
 
         """
         raise NotImplementedError
+
+    async def disconnect(self, opsdroid):
+        """Disconnect from the chat service.
+
+        This method is called when opsdroid is exiting, it can be used to close
+        connections or do other cleanup.
+
+        Args:
+            opsdroid (OpsDroid): An instance of the opsdroid core.
+
+        """
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -39,7 +39,7 @@ class OpsDroid():
         self.connector_tasks = []
         self.eventloop = asyncio.get_event_loop()
         for sig in (signal.SIGINT, signal.SIGTERM):
-            self.eventloop.add_signal_handler(sig, self.stop)
+            self.eventloop.add_signal_handler(sig, self.call_stop)
         self.skills = []
         self.memory = Memory()
         self.loader = Loader(self)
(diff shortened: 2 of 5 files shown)


opsdroid/core.py
```diff
@@ -101,6 +101,9 @@ def restart(self):
 
     def stop(self):
         """Stop the event loop."""
+        for connector in self.connectors:
```
Cadair: I assume it needs to be in this method somewhere, maybe even like this. This however isn't actually made to run, because it's cancelled right after?

jacobtomlinson: Yes. I would gather those disconnect tasks and wait for them to complete.

jacobtomlinson:
Awesome thanks for this. I've answered your question. There is also a drop in coverage as you haven't added any tests for this.

opsdroid/core.py
Cadair: How should I wait for them to complete? Using `run_until_complete` didn't work, I got something along the lines of "loop is already running."

jacobtomlinson: I think as you're already in a coroutine you can simply await the gather, you don't need to run the loop again.

jacobtomlinson: To be honest you don't even need to create them on the loop. Just wait them.

Change

```python
self.eventloop.create_task(connector.disconnect(self))`
```

to

```python
await connector.disconnect(self)
```

jacobtomlinson: Good point, perhaps then the `disconnect` method shouldn't be a coroutine.

Cadair: Certainly the use case I have for it, I would rather it be a coroutine, as the disconnect call is async like the rest of the lib. I could probably work around it though.

jacobtomlinson: Yep that's fair. Perhaps then the `stop` method should be a coroutine.

jacobtomlinson:
Hey. Have you had a chance to take a look at this again?

Cadair:
This seems to work ok for `SIGINT` on my local machine. I am getting some interesting errors if initialisation fails for any reason though.

for instance this:
```
CRITICAL opsdroid.core: All connectors failed to load
INFO opsdroid.core: Exiting application with return code 1
DEBUG asyncio: Using selector: EpollSelector
Error: All connectors failed to load
  File "/opt/miniconda/envs/opsdroid/lib/python3.6/asyncio/base_events.py", line 512, in __del__
    self.close()
  File "/opt/miniconda/envs/opsdroid/lib/python3.6/asyncio/unix_events.py", line 65, in close
    self.remove_signal_handler(sig)
  File "/opt/miniconda/envs/opsdroid/lib/python3.6/asyncio/unix_events.py", line 146, in remove_signal_handler
    signal.signal(sig, handler)
  File "/opt/miniconda/envs/opsdroid/lib/python3.6/signal.py", line 47, in signal
    handler = _signal.signal(_enum_to_int(signalnum), _enum_to_int(handler))
```

Cadair:
I will add tests to this once it's working properly :)

jacobtomlinson:
Interesting about the other exception. Could you raise a separate issue for this?

Cadair:
I think that this change has introduced it. I need to investigate more.

Cadair:
I have added tests and some brief documentation, and also opened #298

opsdroid/connector.py
```diff
@@ -71,3 +71,14 @@ def __init__(self, config):
 
         """
         raise NotImplementedError
+
+    async def disconnect(self, opsdroid):
```
jacobtomlinson: Should this method have a `pass` after the docstring as it (rightly) doesn't `raise`?

Cadair: I can add one to be explicit, but it's not required (just the docstring is sufficient).

jacobtomlinson: Explicit is good as it makes the code more readable.

Cadair: if we were being really explicit about it, it should be `return None` 😉

jacobtomlinson: True! I'll leave it up to you. I'm aiming for readability.

Cadair:
Thanks :)

Issue #284, opened 2017-10-20
Add Issue and PR templates


jacobtomlinson:
It would be nice to have templates for issues and PRs.

Issues would generally require the following information:
- opsdroid version
- python version
- os version (or docker version)
- Configuration file
- Expected functionality
- Experienced functionality (including debug log)

PRs should require the following info:
- Issue number related to
- Description of what the PR adds
- Checkboxes for code, tests and docs

PR #285, opened 2017-10-20
Add issue/PR template - issue #284

FabioRosado:
Decided to give this a try I used bits of other templates I found online to create these, let me know if you are okay with them or you'd rather have something else on them.

### DIFF
--- a/docs/ISSUE_TEMPLATE.md
+++ b/docs/ISSUE_TEMPLATE.md
@@ -1,0 +1,32 @@
+# Description
+Please include a summary of the issue. 
+
+
+## Steps to Reproduce
+Please also include relevant information and steps to reproduce the bug/issue.
+
+
+## Expected Functionality
+Explain what should happen.
+
+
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
Wow! I did not expect that so fast, good work!

I really like what you've done, the only change I would suggest is removing the example config as it's rather large and folks will just be deleting it and adding their own anyway. You could just replace it with a code block which says `your config goes here`.

jacobtomlinson:
Looks like something has gone wrong with your push here. The diff has now changed to show the #275 commits.

FabioRosado:
Sorry it took me a while to fix the issue because I messed up a pull/rebase and lost all the work had to reflog the changes.

Today I've been doing a lot of waiting around because i'm on call for work so I've seen the issues and thought this one I could come up with something quickly haha

jacobtomlinson:
Awesome stuff!

Issue #288, opened 2017-10-21
Warning in tests `coroutine 'parse_crontab' was never awaited`


jacobtomlinson:
# Description
When running tests they complain about `coroutine 'parse_crontab' was never awaited`.

## Steps to Reproduce
Run `tox` and check the warnings.

## Expected Functionality
There should be no warnings

## Experienced Functionality
```
====================================== warnings summary =======================================
tests/test_core.py::TestCoreAsync::test_parse_luisai
  /Users/jacob/Projects/opsdroid/opsdroid/.tox/py36/lib/python3.6/_collections_abc.py:660: RuntimeWarning: coroutine 'parse_crontab' was never awaited
    return self[key]

-- Docs: http://doc.pytest.org/en/latest/warnings.html
```

## Versions
- **Opsdroid version:** v0.9.1-25-g177449b
- **Python version:** 3.5.3
- **OS/Docker version:** macOS 10.13

PR #291, opened 2017-10-21
Fix warning in tests

freedomofkeima:
Hi  there :)

# Description

When running tests they complain about `coroutine 'parse_crontab' was never awaited.` The root cause is that `asyncio.create_task` for `parse_crontab` is not isolated properly. After this issue is solved in https://github.com/freedomofkeima/opsdroid/commit/d4a7d023f67a523e55e09e7095879a63d6d56f94 + https://github.com/freedomofkeima/opsdroid/commit/c4ece48aeb92eca61016bdc79a257e504b969a77, there's another new occurring problem.

```
========================== 113 passed in 2.50 seconds ==========================
 File "/opt/python/3.6.3/lib/python3.6/asyncio/base_events.py", line 512, in __del__
 self.close()
 File "/opt/python/3.6.3/lib/python3.6/asyncio/unix_events.py", line 65, in close
 self.remove_signal_handler(sig)
 File "/opt/python/3.6.3/lib/python3.6/asyncio/unix_events.py", line 146, in remove_signal_handler
 signal.signal(sig, handler)
 File "/opt/python/3.6.3/lib/python3.6/signal.py", line 47, in signal
 handler = _signal.signal(_enum_to_int(signalnum), _enum_to_int(handler))
___________________________________ summary ____________________________________
```

Future investigations (see https://github.com/python/asyncio/issues/396, https://bugs.python.org/issue23548) shows that this problem occurred because of a bug in signal handling and we need to call `close` explicitly. `TestCore` class does a lot of manipulation against `self.eventloop`, which should be fixed in https://github.com/freedomofkeima/opsdroid/commit/b65f8bf271f2434c31eb1286438c2cca65b4ec2f.

**Fixes:** #288

## Status
**READY**

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Run the test at [Travis CI](https://travis-ci.org/freedomofkeima/opsdroid/builds/290881477) and ensure there's no warning given there.

# Checklist:

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -50,7 +50,8 @@ class TestCore(unittest.TestCase):
             opsdroid.load()
             self.assertTrue(opsdroid.loader.load_config_file.called)
 
-    def test_start_loop(self):
+    @asynctest.patch('opsdroid.core.parse_crontab')
+    def test_start_loop(self, mocked_parse_crontab):
         with OpsDroid() as opsdroid:
             mockconfig = {}, {}, {}
             opsdroid.web_server = mock.Mock()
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Nice work! Thanks for this.

PR #304, opened 2017-10-26
Replace regex parser - use re.search instead re.match

FabioRosado:
# Description
As @gabeduke mentioned in gitter, the regex parser is ignoring the regular expression to be searched if it's not at the beginning of the string. After a few testing and reading it seems that the reason why this is happening because re.match will only match a regular expression if ifound at the beginning of the string.

Using re.search seems to fix this issue as this method will search through the whole string for the regular expression specified on the skill.  So I've replaced `re.match` with `re.search` in opsdroid.parsers.regex

I've also updated the documentation to show the importance of using anchors when building skills, please let me know if the examples make sense of if they are too confusing.

I've also replaced line 94 of opsdroid.core: `print("Error: " + error)`. When spaces are not used properly in configuration.yaml the following exception is returned:

```python
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/core.py", line 94, in critical
    print("Error:" + error)
TypeError: must be str, not ParserError
```

By replacing concatenation with a comma the print statement works normally. I've noticed that we are currently logging this, do we still need this print?

**Fixes:** N/A

## Status

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [X] Documentation (fix or adds documentation)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [X] Tested both locally and Travis CI - All tests passed
- [X] Tested different skills to see if any side-effects occur - none experienced

# Checklist:

### DIFF
--- a/docs/matchers/regex.md
+++ b/docs/matchers/regex.md
@@ -2,7 +2,9 @@
 
 This is the simplest matcher available in opsdroid. It matches the message from the user against a regular expression. If the regex matches the function is called.
 
-## Example
+_note: The use of position anchors(`^` or `$`) are encouraged when using regex to match a function. This should prevent opsdroid to be triggered with every use of the matched regular expression_
+
+## Example 1
 
 ```python
 from opsdroid.matchers import match_regex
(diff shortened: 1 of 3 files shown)


docs/matchers/regex.md
```diff
@@ -14,6 +16,52 @@ async def hello(opsdroid, config, message):
...
+#### Fixed example
+```python
+from opsdroid.matchers import match_regex
+
+@match_regex('cold$')
+async def is_cold(opsdroid, config, message):
+    await message.respond('it is')
+```
+
+Now this skill will only be triggered if `'cold'` is located at the end of the message.
```
jacobtomlinson: This wouldn't actually work as the end of the message is `cold?` not `cold`. You either need to remove the question mark from the example or add the question mark to the regex. I'd probably remove the question mark.

opsdroid/core.py
```diff
@@ -92,7 +92,7 @@ def critical(self, error, code):
         """Exit due to unrecoverable error."""
         self.sys_status = code
         _LOGGER.critical(error)
-        print("Error: " + error)
+        print("Error:", error)
```
jacobtomlinson: Thanks for fixing this! I know it's only a tiny change but this probably should've been in a separate PR. I'll leave it for now, but something to bear in mind in the future.

FabioRosado:
Thanks for that Jacob I didn't think about the question mark when I wrote the example. I've removed it from the example so things should look okay now

jacobtomlinson:
Nice one!

Issue #306, opened 2017-10-29
Support for `apiai` in the config has broken


jacobtomlinson:
# Description
Since Api.ai was renamed to Dialogflow in #294 any config file with a parser with the name set to `apiai` will not load.

This is bit of a harsh break and it should probably still work for a while and just show a deprecation message.

## Steps to Reproduce
Configure the Dialogflow paser with the old name of `apiai`

## Expected Functionality
It should still work but show a deprecation warning.

## Experienced Functionality
The parser is not enabled.

## Versions
- **Opsdroid version:** v0.9.1-37-g7e2a067 (master)
- **Python version:** 3.5.4
- **OS/Docker version:** macOS High Sierra

## Configuration File
Please include your version of the configuration file bellow.

```yaml
##                      _           _     _
##   ___  _ __  ___  __| |_ __ ___ (_) __| |
##  / _ \| '_ \/ __|/ _` | '__/ _ \| |/ _` |
## | (_) | |_) \__ \ (_| | | | (_) | | (_| |
##  \___/| .__/|___/\__,_|_|  \___/|_|\__,_|
##       |_|
##                   __ _
##   ___ ___  _ __  / _(_) __ _
##  / __/ _ \| '_ \| |_| |/ _` |
## | (_| (_) | | | |  _| | (_| |
##  \___\___/|_| |_|_| |_|\__, |
##                        |___/
##
## A default config file to use with opsdroid

## Set the logging level
# logging:
#   level: info
#   path: opsdroid.log
#   console: true

## Set the location for opsdroid to install modules
# module-path: "."

## Show welcome message
welcome-message: true

## Configure the web server
# web:
#   host: '127.0.0.1'
#   port: 8080
#   ssl:
#     cert: /path/to/cert.pem
#     key: /path/to/key.pem

## Parsers
parsers:
  - name: apiai
    access-token: $DIALOGFLOW_ACCESS_TOKEN
    min-score: 0.6

## Connector modules
connectors:
  - name: shell
  - name: websocket

## Skill modules
skills:

  ## Hello world (https://github.com/opsdroid/skill-hello)
  - name: hello

  ## Last seen (https://github.com/opsdroid/skill-seen)
  - name: seen

  ## Dance (https://github.com/opsdroid/skill-dance)
  - name: dance

  ## Loud noises (https://github.com/opsdroid/skill-loudnoises)
  - name: loudnoises
```

## Additional Details
Any other details you wish to include such as screenshots, console messages, etc.

[...]

PR #307, opened 2017-10-29
Allows use of apiai in config

FabioRosado:
# Description

Fixed the problem with the parser not working if the apiai parser is set in the configuration.yaml
I apologise for this issue, I should had thought that people could still have the apiai set in their configuration. Everything should work fine now though!

**Fixes:** #306

## Status

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Ran unit tests locally and Travis CI - All passed
- [x] Tested parsing few skills/pages using api in config - worked as it should

# Checklist:

### DIFF
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -205,7 +205,18 @@ class OpsDroid():
                 _LOGGER.debug("Processing parsers...")
                 parsers = self.config["parsers"]
 
-                dialogflow = [p for p in parsers if p["name"] == "dialogflow"]
+                dialogflow = [p for p in parsers if p["name"] == "dialogflow"
+                              or p["name"] == "apiai"]
+
+                # Show deprecation message but  parse message
+                # Once it stops working remove this bit
+                apiai = [p for p in parsers if p["name"] == "apiai"]
+                if apiai:
+                    _LOGGER.warning("Api.ai is now called Dialogflow. This "
+                                    "parser will stop working in the future "
+                                    "please swap: 'name: apiai' for "
+                                    "'name: dialogflow' in configuration.yaml")
+
                 _LOGGER.debug("Checking dialogflow...")
                 if len(dialogflow) == 1 and \
                         ("enabled" not in dialogflow[0] or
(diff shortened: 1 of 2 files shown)


opsdroid/core.py
```diff
@@ -205,7 +205,18 @@ def start_databases(self, databases):
...
                 parsers = self.config["parsers"]
 
-                dialogflow = [p for p in parsers if p["name"] == "dialogflow"]
+                dialogflow = [p for p in parsers if p["name"] == "dialogflow"
+                              or p["name"] == "apiai"]
+
+                # Show deprecation message but  parse message
+                # Once it stops working remove this bit
+                # pylint: disable= using-constant-test
+                if "apiai":
```
FabioRosado: Jacob could you let me know if this statement is acceptable? Pylint was giving me a using-constant-test error so I disabled it. I tried to write this a few different ways but couldn't get the logging message to show. 

I've also added some comments so we can delete this bit of code when apiai stops working, I hope this is okay?

jacobtomlinson: I'm not sure that line makes sense. You're basically checking if a string is `True`.

I would recommend you restore the line:

```
apiai = [p for p in parsers if p["name"] == "apiai"]
```

And then change your if to say:

```
if apiai:
    _LOGGER.warning("Api.ai is now called Dialogflow. This "
                     "parser will stop working in the future "
                     "please swap: 'name: apiai' for "
                     "'name: dialogflow' in configuration.yaml")
```

You wouldn't use the `apiai` variable for anything else.

FabioRosado:
Changed that bit, I thought that if I included the previous line (the apiai one)+ the dialogflow one it would break that piece of code or count as another parser, I was so wrong haha

opsdroid/core.py
```diff
@@ -206,6 +206,16 @@ def start_databases(self, databases):
                 parsers = self.config["parsers"]
 
                 dialogflow = [p for p in parsers if p["name"] == "dialogflow"]
```
jacobtomlinson: This still needs to have the `or p["name"] == 'apiai'`.

jacobtomlinson:
Thanks!

Issue #155, opened 2017-05-31
Rank matching skills on parse


jacobtomlinson:
When matching skills each one should be given a score. This could be based on the certainty returned from an NLU API or based on the amount of matching context.

Issue #154, opened 2017-05-31
Only run one skill per message


jacobtomlinson:
Currently all matching skills will be executed when a message is parsed. This will add complications when dealing with context. Therefore it would be better to just run one skill.

Requires #155

go8ose:
How will ties be broken? If #155 is implemented it's conceivable that two matches will both have rank 1.0 (or 100%, however the ranking system works).  Perhaps core would respond with a message saying this is an error condition, and the configuration and logs should be checked, then not run any skill.

jacobtomlinson:
That's definitely an interesting question. I don't think resulting in an error is necessarily the best way to handle this. But I can't immediately think of an alternative.

gabeduke:
You could give a choice which skill to run

go8ose:
How do you give a choice?  Do you mean you have a third skill, an "arbitrator" skill that only sends messages in this situation, and asks the user how to proceed?  That sounds a bit awkward.

Maybe it should log an error message, and run all the matching skills.  Or maybe there should be a configuration option that picks between run all matches, run 0 matches, or run a random match

… actually, let's not provide a random option.  That sounds like a terrible idea.

gabeduke:
It's not the users fault that 2 skills recognize the same language. I feel like they should still be able to complete their request. The issue is that if it errors out they won't be able to use the skill unless they remove  the conflicting one. if they decide to keep both the bot could let them choose which path to follow.

go8ose:
@gabeduke I agree with you that it's not the users fault.  I just think it might be hard to implement.  What happens if opsdroid sends a reply to the user, and their answer comes back and two matches match this later message?  That's why I proposed a deterministic action driven by the configuration.  If there is also an alert sent to the user, the user can try and ask the administrator of the bot to do something about the situation (maybe altering a weighting that is given to a matcher).

jacobtomlinson:
I've had a go at implementing this in #311. 

The way ties are broken are by creating a list of skills which match and giving them a score, then I use the `sorted` method to order them and select the 0th element from the list. This does lead to a rather arbitrary tie breaker which is akin to randomly selecting one.

I agree with @go8ose that this is not ideal and it would be great for the user to feed back on this. I also agree with @gabeduke that the bot should complete some kind of action.

I'm also hopeful that the confidence scores returned by the NLU apis have enough significant figures that this is a very rare occurrence.

I would appreciate further feedback.

PR #311, opened 2017-10-30
Run one skill

jacobtomlinson:
# Description

This is a major change to functionality.

Instead of looping over all skills in each parser and executing all skills that match, each parser now simply returns a list of matching skills and a score for each one. The core then executes the highest scoring skill. This ensures only one skill is triggered by a message.

The primary benefit of this is that a skill can be decorated by multiple parsers and if the user has more than one enabled it will only be called once. For example a skill may have a `regex` matcher with a reasonably simple expression, and a `luisai` intent which is more flexible. If the user doesn't have LUIS enabled then the regex will work for them, if they do have LUIS enabled they will be able to more flexibly trigger the skill. But if the regex and LUIS match the skill will only be called once as expected.

**Fixes:** #154 and #155

## Status
**Finished**

## Type of change

# How Has This Been Tested?

- Tests have been updated to reflect these changes.
- I've run locally using the [apiai skill](https://github.com/opsdroid/skill-apiai) (with smalltalk enabled in Dialogflow) and the [hello skill](https://github.com/opsdroid/skill-hello). When saying `hello` only the apiai skill triggers. When disabling the apiai skill then the hello skill triggers.

# Checklist:


jacobtomlinson:
The decrease in coverage is due to a reduction in lines of relevant code. This is acceptable.

PR #348, opened 2017-11-27
Log a connector exit

Cadair:
# Description

If a connector exits, especially if it exits with an error we should log this. This adds a done callback on the listener task which logs the exception is one is raised. If the task was cancelled we do not log it as if the task is cancelled opsdroid is shutting down.

I am not sure how to write a unit test for this?

## Status
**READY** 

## Type of change

# How Has This Been Tested?

- Test A

Make a connector raise an exception which is not caught, i.e. to exit the coroutine with an exception.

- Test B

Make a connector where the listen coroutine returns a value.

# Checklist:


jacobtomlinson:
👍 this is a great idea.

Agree with @FabioRosado on testing methods.

Also a couple of questions:
 - Why is this function in the core? Would it not be better in the connector base class?
 - Why are you using partials?

Also I wonder if it would be useful for connectors to have enter and exit methods and to be instantiated with an `async with` instead. Then we could handle the close better.

Cadair:
@jacobtomlinson 

> Why is this function in the core? Would it not be better in the connector base class?

The issue is if the `listen` method exits, as core expects it to be an infinite loop. So the natural place for this felt like core as that is where the `Task` is created. I am not sure the context manager protocol would help without some major refactoring as again, this is to catch the `listen` method not the whole object. I do wonder as I am typing this if we should call the `disconnect` method when the `listen` method exits.

> Why are you using partials?

The Python documentation on `add_done_callback` explicitly tells you to use partials to pass parameters to the callback [here.](https://docs.python.org/3/library/asyncio-task.html#asyncio.Future.add_done_callback)

jacobtomlinson:
@Cadair happy with your comments.

Could you please look at adding tests so I can get this merged.
Cheers!

FabioRosado:
@Cadair I've updated my post, if you still need help creating tests let me know and I'll give you a hand 👍

jacobtomlinson:
@Cadair any update on this?

Cadair:
Sorry, other things have stolen my attention. I will try to get back to it asap.

jacobtomlinson:
It's been more than 30 days since the last activity on this PR. I'm going to close it but please feel free to reopen it if you wish to continue this work.

Issue #66, opened 2016-12-23
Add "user is typing" support


jacobtomlinson:
For chat services which support displaying a "user is typing" message there should be a method available on the base connector class for triggering this.

This would be particularly useful when implemented with a typing delay (#65).

Issue #65, opened 2016-12-23
Thinking delay and typing delay


jacobtomlinson:
Configuration options should be added to allow users to set delays for responses. This would allow a bot to appear more human.

#### Thinking delay

This delay would be added to every message and would simulate a user thinking about what to respond. This option should take the delay seconds as an integer or an array of two integers to determine the delay or the range for random delays respectively.

#### Typing delay

This delay would be added to the response of every message and would simulate the user typing out the message. This option should take an integer of "words per minute" to determine the length of the delay relative to the length of the response. It should also trigger the "user is typing" call during the delay, for connectors that support it.

gtseres:
Should this be added as a section in the configuration file? Maybe a `customizing` section, or something like that?

jacobtomlinson:
@gtseres gah just seen this comment sorry!

Potentially, although it could be best to add it to each connector.

PR #376, opened 2018-01-05
Add typing/thinking delay and user is typing event

FabioRosado:
# Description

This issue is a work in progress that adds thinking and typing functionality, I've actually been thinking about it for a while but only started to work on it now.  I'm making this PR just to show what I've wrote so far and to get some feedback and clarification on how to best implement these functionalities in opsdroid.

**Thinking delay** 
- defaults to 0 seconds - this to keeps opsdroid as fast as possible
- can be set in config.yaml with commands: `thinking-delay: < int or array>`
- accepts an int or array 

**Typing delay**

I'm dividing the character count of opsdroid's response by 6 - this means that every second opsdroid will type 6 characters which is +/- 66 words per second. This seems like a good number for someone that's a fast typer.

**Questions**
- Should I add a way for users to customise the typing delay or just allow the users to turn it on/off?
- Should these functions be left outside the message class?
- Is using get_opsdroid() the best way to to access config.yaml?

Hope this isn't too bad of a PR, please @jacobtomlinson let me know what you think 👍 

**Fixes** #66 & #65 

## Status

## Type of change

# How Has This Been Tested?

- Tox
- Ran opsdroid and tested the delays

# Checklist:

### DIFF
--- a/opsdroid/connector.py
+++ b/opsdroid/connector.py
@@ -72,6 +72,15 @@ class Connector():
         """
         raise NotImplementedError
 
+    async def user_typing(self, opsdroid):
+        """Signals that opsdroid is typing.
+
+        Triggers the "user is typing" event if the chat service that
+        opsdroid is connected to accepts it.
+        """
+
+        raise NotImplementedError
+
     async def disconnect(self, opsdroid):
         """Disconnect from the chat service.
 
--- a/opsdroid/connector.py
+++ b/opsdroid/connector.py
@@ -26,6 +26,11 @@ class Connector():
         self.config = config
         self.default_room = None
 
+    @property
+    def configuration(self):
+        """Class property used to access the connector config."""
+        return self.config
+
     async def connect(self, opsdroid):
         """Connect to chat service.
 
@@ -72,14 +77,17 @@ class Connector():
         """
         raise NotImplementedError
 
-    async def user_typing(self, opsdroid):
+    async def user_typing(self, opsdroid, trigger):
         """Signals that opsdroid is typing.
+
+        Args:
+            opsdroid (OpsDroid): An instance of the opsdroid core.
+            trigger: a bool that allows the event to be triggered on/off
 
         Triggers the "user is typing" event if the chat service that
         opsdroid is connected to accepts it.
         """
-
-        raise NotImplementedError
+        pass
 
     async def disconnect(self, opsdroid):
         """Disconnect from the chat service.
--- a/opsdroid/message.py
+++ b/opsdroid/message.py
@@ -2,7 +2,7 @@
 
 from datetime import datetime
 from copy import copy
-from time import sleep
+import asyncio
 from random import randrange
 
 from opsdroid.helper import get_opsdroid
@@ -22,20 +22,21 @@ class Message:
         self.regex = None
         self.responded_to = False
 
-    async def thinking_delay(self, seconds=0):
-        """Makes opsdroid wait x-seconds before responding."""
-        opsdroid = get_opsdroid()
-
-        if 'thinking-delay' in opsdroid.config:
-            seconds = opsdroid.config.get('thinking-delay')
+    async def _thinking_delay(self, seconds=0):
+        """Make opsdroid wait x-seconds before responding."""
+        seconds = self.connector.configuration.get('thinking-delay', 0)
 
         if isinstance(seconds, list):
             seconds = randrange(seconds[0], seconds[1])
-        sleep(seconds)
 
-    async def typing_delay(self, text):
+        await asyncio.sleep(seconds)
+
+    async def _typing_delay(self, text):
+        """Simulate typing, default is set to 6 characters per second."""
+        seconds = self.connector.configuration.get('typing-delay', 1)
+
         char_count = len(text)
-        sleep(char_count//6)
+        await asyncio.sleep(char_count//seconds)
 
     async def respond(self, text):
         """Respond to this message using the connector it was created by."""
@@ -43,9 +44,9 @@ class Message:
         response = copy(self)
         response.text = text
 
-        await self.thinking_delay()
-        await self.typing_delay(response.text)
-
+        await self._thinking_delay()
+        await self._typing_delay(response.text,)
+        print(self.connector.configuration)
         await self.connector.respond(response)
         if not self.responded_to:
             now = datetime.now()
--- a/opsdroid/message.py
+++ b/opsdroid/message.py
@@ -22,7 +22,7 @@ class Message:
         self.regex = None
         self.responded_to = False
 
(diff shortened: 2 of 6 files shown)


opsdroid/connector.py
```diff
@@ -72,6 +72,15 @@ def __init__(self, config):
         """
         raise NotImplementedError
 
+    async def user_typing(self, opsdroid):
```
jacobtomlinson: I would be tempted to add an argument to pass in `True` or `False` so it can be turned on and off.

This should also `pass` instead of raising an error as it is optional functionality for a connector.

opsdroid/message.py
```diff
@@ -21,11 +22,30 @@ def __init__(self, text, user, room, connector):
...
+    async def thinking_delay(self, seconds=0):
+        """Makes opsdroid wait x-seconds before responding."""
+        opsdroid = get_opsdroid()
+
+        if 'thinking-delay' in opsdroid.config:
+            seconds = opsdroid.config.get('thinking-delay')
+
+        if isinstance(seconds, list):
+            seconds = randrange(seconds[0], seconds[1])
+        sleep(seconds)
```
jacobtomlinson: This should be `asyncio.sleep()` not `time.sleep()` or it will block the thread.

opsdroid/message.py
```diff
@@ -21,11 +22,30 @@ def __init__(self, text, user, room, connector):
...
+
+    async def typing_delay(self, text):
+        char_count = len(text)
+        sleep(char_count//6)
```
jacobtomlinson: Again `asyncio.sleep()`.

FabioRosado:
@jacobtomlinson  I was playing around with the `@property` option in the connector. At the moment things are a bit broken but I think a way to fix one of the issues is if the property is not an async function.

I've also was able to get this from the following function:
```python
@property
def configuration(self):
    return self.config

>>> {'name': 'shell', 'type': 'connector', 'module_path': 'opsdroid-modules.connector.shell', 'install_path': '/Users/fabiorosado/.opsdroid/modules/opsdroid-modules/connector/shell', 'branch': 'master'}
```

So this brings up a question - should the delay be added under the connector in `configuration.yaml`? I could see the benefit in case we want to pass different delays for different connectors. So it could look like this:
```
{'name': 'shell', 'thinking-delay': 3, 'type': 'connector', 'module_path': 'opsdroid-modules.connector.shell', 'install_path': '/Users/fabiorosado/.opsdroid/modules/opsdroid-modules/connector/shell', 'branch': 'master'}
```

Let me know what you think Jacob, thanks!

jacobtomlinson:
Yes I think you're right it should be a connector config item.

In terms of testing I think that test is fine, it's a bit minimal but it check thats the thinking delay method exists and is called.

I would also recommend mocking the `asyncio.sleep` method and assert that it has been called with the right number of seconds. You could also test the range to ensure it is always between the correct values.

jacobtomlinson:
@FabioRosado how are you getting on with this? I'm thinking about cutting a release at some point soon and it would be great to get this in.

FabioRosado:
@jacobtomlinson I've just updated the code, can you check how does it look like?
Also, could you test how long does opsdroid take to reply if there's no delay set in the connector? I tested it on Pycharm and it took a long while (even though it should be 0 or 1 second). 

Thanks

opsdroid/message.py
```diff
@@ -21,11 +22,31 @@ def __init__(self, text, user, room, connector):
...
+
     async def respond(self, text):
         """Respond to this message using the connector it was created by."""
         opsdroid = get_opsdroid()
         response = copy(self)
         response.text = text
+
+        await self._thinking_delay()
+        await self._typing_delay(response.text,)
+        print(self.connector.configuration)
```
jacobtomlinson: Need to remove this or log it with `_LOGGER`.

opsdroid/message.py
```diff
@@ -21,11 +22,31 @@ def __init__(self, text, user, room, connector):
...
+        seconds = self.connector.configuration.get('thinking-delay', 0)
+
+        if isinstance(seconds, list):
+            seconds = randrange(seconds[0], seconds[1])
+
+        await asyncio.sleep(seconds)
+
+    async def _typing_delay(self, text):
+        """Simulate typing, default is set to 6 characters per second."""
+        seconds = self.connector.configuration.get('typing-delay', 1)
```
jacobtomlinson: Why does this default to one character per second? I would expect that if it hasn't been configured there should be no delay. 

I would go for something like this and skip the sleep if it hasn't been configured.

```python
try:
    seconds = self.connector.configuration['typing-delay']
    char_count = len(text)
    await asyncio.sleep(char_count//seconds)
except KeyError:
    pass
```

FabioRosado:
The tests are not passing, all I had to do is add an if statement in the `message.respond` and include both delays. 

I'm a bit confused with thinking delay because no matter what I try on the tests the  `asyncio.sleep`  doesn't seem to be called, even though when I run opsdroid with a thinking delay it takes that into consideration.

Oddly enough, the typing delay doesn't seem to suffer from that. I tried passing response.text to thinking delay but that didn't made the tests pass. 

If you have any idea why this might happening I'd like some help haha

jacobtomlinson:
I'm having a look now and am also confused 😆 .

FabioRosado:
Haha yeah and oddly enough when I print the seconds on the thinking delay function only and run the test only the typing delay shows the printing but it shows their own configuration settings - so if you set it to 6 it will print 6 even though the printing is in another function o.o

Could it be an issue with so many awaits?

jacobtomlinson:
Ah I think I understand. You're mocking the function so only the other function is going to get called. 

So you can **either** mock the function and assert that it has been called **or** mock `asyncio.sleep` and assert that has been called.

FabioRosado:
Oh that makes sense, although the typing_delay raises the right value when I test the sleepmock. Anyhow I've just tweaked the test, I've added both the int and list type in the same test, let me know if you would rather have two different tests.

I wasn't sure how to test `connector.user_typing` so I took the example of `connector.disconnect` and that seem to work, okayish.

Thanks a lot for the help @jacobtomlinson I was really confused as to why the `asyncio.sleep` test was failing.

jacobtomlinson:
Yeah I guess thats because the thinking delay is running with a sleep of 0 which satisfies the test.

Looks like the linter is complaining about a couple of things and the coverage hasn't updated for some reason. If you can just tweak the lint I'll get it merged.

jacobtomlinson:
No worries. I tend to just run `tox -e py36,lint` locally to keep it fast.

It looks like the coverage is due to the tests not actually testing the `_typing_delay` method fully. Could you add a test which doesn't mock that method and has `typing-delay` configured so that all lines are executed?

Issue #385, opened 2018-01-14
FileNotFound error: No such file or directory: pip


FabioRosado:
# Description
Honestly this issue might be just my own fault. I probably installed python 3 wrong when I first got my Macbook pro. Anyway, when I try to run opsdroid with a skill that contains a `requirements.txt`  I get a `FileNotFoundError: [Errno 2] No such file or directory: 'pip'` exception.

If I change 'pip' in  `opsdroid.loader.pip_install_deps`  for 'pip3' everything works fine.

Should a try/except be used just in case someone has the same issue?

## Steps to Reproduce
As this is an issue with my python install I'm not sure how to reproduce it...

## Expected Functionality
Opsdroid to run normally.

## Experienced Functionality
Explain what happened instead(Please include the debug log).

```python
INFO opsdroid: ========================================
INFO opsdroid: Started application
INFO opsdroid: ========================================
INFO opsdroid: You can customise your opsdroid by modifying your configuration.yaml
INFO opsdroid: Read more at: http://opsdroid.readthedocs.io/#configuration
INFO opsdroid: Watch the Get Started Videos at: http://bit.ly/2fnC0Fh
INFO opsdroid: Install Opsdroid Desktop at: 
https://github.com/opsdroid/opsdroid-desktop/releases
INFO opsdroid: ========================================
WARNING opsdroid.loader: No databases in configuration.
Traceback (most recent call last):
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/__main__.py", line 140, in <module>
    init()
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/__main__.py", line 137, in init
    main()
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/__main__.py", line 130, in main
    opsdroid.start_loop()
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/core.py", line 134, in start_loop
    self.loader.load_modules_from_config(self.config)
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/loader.py", line 222, in load_modules_from_config
    skills = self._load_modules('skill', config['skills'])
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/loader.py", line 271, in _load_modules
    self._install_module(config)
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/loader.py", line 317, in _install_module
[...]

PR #387, opened 2018-01-15
Try pip3 if pip is not found

FabioRosado:
# Description

Opsdroid runs smoothly again when the try/except is added to the loader.
I would like a feedback on the test if possible, I'm not sure if I'm testing it properly or not.

**Fixes** #385

## Status

## Type of change

# How Has This Been Tested?

- Tox - tests passed
- Travis - Tests failed

# Checklist:

### DIFF
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -103,21 +103,34 @@ class Loader:
     @staticmethod
     def pip_install_deps(requirements_path):
         """Pip install a requirements.txt file and wait for finish."""
+        process = None
         command = ["pip", "install",
                    "--target={}".format(DEFAULT_MODULE_DEPS_PATH),
                    "--ignore-installed", "-r", requirements_path]
+
         try:
             process = subprocess.Popen(command,
-                                        shell=False,
-                                        stdout=subprocess.PIPE,
-                                        stderr=subprocess.PIPE)
+                                       shell=False,
+                                       stdout=subprocess.PIPE,
+                                       stderr=subprocess.PIPE)
 
         except FileNotFoundError:
+            _LOGGER.debug("Couldn't find the command 'pip', "
+                          "trying again with command 'pip3'")
+
+        try:
             command[0] = "pip3"
             process = subprocess.Popen(command,
-                                        shell=False,
-                                        stdout=subprocess.PIPE,
-                                        stderr=subprocess.PIPE)
+                                       shell=False,
+                                       stdout=subprocess.PIPE,
+                                       stderr=subprocess.PIPE)
+        except FileNotFoundError:
+            _LOGGER.debug("Couldn't find the command 'pip3', "
+                          "install of %s will be skipped.",
+                          str(requirements_path))
+
+        if not process:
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
#388 should resolve the CI issues you were having. I've updated your branch to include the new library.

jacobtomlinson:
Great Travis has passed now.

FabioRosado:
Thanks for the help jacob, I wasn't sure how to implements you suggested fully the first time 👍 
It seems that the test I created for the except raise isn't covering it for some reason, I'll try to play around with it and see if I can fix the issue

jacobtomlinson:
Thinking about it we might want to raise an exception and exit instead of `return False` because if deps fails to install you can be reasonably sure opsdroid will not function correctly.

At the very least it probably needs to log an error saying it can't find pip.

FabioRosado:
Is this the general function for any required installs for opsdroid? If this is just used to install requirements for skill/connectors it shouldn't be too bad, the skill would just not work I guess?

If its used to install everything then yeah I get it might be a problem

jacobtomlinson:
It's for module dependency installs, so connectors, databases and skills.

jacobtomlinson:
Looking good. Needs a test for both exceptions. Just needs a side effect on `subprocess.Popen` to raise the `FileNotFoundError` and the test should assert that it raises the overall error.

Also when raising if both pip installs fail perhaps it should be an `OSError` or a `RuntimeError` instead of a `FileNotFoundError`.

FabioRosado:
I was just writing in gitter. I'm thinking in giving this a try:
```python
```
    def test_no_pip_install(self):
        opsdroid, loader = self.setup()
        with mock.patch.object(subprocess, 'Popen') as mocked_popen:
            mocked_popen.side_effect = FileNotFoundError()
            self.assertRaises(FileNotFoundError, mocked_popen)
            with mock.patch.object(loader, 'pip_install_deps') as pip_install:
                mocked_popen.return_value.\
                    communicate.return_value = ['Test\nTest']
                loader.pip_install_deps("/path/to/some/file.txt")
                self.assertTrue(pip_install.called)
```

The way I think about it is that: popen raises the exception. Then it tries again and returns the right value (with pip3 I guess) and then it all passes. The code works on my end but just wanted to check with you. 

I was unsure which exception to raise, I'll change it with OSError one instead.

FabioRosado:
Hmm... okay that change in the tests didn't work

jacobtomlinson:
I find it easier to use `self.assertRaises` in a with statement. That might help.

You've also patched `with mock.patch.object(loader, 'pip_install_deps')` further down which means your `mocked_popen` wont get called at all within that block.

FabioRosado:
Oh wow... I have to say that I didn't expect this tests to pass but.... hey, they did! 
I would still love to get some feedback from you @jacobtomlinson just to know if you would test this try/except block differently.

Thanks 👍

jacobtomlinson:
This looks great! The only comment I have is that you don't need the OSError side effect in the test. The code will already be raising that error. You should be able to just remove that from the array and it's ready to be merged.

PR #406, opened 2018-01-18
remove useless pylint disable comment

afibanez:
# Description

I was reading some code to understand Opsdroid before select some issue to collaborate and I noticed that there is one useless pylint disable comment.

Then I searched across all the repo looking for more, but that's the only one :smile: 

## Status
**READY**

## Type of change

_It's not exactly a bug fix, but it's the closest category_

# How Has This Been Tested?

I just run again the linter and all the testsuite with tox

# Checklist:

### DIFF
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -315,7 +315,6 @@ class Loader:
         return loaded_modules
 
     def _install_module(self, config):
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Good spot! That's definitely not necessary any more.

Issue #295, opened 2017-10-22
When installing modules from a git repo, make sure they are up to date on startup.


Cadair:
Correct me if I am wrong but it appears that when you pull a component in using the `repo:` or in fact when it comes from the opsdroid org, when opsdroid starts it does not fetch updates to the repo?

Is this intended behaviour or something that could be changed?

jacobtomlinson:
This is definitely desired functionality 😄 .

afibanez:
I will do a PR with a proposal :smiley:

afibanez:
So now we can close this issue :wink:

PR #409, opened 2018-01-19
Update git modules on startup

afibanez:
# Description

I saw an issue asking for updating modules from git repositories and I found it interesting.

This change just add git pull capabilities to Opsdroid, and avoids strange behaviours forcing only fast forward.

Some other points to consider:
- The common code in `process.communicate()` is refactored to a method, and the `wait()` call is suppressed (`communicate` [already waits](https://docs.python.org/3.6/library/subprocess.html#subprocess.Popen.communicate))
- `build_module_path` is refactored and splitted into two methods, to avoid the use of [flag arguments](https://www.martinfowler.com/bliki/FlagArgument.html)
- `_is_module_installed` and `_is_local_module` are refactors of some `ifs` to simplify the level of abstraction

**Fixes** #295

## Status
**READY**

## Type of change

# How Has This Been Tested?

I fixed `build_module_path` test, removed `test_install_existing_module` (as now has no sense), and added some tests to check the new behaviour:

- test_load_existing_modules
- test_update_existing_local_module
- test_update_existing_git_module

# Checklist:

### DIFF
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -226,6 +226,8 @@ connectors:
     repo: https://github.com/username/myconnector.git
 ```
 
+_Note: When using a git repository, Opsdroid will try to update it at startup pulling with fast forward strategy._
+
 #### Local Directory
 
 A local path to install the module from.
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -76,16 +76,17 @@ class Loader:
             if os.path.isfile(config["install_path"] + ".py"):
                 os.remove(config["install_path"] + ".py")
 
-    def build_module_path(self, path_type, config):
-        """Generate the module path from name and type."""
-        if path_type == "import":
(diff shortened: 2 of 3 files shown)


jacobtomlinson:
Thanks @afibanez this is a great PR. I'll give it a test locally but it looks like you've been thorough with it.

As for docs it should probably go in the configuration reference section, perhaps [here](http://opsdroid.readthedocs.io/en/latest/configuration-reference/#git-repository).

afibanez:
Thanks for the feedback @jacobtomlinson. Of course you have to test it, I could have miss something.

For docs... should I write at the end of the section you mentioned something like this?
> Note: When using a git repository, Opsdroid will try to update it at startup pulling with fast forward strategy.

jacobtomlinson:
@afibanez sounds good!

jacobtomlinson:

Issue #422, opened 2018-01-24
Lozalize opsdroid


afibanez:
Opsdroid is designed with English in mind. A developer can write his skills (regex and responses) in any other language, but actually it has some complications:
- Some parts of the core code assume English (see [this](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/core.py#L223)  or [this](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/parsers/recastai.py#L15) examples)
- As a developer you could want have your skill ready for multiple languages:
    - But you can only have one matcher, and write a regexpr ready for various languages it's hard and difficult to mantain
    - And there is no system to translate the answer strings, so it has to be implemented in every skill that need it

So let's discuss what we need to do to go multilanguage (i already used some @jacobtomlinson ideas)

Breaking changes
----------------
The idea of the multi language design is to don't break anything. So if you don't change the language, Opsdroid has to assume english and work as normally.

Configuration
-------------
We need a new `language` configuration item, that can be filled with any [ISO 693-1 two-letter code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes). By default it will be `en`.

Core localization
-----------------
We need to use some Python translation library. In my opinion, we should use the [gettext](https://docs.python.org/3/library/gettext.html) library, it is extended, easy to use, and in standard libraries. [Here](https://inventwithpython.com/blog/2014/12/20/translate-your-python-3-program-with-the-gettext-module/) is a post explaining the usage.

So the rule should be mark as translatable every string that can be read by the end user through the chat (like the Whoops error).

Parser/Matcher localization
---------------------------
Now we have different behaviours:
- For `always` or `crontab` there is no problem.
- For `regex` we should add some extra decorators with the locale code at the end (for example `@match_regex_es`). The basic decordator will be mandatory, and it will point always to english. And if the configurated language is not defined by a decorator, it will default to english. For example, for the `skill-hello`, it can be like:
```python
@match_regex(r'hi|hello|hey|hallo')
@match_regex_es(r'hola|buenas|saludos')
async def hello(opsdroid, config, message):
    [...]
```
- `recastai` or `dialogflow` requires a language code to be sent in every request. So my proposal is to define a list of accepted languages (defined by the service) and check if the configurated language is in it. If the configurated language it's not suported by the service, fallback to english.
[...]

PR #423, opened 2018-01-24
Update documentation/parsers to allow different languages

FabioRosado:
# Description

Start tackling work for #422, since only two of the parsers need the language to be specified in the API Headers the documentation was updated as well to mention how to use the matcher in a different language.

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Unittest + Tox
- Try to use recastai without a lang parameter in the configuration. 

# Checklist:

### DIFF
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -8,6 +8,7 @@
    - [Installation Path](#installation-path) 
    - [Skills](#skills)
    - [Time Zone](#time-zone)
+   - [Language](#language)
    - [Web Server](#web-server)
 - [Module Options](#module-options)
    - [Install Location](#install-location)
@@ -192,11 +193,14 @@ timezone: 'Europe/London'
 ```
 
 ### Language
-Configure the language to use Opsdroid.
+Configure the language to use opsdroid.
 
 To use opsdroid with a different language other than English you can specify it in your configuration.yaml. The language code needs to be in the standardized [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes).
+
+_Note: If no language is specified, opsdroid will default to English._
+
 ```yaml
-language: <ISO 639-1 code -  example: 'en'>
+lang: <ISO 639-1 code -  example: 'en'>
 ```
 
 ### Web Server
@@ -234,7 +238,7 @@ connectors:
     repo: https://github.com/username/myconnector.git
 ```
 
-_Note: When using a git repository, Opsdroid will try to update it at startup pulling with fast forward strategy._
+_Note: When using a git repository, opsdroid will try to update it at startup pulling with fast forward strategy._
 
 #### Local Directory
 
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -191,6 +191,14 @@ This timezone will be used in crontab skills if the timezone has not been set as
 timezone: 'Europe/London'
 ```
 
+### Language
+Configure the language to use Opsdroid.
+
+To use opsdroid with a different language other than English you can specify it in your configuration.yaml. The language code needs to be in the standardized [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes).
+```yaml
+language: <ISO 639-1 code -  example: 'en'>
+```
+
 ### Web Server
 
 Configure the REST API in opsdroid.
--- a/docs/matchers/dialogflow.md
+++ b/docs/matchers/dialogflow.md
@@ -30,7 +30,7 @@ The above skill would be called on any intent which has a name of `'myIntent'`.
 
 You can find a quick getting started with Dialogflow guide [here](https://dialogflow.com/docs/getting-started/basics).
 
-If you want to use Dialogflow in a different language other than English, all you need to do is specify the `lang` parameter in Opsdroid's configuration. Then change/add another language to your Dialogflow agent in the Language tab of the agent settings.
+If you want to use Dialogflow in a different language other than English, all you need to do is specify the `lang` parameter in opsdroid's configuration. Then change/add another language to your Dialogflow agent in the Language tab of the agent settings.
 
 _Useful Links: [Languages Reference](https://dialogflow.com/docs/reference/language), [Multi-language Agents Reference](https://dialogflow.com/docs/multi-language)_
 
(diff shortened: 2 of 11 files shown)


jacobtomlinson:
Thanks for raising this!

I see you've just added the `lang` option to the recast.ai config. Will it not get from the main `lang` parameter in the config file? Are there benefits from also configuring it separately? 

I seem to remember this being in @afibanez s example so perhaps he has some thoughts.

afibanez:
Yeah, I think that for now it's better to have lang configured one time in main configuration.
Configure the lang inside the parser config only has sense if we move to a multilang implementation (having multiple languages working at same time),
But as we agree, the objective now is localize opsdroid with one language active.

FabioRosado:
I see what you mean, I've tried to get the lang parameter from the main config but I wasn't able to do it yet. If I try to add opsdroid to the `call_recastai` opsdroid hangs. I'll try to think for a way to get the main config file and use that lang parameter 👍

jacobtomlinson:
You could add it to the connector config in `parse_recastai`.

```python
async def parse_recastai(opsdroid, message, config):
    """Parse a message against all recastai intents."""
    if 'lang' not in config:
        config['lang'] = opsdroid.config.get('lang', 'en')
    ...
```

FabioRosado:
This is what I came up with, it seems to work okay. Could you let me know what you think about it?
I've also added the language reference in the `configuration-reference.md`

docs/configuration-reference.md
```diff
@@ -191,6 +191,14 @@ This timezone will be used in crontab skills if the timezone has not been set as
 timezone: 'Europe/London'
 ```
 
+### Language
+Configure the language to use Opsdroid.
+
+To use opsdroid with a different language other than English you can specify it in your configuration.yaml. The language code needs to be in the standardized [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes).
+```yaml
+language: <ISO 639-1 code -  example: 'en'>
```
jacobtomlinson: This should be `lang` not `language`.

jacobtomlinson: You might also want to mention that it defaults to English.

docs/matchers/dialogflow.md
```diff
@@ -30,6 +30,11 @@ The above skill would be called on any intent which has a name of `'myIntent'`.
 
 You can find a quick getting started with Dialogflow guide [here](https://dialogflow.com/docs/getting-started/basics).
 
+If you want to use Dialogflow in a different language other than English, all you need to do is specify the `lang` parameter in Opsdroid's configuration. Then change/add another language to your Dialogflow agent in the Language tab of the agent settings.
```
jacobtomlinson: Let's have a lowercase opsdroid for consistency.

docs/matchers/recast.ai.md
```diff
@@ -36,6 +36,10 @@ You need to [register](https://recast.ai/signup) on Recast.AI and create a bot i
 
 You can find a quick getting started with the Recast.AI guide [here](https://recast.ai/docs/create-your-bot).
 
+If you want to use Recast.AI in a different language other than English, all you need to do is specify the `lang` parameter in Opsdroid's configuration.
```
jacobtomlinson: Lower case opsdroid again.

FabioRosado:
Do you agree with the change on the `call_recastai` and `call_dialogflow`? I need to add a default option on the lang param tho.

docs/configuration-reference.md
FabioRosado: I wasn't sure what to use so I thought using the full term would be more specific, i'll change this in the fix

jacobtomlinson:
Yeah I'm happy with that. You're already setting the default option in the `config.get()` call so I wouldn't double up. You might want to create a constant in `const.py` called something like `DEFAULT_LANGUAGE` and import and use that rather than dropping the `'en'` string around.

FabioRosado:
I didn't want to touch the tests to make this PR pass so I changed things around a bit. Added the DEFAULT_LANGUAGE int he const and used it as a default to `lang` in the `call_parser` function.

Changed the `.get('lang', 'en')` to just `.get('lang')` if the key is not found the default value is still replaced.

I tried using different lang strings in the config to see if the parser would show the right language (it does), if no value is passed, then the DEFAULT_LANGUAGE value is used.

Let me know what you think 😄

jacobtomlinson:
If you do not specify a default option in `.get()` it will return `None`. This will then be passed to the method and so the `DEFAULT_LANGUAGE` will never be used.

FabioRosado:
Ah I see. Since the test was showing the English language I assumed everything was going okay. 
Am I wrong thinking that the default value in the lang should be used due to backwards compatibility? Or I can just add the `lang` param in the tests?

jacobtomlinson:
Feel free to test both.

You're right that it should default to 'en' for backward compatibility, but at the same time you are also updating every call to that function so you can be pretty sure that you wont break backward compatibility.

I would just restore the default option in the `.get()` methods using the `DEFAULT_LANGUAGE` constant then you've covered all the bases. It's one of the reasons I suggested creating the constant, so you can use it in both places.

FabioRosado:
Okay I've added the default value in the .get() and removed it from the call_parser method. 
I had to update the test to add another param to call_parser (Just added None). Do you think this would be ok?

afibanez:
I think it's better to maintain the default arg in the method definition, and then use it in the get:
```pyhon
async def call_dialogflow(message, config, lang=DEFAULT_LANGUAGE):
[...]
opsdroid.config.get("lang", lang))
```

jacobtomlinson:
@afibanez the point from before is that the `opsdroid` object is not accessible in `call_dialogflow` so what you propose is not possible.

I was suggesting having `DEFAULT_LANGUAGE` in **both** the method definition and the `opsdroid.config.get()` method. It's potentially unnecessary but it covers all bases.

afibanez:
@jacobtomlinson ok! I was tricked reading the change in github, with the summarize of changes i was confused about where the get line is :confused:

FabioRosado:
Apologies about this, I miss understood what you meant Jacob and kind off messed up a bit.
I've applied the changes to the call_parser and reverted the tests back to normal since we don't need to touch them to pass. Thanks for all the help!

PR #433, opened 2018-02-03
Use gettext for translations

anxodio:
# Description

Add gettext as agreed on #422.

Gettext in python is a little mess to get it working (and there are only few examples and doc), I will do a blog post explaining all the process, but now in opsdroid it's configured with [babel](http://babel.pocoo.org/en/latest/index.html) to be easy to use.

I must be grateful to this [blog](https://www.mattlayman.com/2015/i18n.html) post, with it and the [source](https://github.com/handroll/handroll/blob/master/setup.py) I realized how it works.

@jacobtomlinson, as you can see in the docs update, python only loads compiled messages that aren't versioned by convention. So like in the post, I added some `setup.py` improves in order to assert that compiled messages are updated when building.

## Status
**READY**

## Type of change

# How Has This Been Tested?

There are no automated tests for translations. I checked that all opsdroid automated tests continue working, so the introduction of translations don't break anything.

And I tested the new localization system with Catalan and Spanish translations.

# Checklist:


afibanez:
Ok, I don't know why in the travis build flake8 is looking into babel source and failing 😞

jacobtomlinson:
Sorry for the delay in reviewing I've been on holiday. Will take a look now.

jacobtomlinson:
Looks like you might need to add `.eggs` to the list of excluded directories [here](https://github.com/opsdroid/opsdroid/blob/master/setup.cfg#L9).

anxodio:
First of all, review all my english, because I need to polish a lot of things! Thanks for the grammar fixes ;)

The `_` is defined in `__main__.py`, in the line `gettext.install('opsdroid')`. Gettext installs the function `_` in the built in namespace (read [this](https://docs.python.org/3/library/gettext.html#localizing-your-application)).

jacobtomlinson:
No worries, the rest of it looks good!

Ah that makes sense. Do you think that means it will be available in all skills without any changes?

docs/contributing.md
```diff
@@ -84,6 +88,39 @@ Things that we need help with:
...
+```shell
+python setup.py compile_catalog
+```
+
+### Starting a new language
+If your language is not in the `locale` folder, you can initialize it. You will need the [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) code of the language. For example, in order to start a new [Esperanto](https://en.wikipedia.org/wiki/Esperanto) translation:
+```shell
+python setup.py init_catalog -l eo
+```
+And then you can translate it in `locale/eo/LC_MESSAGES/opsdroid.po`, then compile it, etc.
```
jacobtomlinson: I would remove the `And` from the beginning of this sentence.

anxodio:
> Do you think that means it will be available in all skills without any changes?
I would say yes, because it is in opsdroid namespace and the skills are executed by opsdroid

anxodio:
I fixed the doc text, added .eggs into pep8 excludes (thanks @jacobtomlinson ) and discovered the way to compile the messages with `pip install -e .` (so the extra step for installing is removed) 🎉

Now I see that I have to look for the way to test `configure_lang` method 🤔

jacobtomlinson:
Ah yes I always forget about `-e`.

I would write two tests. In both of them patch `gettext.translation`. Then in one pass `{}` to `configure_lang` and assert that `lang.install()` is not called, then in the second pass `{'lang': 'es'}` and assert that it is called.

jacobtomlinson:
Awesome!

Issue #447, opened 2018-02-09
Update regex docs to encourage named regex groups


jacobtomlinson:
# Description
Inspired by #446 we should update the documentation on the regex matcher to encourage people to use [named regex groups](https://docs.python.org/3/howto/regex.html#non-capturing-and-named-groups).

When capturing a group from a regular expression you can write something along these lines

```python
@match_regex('my name is (\w+)')
async def my_name_is(opsdroid, config, message):
    name = message.regex.group(0)
```

However this means that if you include multiple groups you need to know their order to get them our successfully, which can make things complicated.

```python
@match_regex('my name is (\w+) and my wife is called (\w+) and my son is called (\w+)')
async def my_name_is(opsdroid, config, message):
    name = message.regex.group(0)
    wife = message.regex.group(1)
    son = message.regex.group(2)
```

As #446 will add the ability to extract entities by their name then this breaks down as it will not be possible to request the group by it's index.

In the python `re` library it is possible to [give your groups names](https://docs.python.org/3/howto/regex.html#non-capturing-and-named-groups) and then retrieve them using those names.

```python
@match_regex('my name is (?P<name>\w+) and my wife is called (?P<wife>\w+) and my son is called (?P<son>\w+)')
async def my_name_is(opsdroid, config, message):
    name = message.regex.group('name')
    wife = message.regex.group('wife')
    son = message.regex.group('son')
```

This should be the encouraged behaviour as it will simplify entity retrieval in the long run. We should update the documentation page to encourage this.

Thai-Morris:
Hello, I am new to this open source contribution thing. How can I help?

jacobtomlinson:
Welcome to open source! Great to have you.

The documentation page this issue is referring to can be found [here](https://github.com/opsdroid/opsdroid/blob/master/docs/matchers/regex.md). 

You'll need to fork the repo, edit the page, and then open a pull request. [GitHub has a guide here](https://help.github.com/articles/creating-a-pull-request-from-a-fork/).

If you need any further help you can [reach out to us on gitter](https://gitter.im/opsdroid)!

PR #450, opened 2018-02-09
add a parser using parse lib (proposal)

anxodio:
# Description

I was reading the issue #447, thinking about the code in:
```python
@match_regex('my name is (?P<name>\w+) and my wife is called (?P<wife>\w+) and my son is called (?P<son>\w+)')
async def my_name_is(opsdroid, config, message):
    name = message.regex.group('name')
    wife = message.regex.group('wife')
    son = message.regex.group('son')
```
I thought that that's very "complicated" for just get some parameters. Regex it's very powerful and flexible, but for easy cases like this it's an overhead.

That's the reason why I make this proposal aiming to add a new core parser using [parse](https://github.com/r1chardj0n3s/parse) python lib.

It simplify a lot that type of matchers, using the classic format python style. For example, the example above will become:
```python
@match_parse('my name is {name} and my wife is called {wife} and my son is called {son}')
async def my_name_is(opsdroid, config, message):
    name = message.parse_result['name']
    wife = message.parse_result['wife']
    son = message.parse_result['son']
```

It can be very useful for easy matches, and can help to developers with few regex skills.

What do you think? If you like it, I will finish the implementation with tests and doc 😄 

## Status
**UNDER DEVELOPMENT**

## Type of change

# How Has This Been Tested?

It must be tested if we find it interesting.

# Checklist:

### DIFF
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -14,6 +14,7 @@ from opsdroid.database import Database
 from opsdroid.loader import Loader
 from opsdroid.parsers.always import parse_always
 from opsdroid.parsers.regex import parse_regex
+from opsdroid.parsers.parseformat import parse_format
 from opsdroid.parsers.dialogflow import parse_dialogflow
 from opsdroid.parsers.luisai import parse_luisai
 from opsdroid.parsers.recastai import parse_recastai
@@ -228,6 +229,7 @@ class OpsDroid():
         """Take a message and return a ranked list of matching skills."""
         skills = []
         skills = skills + await parse_regex(self, message)
+        skills = skills + await parse_format(self, message)
 
         if "parsers" in self.config:
             _LOGGER.debug(_("Processing parsers..."))
--- a/opsdroid/matchers.py
+++ b/opsdroid/matchers.py
@@ -5,7 +5,6 @@ import logging
 from opsdroid.helper import get_opsdroid
 from opsdroid.web import Web
 
-
 _LOGGER = logging.getLogger(__name__)
 
 
@@ -16,6 +15,19 @@ def match_regex(regex, case_sensitive=True):
         opsdroid = get_opsdroid()
         opsdroid.skills.append({"regex": {"expression": regex,
                                           "case_sensitive": case_sensitive},
+                                "skill": func,
+                                "config":
+                                opsdroid.loader.current_import_config})
+        return func
+    return matcher
+
+
+def match_parse(format_str):
+    """Return parse match decorator."""
+    def matcher(func):
+        """Add decorated function to skills list for parse matching."""
+        opsdroid = get_opsdroid()
+        opsdroid.skills.append({"parse_format": format_str,
                                 "skill": func,
                                 "config":
                                 opsdroid.loader.current_import_config})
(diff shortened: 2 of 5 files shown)


jacobtomlinson:
Love it! Please continue. 

This would be perfect for getting started with opsdroid.

anxodio:
I'm thinking about the score... I don't know how to calculate it. I don't understand the approach used in regex :/

My point is that you use something like:
```python
@match_parse('hello')
@match_parse('hello {name}, how are you?')
``` 
And parse resolve it, probably it's a 100% score. But the problem can be:
```python
@match_parse('hello {name}')
```
In this case, name will parse all things until the end of the sentence. It will match `hello Jacob`, but it will match `hello Jacob, how are you` too (setting `name=Jacob, how are you`).
So use here a 100% score its a bad thing.

Any troughs?

jacobtomlinson:
Sorry for the delay! This is a tricky one so I'll try and explain my thinking here, I'm really keen to get feedback too.

In the case where multiple skills match a message only the most certain match is used. All NLU APIs provide a certainty score for the matches. This means that skills can have numerous matchers on them and depending on the configuration only some of them will be used, but in the case where two match the skill is only called once.

For example if I configure both wit and luis and they both match the one which is most confident is used.

Applying this to regex was tricky as it doesn't provide a confidence, so I had to decide on a way to calculate this myself. Here were my thoughts when implementing #311:

- Matches from APIs such as LUIS should usually win over regex, therefore regex should have a lower score distribution.
- The more complicated the regex the higher the score should be. Currently length is used to calculate this.
- Regex skills should have some kind of upper limit on their score.

Therefore I think the same principals should be applied here, but I'm keen to know what you think!

anxodio:
> Matches from APIs such as LUIS should usually win over regex, therefore regex should have a lower score distribution.

I'm thinking about this and I'm not sure if I agree. The point for me is that NLU APIs are less specific than regex or parse.

For example, you add a _greetings_ intent to some NLU training with various sentences like 'hello', 'hello opsdroid', 'hi', 'hello there', 'whats up', 'good morning'...
And you have the min-score at 0.6.

Then you have a regex matcher with something like 'hello from (?P<location>\w+)!'.

If a user writes "hello from Barcelona!" opsdroid will prioritize the NLU response, because it's almost secure that the NLU platform will detect the intent into the sentence.

So you regex will be never executed, even being more specific.

jacobtomlinson:
Yep that's totally fair! Thanks for the feedback. 

Do you have any thoughts on an alternative approach?

anxodio:
I think I would prioritize regex and parse over NLU engines, because they are more specific.
My point here is that if one developer is using an NLU engine, probably most of his/her opsdroid skills will be under the NLU engine. So if it's using some regex/parse, will be for something more concrete and if the regex/parse skill matches, probably it's the expected result of the developer.

My bet will be prioritize always parse > regex > NLU, from more specific to less specific, and remove score from parse/regex, or just use it if two different regex/parse are matching at the same time.

But, this is my experience and opinion. I don't know any real use cases other than mine, so I think the best strategy here is talk with the community and users, and try to discover what would be the preferred way to resolve this kind of decision.

anxodio:
I will let this PR in standby waiting for #458 , because the final implementation will change depending on the resolution of the issue.

jacobtomlinson:
#458 has gone stale now so I'm going to close this off. Would be great to revisit it though.

PR #453, opened 2018-02-10
Update logging messages to use gettext

FabioRosado:
# Description

This is another part of #422 most logging messages will now be translated - some were causing texts to fail (mostly in `opsdroid.loader`) so i had to leave them untouched.

I've replaced the logging formating from: `logging.info(msg, arg)` to `logging.info(msg % arg)` I'm not sure if this is the best approach. It's possible to just translate the text from the logging by using the format: `logging.info(_(msg), arg)` let me know if this is preferred than the one I've used in this PR and i'll replace them again.

## Status

## Type of change

# How Has This Been Tested?

- Tox
- Pycharm tests
- Ran opsdroid

# Checklist:

### DIFF
--- a/opsdroid/__main__.py
+++ b/opsdroid/__main__.py
@@ -68,7 +68,7 @@ def configure_logging(config):
         file_handler.setFormatter(formatter)
         rootlogger.addHandler(file_handler)
     _LOGGER.info("="*40)
-    _LOGGER.info("Started application")
+    _LOGGER.info(_("Started application"))
 
 
 def get_logging_level(logging_level):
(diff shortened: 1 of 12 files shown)


anxodio:
Hi Fabio!

There is a problem with `logging.info(msg % arg)`, I will try to explain it. For example, in this code example:
```python
_LOGGER.debug(_("Adding database: %s" % name))
```
Let's say that `name=opsdroid`. So Python will format the string to `Adding database: opsdroid`, and then that text will be passed to `gettext` to be translated. That's a problem, because we need to translate the text with the `%s`, not with the final text.

So I think the solution is something like:
```python
_LOGGER.debug(_("Adding database: %s"), name))
```
There, Python will first call gettext with `Adding database: %s` that will be translated (for example to catalan `Afegint base de dades: %s`). Then, the string will be formatted and name will replace `%s`.

jacobtomlinson:
Thanks @anxodio that's a good point. It doesn't usually cause problems as we use `.format()` and they string is formatted after translation. But with logging you must use `%` or comma separated variables. 

@FabioRosado are you ok to update this?

FabioRosado:
Awesome thanks for the great explanation @anxodio ! I was unsure if this was the right approach that's why I asked. I'll revert the changes to follow the `_LOGGER.debug(_("Adding database: %s"), name))` style 👍

Issue #442, opened 2018-02-07
Add message.react(emoji)


jacobtomlinson:
# Description
In opsdroid/connector-slack#9 there is a request to be able to [react to messages via Slack](https://get.slack.help/hc/en-us/articles/206870317-Emoji-reactions).

Rather than implement a Slack specific feature in the connector it would be good to create a generic method on the `Message` class which can be extended by connectors to react to messages. This would allow other connectors which support reactions to implement the method too.

As reactions are mainly a slack specific feature this is a "nice to have" and shouldn't raise `NotImplementedError`s. It should return `False` and log when the method hasn't been implemented in the connector and silently fail, perhaps even be a debug level log message. It should also be noted in the documentation that it is encouraged to respond to a message if a reaction fails.

PR #459, opened 2018-02-18
add message reactions

anxodio:
# Description

This PR add reaction capabilities to opsdroid, adding a new `react` method in `Message` and `Connector`.

By default it just log a debug message and return False, but can be implemented by connectors that have react capabilities (like Slack or Github).

**Fixes** #442

## Status
**READY**

## Type of change

(By now it will be documented in the Slack connector, because it will be the only one that implements it)

# How Has This Been Tested?

I tested and checked that the new `react` methods returns False as expected.

# Checklist:

### DIFF
--- a/docs/extending/connectors.md
+++ b/docs/extending/connectors.md
@@ -19,6 +19,8 @@ Connectors are a class which extends the base opsdroid Connector. The class has 
 #### respond
 *respond* will take a Message object and return the contents to the chat service.
 
+#### react
+*react* will take a Message object and an emoji name and add this emoji as a Message reaction (if the chat service supports this feature).
 
 #### disconnect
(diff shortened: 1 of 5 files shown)


jacobtomlinson:
This looks great!

You're missing a test for the case there `thinking-delay` is set in the config. As there is no else statement you could always just add it to the config on the connector in `test_react()`.

anxodio:
Sorry, but i'm not sure of where to put this in the documentation. As I writed:
> (By now it will be documented in the Slack connector, because it will be the only one that implements it)

And the documentation is in the other PR: https://github.com/opsdroid/connector-slack/pull/10 

So, if we have to add it here, where you would add it?

FabioRosado:
Since this is a new method of the connector class I think the right place for it would be in the [connector documentation](http://opsdroid.readthedocs.io/en/stable/extending/connectors/)

Hope this helps 👍

jacobtomlinson:
Awesome many thanks!

Issue #457, opened 2018-02-14
Upgrading aiohttp beyond v3 breaks tests


jacobtomlinson:
# Description
As demonstrated in #456 upgrading aiohttp to v3 or above causes a bunch of tests to fail. 

This is likely an issue with the tests but it should be investigated to ensure there are no fundamental issues with opsdroid.

The tests need to be fixed and the version updated in a manual replacement for #456.

redfruitt:
I'll take this up

PR #462, opened 2018-02-19
Fix aiohttp tests

redfruitt:
# Description

The PR fixes the tests which were breaking post the aiohttp update from 2.3.10 to 3.0.1.

Majority tests breaking were the parsers which were using aiohttp v2's create_future helper method
to create an asyncio future and call the set_result method within the future to replace it with mock results.

The create_function helper method is no longer available within the aiohttp v3, the simplest solution seemed to create a mock future right within the test, as thats what was effectively happening with the use if create_future helper method? Please correct me if I got this wrong, I've never worked with asyncio before.

Below is how the the create_future worked in v2. 

```python
if PY_352:
    def create_future(loop):
        return loop.create_future()
else:
    def create_future(loop):  # pragma: no cover
        """Compatibility wrapper for the loop.create_future() call introduced in
        3.5.2."""
        return asyncio.Future(loop=loop)
```

I think we were passing self.loops below to get around the compatibility wrapper within the helpers module.

```python
with amock.patch('aiohttp.ClientSession.get') as patched_request:
    patched_request.return_value = helpers.create_future(self.loop)        
    patched_request.return_value.set_result(result)
    await witai.call_witai(message, config)
    self.assertTrue(patched_request.called)
```

If we replace the above with below, we don't have to worry about py3.5 compatibility issues.

```python
with amock.patch('aiohttp.ClientSession.get') as patched_request:
     patched_request.return_value = asyncio.Future()
     patched_request.return_value.set_result(result)
     await witai.call_witai(message, config)
     self.assertTrue(patched_request.called)
```

 
**Fixes** #457 

## Status
**READY** 

## Type of change

# How Has This Been Tested?

# Checklist:


jacobtomlinson:
This is awesome thanks!

It doesn't look like you've actually increased the version of `aiohttp` in this PR though? Can you do it so we can check the tests pass properly.

Also you're included `tox-generated.ini` which I think shouldn't be there.

Issue #444, opened 2018-02-08
Add no-deps


jacobtomlinson:
# Description
To speed up skill development it could be good to skip skill dependancy installation as it will likely be installed already.

Similar to `no-cache` we should add `no-deps` which defaults to `False` in the skill config.

```yaml
skills:
  - name: myawesomeskill
    no-cache: true
    no-deps: true
```

FabioRosado:
I've been checking the issues that we have open at the moment and came across this one again.

The idea of 'no-deps' would be to just  prevent opsdroid from running the method `_install_module_dependencies` right?

jacobtomlinson:
Yeah.

PR #466, opened 2018-02-21
Add no-dep flag

FabioRosado:
# Description

This adds the `no-dep` flag to opsdroid so the installation of dependencies of a specific module can be skipped. I've added the if statement on the `_install_module_dependencies()` method, but I'm not sure how to test it.

I had this version before:

```python
    def _install_module_dependencies(self, config):
        if not config.get('no-dep', False):
            if os.path.isfile(os.path.join(
                    config["install_path"], "requirements.txt")):
                self.pip_install_deps(os.path.join(config["install_path"],
                                                   "requirements.txt"))
```

And tested to see if `_install_module_dependencies` returned None when the flag `no-dep` is set to true. My tests with a mock function worked, but not with opsdroid, so I changed the implementation to just log that the installation of dependencies will be skipped.

What do you think it's the best version? 
How should I test this bit better than what I have done?

**Fixes** #444

## Status

## Type of change

# How Has This Been Tested?

- Tox (green)
- Ran tests on pycharm(green)

# Checklist:

### DIFF
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -15,6 +15,7 @@
    - [Git Repository](#git-repository)
    - [Local Directory](#local-directory)
    - [Disable Caching](#disable-caching)
+   - [Disable dependency install](#disable-dependency-install)
 - [Environment Variables](#environment-variables)
 - [Include Additional Yaml Files](#include-additional-yaml-files)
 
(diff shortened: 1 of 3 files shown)


jacobtomlinson:
This looks great. 

Maybe we should update the `_install_module_dependencies` method to return some values. Then we can test that the correct values come back.

I'm thinking
- `True` if the deps installed successfully
- `False` if they failed to install
- `None` if skipped

FabioRosado:
That sounds like a great idea. Tomorrow I should have some time to work again on this

jacobtomlinson:
I would mock the subprocess to throw the `OSError`.

FabioRosado:
I've added return values to the `_install_module_dependencies` method, the reason why I didn't return False when the installation fails is because `pip_install_deps` will raise the OSError, so I thought there was no point in adding a try/except here. Please let me know if you would rather have the try/except and the return False value there

jacobtomlinson:
Yeah you're right, raising an error is better.

jacobtomlinson:
I would be tempter to also return `None` if there is no `requirements.txt`.

jacobtomlinson:
Good work. Not sure why the coverage didn't update. Just restarted the build to try and force it.

PR #461, opened 2018-02-18
Fix reload

redfruitt:
# Description

Fixes the buggy reload feature which does not work consistently.

**Fixes** #413 

## Status
**READY** 

## Type of change

- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

# Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #471, opened 2018-02-27
Revert changes to loader.py

FabioRosado:
# Description

After the commit #461 was merged into opsdroid by me, opsdroid stopped answering to commands. 
I did some digging around and reverting the changes to `opsdroid.loader` fixed the issue.
This PR is just a quick fix to opsdroid in order to keep it working. 

I've tried different things to try and understand why the method `_reload_modules`, the thing that makes everything work seems to be the line `importlib.reload(sys.modules[module_name])` but, to be honest, I have no idea why.

Also, should I raise an issue about this in order to have this method removed?

## Status
**READY** 

## Type of change

# How Has This Been Tested?

- Ran opsdroid to see if opsdroid will respond to a command
- Tox - all green

# Checklist:

### DIFF
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -7,6 +7,7 @@ import shutil
 import subprocess
 import importlib
 import importlib.util
+from types import ModuleType
 import re
 from collections import Mapping
 import yaml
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
Yes please raise an issue. I'd rather not restore this code as it is not required any more. We should keep digging to find the root cause of the problem.

Issue #412, opened 2018-01-21
opsdroid fails to start if Rasa NLU is not available


jacobtomlinson:
# Description
When starting opsdroid with Rasa NLU configured but without Rasa NLU running a long traceback is thrown.

## Steps to Reproduce
Configure opsdroid to use Rasa NLU without running Rasa NLU. Start opsdroid.

## Expected Functionality
In the startup log there should be an error stating Rasa NLU is unavailable. Or perhaps it should block and attempt to reconnect.

## Experienced Functionality
```
INFO opsdroid.parsers.rasanlu: Starting Rasa NLU training.
DEBUG asyncio: Using selector: KqueueSelector
Traceback (most recent call last):
  File "/Users/jacob/.pyenv/versions/3.5.4/lib/python3.5/site-packages/aiohttp/connector.py", line 797, in _wrap_create_connection
    return (yield from self._loop.create_connection(*args, **kwargs))
  File "/Users/jacob/.pyenv/versions/3.5.4/lib/python3.5/asyncio/base_events.py", line 776, in create_connection
    raise exceptions[0]
  File "/Users/jacob/.pyenv/versions/3.5.4/lib/python3.5/asyncio/base_events.py", line 763, in create_connection
    yield from self.sock_connect(sock, address)
  File "/Users/jacob/.pyenv/versions/3.5.4/lib/python3.5/asyncio/selector_events.py", line 451, in sock_connect
    return (yield from fut)
  File "/Users/jacob/.pyenv/versions/3.5.4/lib/python3.5/asyncio/futures.py", line 381, in __iter__
    yield self  # This tells Task to wait for completion.
  File "/Users/jacob/.pyenv/versions/3.5.4/lib/python3.5/asyncio/tasks.py", line 310, in _wakeup
    future.result()
  File "/Users/jacob/.pyenv/versions/3.5.4/lib/python3.5/asyncio/futures.py", line 294, in result
    raise self._exception
  File "/Users/jacob/.pyenv/versions/3.5.4/lib/python3.5/asyncio/selector_events.py", line 481, in _sock_connect_cb
    raise OSError(err, 'Connect call failed %s' % (address,))
ConnectionRefusedError: [Errno 61] Connect call failed ('127.0.0.1', 5000)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/jacob/.pyenv/versions/3.5.4/lib/python3.5/runpy.py", line 193, in _run_module_as_main
[...]

PR #476, opened 2018-03-02
Fix opsdroid failing if Rasa NLU is not available

FabioRosado:
# Description

I've sent a message about this in gitter but with all the chatter I think it got lost. Anyhow, this seems to fix the issue #412 a simple try/except made opsdroid work without any problems. I'm not sure if this is the best option to fix this issue tho.

When running opsdroid with rasanlu set on the config file the debug console shows this:

```
DEBUG opsdroid.core: Parsing input: hi
DEBUG opsdroid.core: Processing parsers...
DEBUG opsdroid.core: Checking Rasa NLU...
ERROR opsdroid.parsers.rasanlu: Unable to connect to Rasa NLU
opsdroid> ERROR opsdroid.parsers.rasanlu: Rasa NLU error - No intent found. Did you forget to create one?
DEBUG root: Responding with: Hi fabiorosado
```

The reason why I decided not to create a logging function on the pass method was because of the previous logging  message when `clientOSError` is raised "Unable to connect to Rasa NLU"

It's also possible to make opsdroid work when using the except: `aiohttp.client_exceptions.ClientConnectionError` but since we are already raising ClientOSError in the parse function, I've decided to just just ClientOSError (as both except makes opsdroid work)

Would love to know your opinion on this. It will probably need a test on the try/except which I can create if needed.

**Fixes** #412

## Status

## Type of change

# How Has This Been Tested?

- Tox - all green
- Ran opsdroid - all good

# Checklist:

### DIFF
--- a/opsdroid/parsers/rasanlu.py
+++ b/opsdroid/parsers/rasanlu.py
@@ -68,12 +68,15 @@ async def _get_existing_models(config):
     """Get a list of models already trained in the Rasa NLU project."""
     project = config.get("project", RASANLU_DEFAULT_PROJECT)
     async with aiohttp.ClientSession() as session:
-        resp = await session.get(await _build_status_url(config))
-        if resp.status == 200:
-            result = await resp.json()
-            if project in result["available_projects"]:
-                project_models = result["available_projects"][project]
-                return project_models["available_models"]
+        try:
+            resp = await session.get(await _build_status_url(config))
+            if resp.status == 200:
+                result = await resp.json()
+                if project in result["available_projects"]:
+                    project_models = result["available_projects"][project]
+                    return project_models["available_models"]
+        except aiohttp.ClientOSError:
+            pass
     return []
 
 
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
Thanks for looking at this. I'm happy with your solution.

If you could add a test that would be great.

FabioRosado:
So I kind off used the example of `test_get_existing_movels` to test if the raise would be called.  The test passes but the coverage is still down, the issue might be because I'm testing if ClientOSError is raised but giving the status of `ConnectionRefusedError` as this is what we get when Rasa NLU is not available.

 Is there any better way to test an except:pass? If there is none I'll try to check with ClientOSError to see if it helps with the coverage.

Also, what's your opinion about adding a comment above the pass statement explaining why no logging message was created?

jacobtomlinson:
You need to get it to raise within the try using a side effect. Then you'll see if it calls the pass properly.

FabioRosado:
Okay that was actually easier than I expected haha I thought setting the result as ClientOSError would work but I was mistaken. Thank you for the help Jacob 👍

PR #493, opened 2018-03-19
Pass extra kwargs to connector.respond()

SolarDrew:
# Description

Allows `Message.respond()` to take extra keyword arguments which are passed to `connector.respond()`. Doesn't fix any existing issue, but is required for the Matrix connector to be able to [respond in arbitrary rooms](https://github.com/opsdroid/connector-matrix/pull/18). No dependencies are required for this change.

## Status
**READY**

## Type of change

# How Has This Been Tested?

No formal tests, but since the only change is to add optional keyword arguments, nothing else should break. Works locally with Matrix connector patched to take a room name and respond in that room.

# Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works (not necessary?)

### DIFF
--- a/docs/extending/skills.md
+++ b/docs/extending/skills.md
@@ -69,9 +69,10 @@ A _string_ containing the name of the room or chat channel the message was sent 
 
 A pointer to the opsdroid _connector object_ which receieved the message.
 
-### `respond(text)`
+### `respond(text, **connector_kwargs)`
 
-A method which responds to the message in the same room using the same connector that it was received.
+A method which responds to the received message using the same connector.
+By default the response is sent to the same room, but arguments may be passed to `connector.respond()` using `**connector_kwargs` to change this behaviour, if the connector supports this.
 
 ## Persisting data
 
--- a/docs/extending/skills.md
+++ b/docs/extending/skills.md
@@ -69,10 +69,10 @@ A _string_ containing the name of the room or chat channel the message was sent 
 
 A pointer to the opsdroid _connector object_ which receieved the message.
 
-### `respond(text, **connector_kwargs)`
+### `respond(text, room=None)`
 
 A method which responds to the received message using the same connector.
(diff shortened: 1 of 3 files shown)


jacobtomlinson:
This is interesting. Currently the way I would respond to a different room from a skill is to change the `room` property of the message before responding.

Why is that approach not suitable here?

SolarDrew:
Because it hadn't occurred to me that you could do that.....

That is a good suggestion, thanks, and it's definitely suitable. I think this approach is more generally useful though, in that it it allows more control over the behaviour of the connector, which can apply to more things than just the room.

That being the case I would lean towards this approach rather than yours, but yours does also work so I'm happy to take that route if you think this isn't suitable.

jacobtomlinson:
I am also happy with your approach as it does feel more natural, however I am concerned about connector consistency.

I don't want skill developers to have to know which connector is being used. Therefore the respond method should take a consistent set of kwargs. I'm happy for things to degrade gracefully, for example you can `react` to a message with an emoji and if the connector supports it then it will happen, if not you will get an error which you can catch and then implement a text response if appropriate.

Therefore perhaps it would be good to have an explicit `room` kwarg for `Message.respond()` rather than an arbitrary list. It could default to `None` which will respond to the same room the message came from.

SolarDrew:
Valid concerns. That did occur to me, but I wasn't really sure what the best approach would be. I like having just a `room` keyword set to `None`, as that's exactly what I've done for the Matrix connector, so that's nicely consistent.

Thanks for your input :)

jacobtomlinson:
Great! Thanks for raising the PR! 

I'm happy to merge once you've made the changes.

jacobtomlinson:
You'll need to update the [connector base class](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector.py#L68) with the `room=None` kwarg too otherwise you'll get test failures.

SolarDrew:
Ah, of course! I'll do that now.

PR #495, opened 2018-03-20
Maintainer scripts

FabioRosado:
# Description

I've noticed that the `example_configuration.yaml` as been outdated for a while so I tried to run the maintainer script, it was breaking in a lot of places so I've been fixing bit by bit and finally I was able to make it update the file.

Things that got changed:
- Had to update the path to  `configuration.j2` file - before I kept getting a templateNotFound exception when running the script from the command line and in the root directory of opsdroid.
- Added a print statement just to let the maintainer know that the script is running and not just hanging.
- Had to update the default path of  `example_configuration.yaml`, since the script was being run from the script folder the path was looking in this folder for the file `opsdroid/configuration/example_configuration.yaml`
- Changed the open mode in the function `update_config` so a maintainer could create a file elsewhere if the path was provided when running the script command.
- Updated the Readme to provide a bit more information on how to use this script - might be useful in the future for other maintainers (more documentation is good right? 😄 )

This is a short fix to the issues with the script. It probably took me way longer to fix due to running the command in pycharm and then being broken in the command line (file paths are funny).

I'm also thinking to update this script in the future to add all the new parsers to the new updated file since the script only updates skills atm. Let me know what do you think and if I should have done things different with the file paths (i'm sure there is a better way of doing with the os module).

Fixes # n/a

## Status

## Type of change

# How Has This Been Tested?

-  Successfully updated the file `example_configuration.yaml` 

# Checklist:


jacobtomlinson:
Good stuff!

PR #523, opened 2018-04-07
Update delay

### LINKED ISSUE #517: Setting typing delay to 0 means bot never responds
# Description
When a typing delay is defined it is an integer which states how many characters per second opsdroid should type. If you set this to `0` then it can type no characters per second and therefore never responds.

## Steps to Reproduce
- Configure a connector with a `typing-delay` of `0`. 
- Talk to the bot

## Experienced Functionality
The bot never responds.

## Expected Functionality
I would expect the bot to respond eventually. Perhaps even immediately and log an error stating `0` is an invalid response.

Thinking a little more about this I wonder if this functionality is the wrong way round. With `thinking-delay` that is the number of seconds that opsdroid should way before responding, so as the number gets bigger the longer opsdroid waits. But with `typing-delay` it is the number of characters per second that opsdroid can type, so the bigger the number the less opsdroid waits. The word `delay` suggests that the higher the number the longer the wait.

These opposites could be confusing (it confused me this evening). I think it would be good that both numbers represent time, rather than one representing characters per second. That would involve changing it to be seconds per character. This would result in a bigger number causing a bigger delay. 

## Versions
- **Opsdroid version:** master
- **Python version:** 3.5.4
- **OS/Docker version:** macOS 10.13

## Configuration File

```yaml
connectors:
  - name: shell
    typing-delay: 0

skills:
  - name: hello
```

FabioRosado:
# Description

This is my first attempt at fixing the issue after talking with Jacob. 

I was wondering why I implemented the typing delay how it was and finally I remembered that the reason was because I tried to implement a way to make opsdroid take longer to reply if the reply is a long sentence (as it should take longer to type).

The typing function is the same as the thinking delay, so perhaps it could be better to merge them together and get the sum of both seconds (typing/delay) and just use one function just to remove the double functions.

I would also know if should implement this issue in a different way 👍 

Fixes #517 

## Status

## Type of change

# How Has This Been Tested?

- Ran unitests in pycharm - all green
- Ran tox - all green
- Ran opsdroid with a typing delay in the connector-shell with 0 and 0.3 - both worked as expected

# Checklist:

### DIFF
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -96,7 +96,7 @@ connectors:
 Some connectors will allow you to specify a delay to simulate a real user, you just need to add the delay option under a connector in the `configuration.yaml` file.
 
 **Thinking Delay:** accepts a _int_, _float_ or a _list_ to delay reply by _x_ seconds.
-**Typing Delay:** accepts a _int_, _float_ or a _list_ to delay reply by _x_ seconds.
+**Typing Delay:** accepts a _int_, _float_ or a _list_ to delay reply by _x_ seconds - this is calculated by the length of opsdroid response text so waiting time will be variable.
 
 
 Example:
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -96,7 +96,7 @@ connectors:
 Some connectors will allow you to specify a delay to simulate a real user, you just need to add the delay option under a connector in the `configuration.yaml` file.
 
 **Thinking Delay:** accepts a _int_, _float_ or a _list_ to delay reply by _x_ seconds.
-**Typing Delay:** accepts a _int_ or _float_ that represents the number of characters typed per second. This number should be larger or equal to 6 to avoid lagging. 
+**Typing Delay:** accepts a _int_, _float_ or a _list_ to delay reply by _x_ seconds.
 
 
 Example:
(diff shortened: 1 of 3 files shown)


jacobtomlinson:
Thanks for working on this!

I agree that the delay should be longer if the response is longer. So I imagine the delay is how many seconds it takes to type each character. 

I just did a quick test online and I can type around 200 characters per minute. That is just over 3 characters per second, which is the same as saying it takes 0.3 seconds for me to type one character. So I would need to set `typing-delay` to `0.3` to simulate my typing speed.

FabioRosado:
I've just updated the code, I will need to update tests before this PR is ready.

Some notes about this update:
if the delay is under 0 seconds I assumed that the user  specified a delay for each character typed.
if the delay is over 0 seconds I assumed that the user specified a delay for the general text so I'm dividing the seconds per the total of characters * 2 - the only reason why I am multiplying the char count per 2 is because if the user specifies a one digit delay timing then opsdroid will reply almost immediatly.

Let me know what do you think about this and I'll update the tests 👍

jacobtomlinson:
This is an interesting approach, but I think users would find it confusing. Why have two different kinds of functionality depending on the value? 

Also with your logic for larger values that means long messages will have a shorter delay than short messages? If I configure it to a delay of `10` then a ten character response will be delayed by `10 / 10**2` which is `0.1` seconds. For a twenty character response it will be delayed by `10 / 20**2` which is `0.025` seconds. This doesn't really make sense as longer responses should take longer to type.

Also I think you meant greater or less than one, so decimal or not? Rather than zero which would be a negative number or positive number.

I recommend keeping it simple. The total delay should be the typing delay multiplied by the number of characters.

FabioRosado:
Updated the function and the tests, do you think its ready for merge?

jacobtomlinson:
Ah sorry you didn't comment! I must've missed it. Yes looks good to merge.

PR #529, opened 2018-04-17
Use appdir library to move config/logs

### LINKED ISSUE #513: Switch config location to use appdirs
Currently opsdroid config gets dropped into `~/.opsdroid`. This involves some manual management of this directory and isn'y very platform agnostic.

We should switch the config to use [`appdirs`](https://github.com/ActiveState/appdirs). It's a simple library which provides appropriate paths to store config in.

This change would require:
- Update the config directory to use `appdirs`
- Add some kind of flag to the CLI which tells you where your config lives
- Backward compatibility for existing setups by migrating config to the new structure

It also has support for different directories to store logs or act as a cache. Would be neat to make use of those.

FabioRosado:
# Description

This is my first PR in which I implement appdirs to opsdroid. Things I have done so far:
- Changed default config/log file to use appdirs
- check if config.yaml is in file ~/.opsdroid, copy it to appdir and remove it
- move logs to appdir logs
- create flag to edit config from command line

I'm not sure if my logic is right when doing the config bit. 

Tox is breaking on my end because it says that appdirs module could not be found, I tried to add it to both requirements.txt and requirements_test.txt but that didn't fix the issue.

What do you think of this so far Jacob? Anything I should do differently?

Fixes #513

## Status

## Type of change

# How Has This Been Tested?

- Tox - all green
- Ran opsdroid on MacOS - all working as expected
- Split configuration is imported to new dir and works as expected

# Checklist:


jacobtomlinson:
This is really good work!

I really like many of the decisions you've made here. I don't have time to fully review now but I'll look again tomorrow.

opsdroid/__main__.py
```diff
@@ -137,6 +145,10 @@ def welcome_message(config):
 @click.option('--version', '-v', is_flag=True, callback=print_version,
               expose_value=False, default=False, is_eager=True,
               help='Print the version and exit.')
+@click.option('--edit-config', '-c', is_flag=True, callback=edit_config,
```
jacobtomlinson: I love this idea! Really really useful.

We should have a look at other command line applications which have this kind of functionality and see what flags they use as I'm not 100% happy with these.

We could go for `--edit-config/-e`, which would be familiar to people who have run `crontab -e`. Or we could go for something like `--configure/-c`.

opsdroid/__main__.py
```diff
@@ -110,6 +111,13 @@ def print_example_config(ctx, param, value):
     ctx.exit(0)
 
 
+def edit_config(ctx, param, value='open'):
+    """Easy way to edit the config file."""
+    subprocess.run([value, DEFAULT_CONFIG_PATH])
```
jacobtomlinson: I think `open` is a Mac specific command so this will not work on Linux or Windows.

There is an environment variable called `EDITOR` which is used by tools like `git` and `crontab` to know how to open a file for editing. 

As opsdroid is a command line application I think we should be using the command line editor functionality. So perhaps check `EDITOR` and if it hasn't been set then run `vi`. That way users can set sublime, vscode, atom or whatever to their default editor and they should get the same functionality as with `open`.

opsdroid/const.py
```diff
@@ -1,15 +1,17 @@
...
 DEFAULT_GIT_URL = "https://github.com/opsdroid/"
 MODULES_DIRECTORY = "opsdroid-modules"
 DEFAULT_ROOT_PATH = os.path.expanduser("~/.opsdroid")
-DEFAULT_LOG_FILENAME = os.path.join(DEFAULT_ROOT_PATH, 'output.log')
+DEFAULT_LOG_FILENAME = os.path.join(user_log_dir(NAME, appauthor=False), 'output.log')
 DEFAULT_MODULES_PATH = os.path.join(DEFAULT_ROOT_PATH, "modules")
 DEFAULT_MODULE_DEPS_PATH = os.path.join(DEFAULT_ROOT_PATH, "site-packages")
-DEFAULT_CONFIG_PATH = os.path.join(DEFAULT_ROOT_PATH, "configuration.yaml")
+DEFAULT_CONFIG_PATH = os.path.join(user_config_dir(NAME, appauthor=False), "configuration.yaml")
+OLD_CONFIG_PATH = os.path.join(DEFAULT_ROOT_PATH, "configuration.yaml")
```
jacobtomlinson: As time goes by we may forget when this was added and the word `OLD` kinda loses meaning. Perhaps this should be `PRE_0_12_0_CONFIG_PATH` to make it easier to understand in the future.

What do you think?

opsdroid/loader.py
```diff
@@ -179,6 +179,13 @@ def load_config_file(self, config_paths):
                 break
 
         if not config_path:
+            if os.path.isfile(OLD_CONFIG_PATH):
```
jacobtomlinson: I would be keen to break this out into a helper function. Just because `loader.py` is already bloated.

jacobtomlinson:
The `tox` issue is due to `setup.py` importing the version constant `from opsdroid.const import __version__` which in turn tries to import `appdirs` which is not in the standard library. I think `setup.py` needs to be able to run just using the standard library.

One way to fix it would be moving `__version__` into `__init__.py` instead. Looking at [PEP 396](https://www.python.org/dev/peps/pep-0396/) it should probably be in there anyway. For backward compatibility you could always add `from opsdroid import __version__` into `const.py`, but be sure to update the `setup.py` to import from `opsdroid` and not `opsdroid.const`.

opsdroid/loader.py
FabioRosado: I created it in help first but then moved back to loader haha I'll change it back to help.
One question about this, should I check if the DEFAULT_CONFIG_PATH exists before copying the files? I'd assume an exception will be thrown if the dir doesn't exists

FabioRosado:
I've updated the PR with your recommendations. Let me know if you managed to read my question about the moving the old config to the new location.

On my end tox keeps complaining but Travis passed all the tests but linting issues so I'm happy with it.

I'm curious about one thing though. I was looking at the appdir github readme and it made me feel that only linux based systems would have a `user_config_di`r - at least its the only one with this example. If that is the case should we just move all of the files (modules and config) to `user_data_dir` what do you think?

If we go with this approach we can also cut back on a lot of variables located in `opsdroid.const.py` but I'm not sure if it's ideal to be honest.

go8ose:
At some point when Jacob was suggesting changes to a commit on this PR he wrote: 

> I think open is a Mac specific command so this will not work on Linux or Windows…perhaps check EDITOR and if it hasn't been set then run vi

On Linux vi is what we kind of expect linux sysadmins to know. But it's not very user friendly. I love it, but if you don't know what it is… you aren't going to have much fun. Perhaps if editor isn't set the script should look for vi, and pico, nano, etc., and prompt the user for which one to run. If so, the user interface chosen might look like what "update-alternatives --config editor" does on a Debian system. I'd suggest the output should also mention the EDITOR environment variable, so the user can see what they need to do to avoid making this choice each time.

jacobtomlinson:
@go8ose I totally agree and I imagine if we go with my suggestion then lots of people will be googling "how do I exit vi?". However I'm also keen to keep opsdroid consistent with other applications like git. I would expect that if a user doesn't know how to use vi they probably also don't know the difference between vi, nano, pico, etc so giving them a choice may not be helpful. I still think opsdroid should use the system default if `EDITOR` is not set. 

We could display a message explaining that the config will now be opened in an editor. They can set `EDITOR` to select which editor to use. We could also detect whether they will drop into vi and give them a heads up and a link to a vi tutorial. We could then give them the option to continue to the editor, cancel or never show this message again.

@FabioRosado I think `appdirs` should work on any platform so all of those methods should work, but as you say some may be the same directory so we should be aware of that.

Yes please check if the directory exists. Make sure the old user config cannot be removed unless it has successfully copied. You may even want to compare the new and old to be sure before deleting.

I still think the modules should be in a sub directory called `modules` and their deps should be in `site-packages`. 

Also I remember a while back we added the ability to break the config out into multiple files ([docs](https://github.com/opsdroid/opsdroid/blob/master/docs/configuration-reference.md#include-additional-yaml-files)). Have you thought about supporting users who have done that?

jacobtomlinson:
Hmm interesting. I think you want it to be an argument not a flag. But make sure you set the default value of the argument.

FabioRosado:
Changing is_flag to false did help a bit, but now if no extra arguments are passed click will just complain that it requires an argument (even if value is set up as value='vi' or default as default='vi') for some reason it just ignores the default value if none if given.

We could just ignore the default value and tell the user to specify his preferred editor but that kind of defeats the point of keeping things simple lol

jacobtomlinson:
You should be able to set `required=False`.

https://github.com/pallets/click/issues/94

FabioRosado:
I believe I might be doing something wrong, here is what the decorator looks like:
```python
@click.option('--edit-config', '-e', is_flag=False, callback=edit_config,
              required=False, default=False,
              help='Opens configuration.yaml with your favorite editor'
                   ' and exits.')
```
When I run the command `python3 -m opsdroid -e` I still get the message "Error: -e option requires an argument", from that issue that you have linked it seems that the required argument would only work with the decorator @click.argument. 

I tried to change this decorator to @click.argument but it didn't work either. I'm a bit at a loss here to be honest, perhaps I'm trying to do something that isn't intended to be done? I could just do as you suggested or in a prompt ask for which editor the user would like to edit the file and then record that... somehow o.o

FabioRosado:
Okay I'm going back to the default 'vi' editor. I was reading and it appears that windows doesn't have vi/vim installed by default - this might be a problem for people wanting to use opsdroid on windows so I will add a note on the config documentation so people know about this.

At the moment how the function is built it will look for the environment variable EDITOR and if none is found it will default to vi. Which means that the file will be open straight away.

So I was thinking to add a click.prompt asking the user if he wants to read a tutorial on vi (I have created one). So the prompt could default "NO" and the user can just press enter and edit the file straight away or choose yes and a link would show up pointing to the vim tutorial. 

What do you think of this approach? I assumed that you wont be editing the config all the time so it shouldn't be too much of an issue if the prompt is shown every time the command is run.

FabioRosado:
Here's another update for this PR. 
I have created a log flag because I assumed that with the new directory no one would really know where the log files are, this should make things easier for everyone.

I have also added a time.sleep before opening the file in vim, the only reason why I did this is because subprocess will be called straight away and the link to the tutorial won't show up until the user quits vim (which will be a bit too late).

I need help with this, should we use this approach or just use to prompt to show where the tutorial is located, then the user press enter to continue and the subprocess will run (this seems a better approach to be honest)

_Note: Click.prompt is a funny thing, if you type anything in the prompt the flag will turn to True, so in order to skip the link the user has to either type False or press enter._

I've also changed the name of the function from edit_config to edit files (but forgot to edit the docstring <.< will be updated on the next push) as the function will now open both files.

An introduction to vim was created in the tutorial section - the reason why I decided to create this documentation was due to the fact that I didn't want to use external links - the user can just read everything he needs from the opsdroid documentation.

On the `opsdroid.helper` front what I have done was search for every .yaml file, and do a route on these files: copy the file to new location, compare them, remove file from `~/.opsdroid` this should give backwards compatibility with the split config feature - I would need to do some testing before being sure it is compatible though.

**Things I still need to do**
- [ ] Double check the documentation
- [ ] Write tests for all these changes
- [ ] Test opsdroid by running it and calling skills
- [ ] Test opsdroid on my windows machine
- [ ] Test if split configuration works 
- [ ]  Test if log files are updated

opsdroid/__main__.py
```diff
@@ -111,11 +112,25 @@ def print_example_config(ctx, param, value):
...
+        file = DEFAULT_CONFIG_PATH
+    elif value == 'log':
+        file = DEFAULT_LOG_FILENAME
+    else:
         return
-    subprocess.run([os.environ.get('EDITOR', 'vi'), DEFAULT_CONFIG_PATH])
+
+    editor = os.environ.get('EDITOR', 'vi')
+    if editor == 'vi':
+        if click.prompt("You are about to open a file in vi/vim, "
```
jacobtomlinson: We could add a config option similar to `welcome-message` which allows you to disable this prompt.

jacobtomlinson:
I'm really happy with the approach you're going for. If you implement it exactly as you suggest I would be happy to merge it. This has been a pretty in depth PR and I really appreciate your hard work.

I really like the tutorial you've added to the docs. I would also be happy if you published that on a blog on Medium or something and link out to there. I would say that as the content is not related to opsdroid specifically you might want to publish it somewhere more general. I'll leave that entirely up to you but I know from experience that having a personal tech blog with content like that has been very useful in my career so far.

FabioRosado:
Thank you so much for your kind words Jacob, this PR has been quite interesting to work on to be honest and I feel a bit more confident working with click - even if there is a lot to read still.

I started my personal site in github pages and I try to write about the things I learn, not sure if its any good tho haha I could create a post about vim over there and use this tutorial then.

The tests are still failing mostly because of the `PRE_0_12_0_ROOT_PATH` I will try to sort it out when I have a go at the tests.

FabioRosado:
Okay so previous tests are currently green I still need to test that move_to_appdir function but I am unsure how to do it. Should I add params to the function in order to pass the folder "tests/configs/"? I suppose I could mock that const variable to point there but I would end up have a massive amount of mocked things so Im sure there is a better way to do it haha

jacobtomlinson:
Yeah I would make all the constants arguments to the function and pass them in when you call it in the loader. That should make testing much easier. 

I would create a temporary directory with the example config in it and another temporary directory for it to be moved to, then run the function and test that it moves correctly.

FabioRosado:
Thank you so much for the help Jacob, using the destination as params for the helper function make testing much easier. 

All tests are green but I still want to try this in my windows machine, meanwhile I got a few questions for you:

- I've replaced all imports of __version__ from opsdroid.const to opsdrois 
  - I'm importing version from opsdroid in opsdroid.const and doing `__version__ = __version__` should it just be 0.11.0?
- On the helper test should I move the files to a temp folder using `tempfile.gettempdir()`?
- locally tox is still broken and i keep getting that error unable to import click and appdirs. Should I just reclone the repo?

opsdroid/const.py
```diff
@@ -1,15 +1,22 @@
 """Constants used by OpsDroid."""
 import os
+from appdirs import user_log_dir, user_config_dir, user_data_dir
+from opsdroid import __version__
 
-__version__ = "0.11.2"
-
+__version__ = __version__
```
jacobtomlinson: This is unnecessary as line 4 already achieves this.

requirements_test.txt
```diff
@@ -9,3 +9,4 @@ pydocstyle==2.1.1
 asynctest==0.12.0
 requests_mock==1.4.0
 mypy-lang==0.5.0
+appdirs==1.4.3
```
jacobtomlinson: This is redundant as it is already in `requirements.txt`

jacobtomlinson:
I've added a couple of comments inline.

- The version should be whatever the current release is. So 0.11.0 is correct.
- Yes that's a good idea. I've done [something similar](https://github.com/opsdroid/opsdroid/pull/530/files#diff-f6247715790dd30f031c9c3d67d5814eR308) in #530.
- Try deleteing `.tox` in your working directory and run `tox` again. This should force a re-install of the dependencies.

FabioRosado:
Okay this is my final commit to this PR.
I've updated the tests to use the tempfile library, I tried to follow your example using the `with tempfile.TemporaryDirectory` but the os.mkdir wouldn't be called so instead I used this approach.

I might have gone a bit over line with all this mock.patching but coverage is happy so I'm happy as well, what do you think?

On another note, I did try this in windows but changing the directories using appdirs didn't do anything and opsdroid still hangs up when you type a command like "hello". I tried to debug but haven't figured it out why yet so I will try to spend some time on opsdroid and windows and see if I can somehow fix it but I'm not sure to be honest.

Hope everthing looks good, let me know if you need me to change anything else 👍

PR #543, opened 2018-05-25
Replace except pass with contextlib.supress

FabioRosado:
# Description

I subscribe to Dan Bader daily python tips, on one of those emails he showed how you can ignore specific exceptions in a better way than using `except <exception> : pass` I remembered that we have a few bits of code like this so I decided to try and refactor the code in order to use the suggested `with contextlib.suppress(<exception>):`

Everything seems to be working fine, but I'm not sure if this improves readability since new people might need to search what the heck does contextlib.supress means (I didn't know about it before). 

Let me know if this is good or not 👍 

Fixes #<issue>

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Tox - all green
- Ran opsdroid all good

# Checklist:

### DIFF
--- a/opsdroid/__main__.py
+++ b/opsdroid/__main__.py
@@ -6,6 +6,7 @@ import sys
 import logging
 import gettext
 import time
+import contextlib
 
 import click
 
@@ -57,11 +58,9 @@ def configure_logging(config):
     console_handler.setFormatter(formatter)
     rootlogger.addHandler(console_handler)
 
-    try:
+    with contextlib.suppress(KeyError):
         if not config["logging"]["console"]:
             console_handler.setLevel(logging.CRITICAL)
-    except KeyError:
-        pass
 
     if logfile_path:
         logdir = os.path.dirname(os.path.realpath(logfile_path))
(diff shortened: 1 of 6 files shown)


FabioRosado:
That’s a good question I assume that if we suppress Exception there python will still try to run the await skill() if an exception happens it will suppress it and run the following code which could be the next bit:
If message... logger...

I will need to give it a test tho. I’m currently on holidays but I’ll be back Friday night so I can give it a go Saturday 😀👍

FabioRosado:
I played around with this idea. If we suppress Exception then when we do this bit:
```if message:
...
```
I get the exception `TypeError: 'NoneType' object is not subscriptable` if we take this conditional then we get an AttributeError because message is None and it doesn't have the `respond` method.

I could only test this with tox because I am unable to run opsdroid through pycharm due to click giving me an encoding error (I tried to export the LC_ALL and LANG env var but didn't help - same when I add them to pycharm settings).

So I'm not sure if we supress the Exception and do a type check if message is none then get those error messages?

jacobtomlinson:
Ok no worries. Worth a try! Are you ready to merge the other changes?

FabioRosado:
I was reading the contextlib documentation and didn't find any method that could help with the broad exception as well we might need to think about something else as for merging yeah I'm happy with it 😄

Issue #195, opened 2017-07-29
Allow sending images via connectors


jacobtomlinson:
It should be possible to send images to chat services which support that. The base `Connector` class should have some kind of `send_image` method. It's default behavior can be to respond with an error saying it is not supported.

The `Message` object should have a wrapper method similar to `respond`, perhaps called `respond_with_image`.

sgreszcz:
Would be great to have this extended to any binary types: send_binary method.

Issue #536, opened 2018-05-09
Add support for non-Message events


Cadair:
Currently the main opsdroid bus only handles connectors receiving messages and passing them to skills as `Message` objects. For opsdroid to be used to do things other than a chat bot it would be good if there was support for generic event types, which could be defined on a per-connector level.

Examples of these events that have come to mind:

- A `NewChannel` event in the slack connector.
- A `RoomInvite` event in the matrix connector
- An `Issue` or `PullRequest` or `Comment` or `Review` event in the github connector (or a new GitHub App connector).

I think that the important thing with these events will be that the connector can define and emit them, and any skill can match to any event (maybe also matching on properties of event type). The simple way I have thought to do this is import the event types from the connector, but due to the way opsdroid installs it's own modules this seems to be non-trivial.

jacobtomlinson:
I agree that this would be a useful thing to add. 

This is probably also related to #195 which covers sending different types of responses other than text such as images or binaries.

From an OOP point of view I feel like a message is probably a particular type of event and so we should probably create an `Event` class which `Message` inherits from. 

We should also think about the matchers as they are all for processing `Message` objects. We could add a `match_event` matcher which takes strings like `NewChannel`, `RoomInvite`, etc as arguments. We could also need to add a `process_event` method to opsdroid which would be similar to the `parse` method. Connectors would call `process_event` when they get an event and opsdroid would then run through the skills with `match_event` and run the skill function if the event string matches.

Cadair:
> This is probably also related to #195 which covers sending different types of responses other than text such as images or binaries.

This looks like a nice way to solve this problem, `Image` could be a subclass of `Message`?

> From an OOP point of view I feel like a message is probably a particular type of event and so we should probably create an `Event` class which `Message` inherits from.

Agreed, the current `match_always` matcher then becomes `match_event('Message')`? (Where as `match_always` would be for all events? (backwards compatibility not withstanding))

> We could add a match_event matcher which takes strings like NewChannel, RoomInvite, etc as arguments.

I feel like it would be better if instead of strings it could take the actual classes, but I can't think of a nice way to get hold of the classes in the skill?

I agree with `process_event` etc.

SolarDrew:
Perhaps it could take multiple properties as a tuple only run if they all match? `or`s could be done by just specifying multiple matchers, so it wouldn't need to handle both, though I don't know if having different methods for `and` and `or` is particularly intuitive....

jacobtomlinson:
[...]

PR #559, opened 2018-06-17
Add a kwarg to respond to send an image

Cadair:
# Description

This is a very fast drive-by at #195. It adds a keyword argument to `respond` that can take an image (format up to connectors). This still requires that some text is sent as well, both to maintain compatibility with non-image supporting connectors and as alt-text.

## Type of change

### DIFF
--- a/opsdroid/message.py
+++ b/opsdroid/message.py
@@ -42,12 +42,12 @@ class Message:
 
         await asyncio.sleep(char_count*seconds)
 
-    async def respond(self, text, room=None):
+    async def respond(self, text, room=None, image=None):
         """Respond to this message using the connector it was created by."""
         opsdroid = get_opsdroid()
         response = copy(self)
         response.text = text
-
+        response.image = image
         if 'thinking-delay' in self.connector.configuration or \
            'typing-delay' in self.connector.configuration:
             await self._thinking_delay()


FabioRosado:
Hello @Cadair  and thank you for starting working on this.
I believe that the Connector class should be updated as well with a method `send_image` as suggested by Jacob, could you add this bit if you have time?

jacobtomlinson:
Thanks for raising this @Cadair. I think this opens up a lot of design questions about how the response method should work. 

For example if you are responding with an image the text arg seems unnecessary. 

Also this opens us up to adding many kwargs in the future for all the different response types we want to support. Just looking at the Telegram API it supports text, images, audio, video, documents, locations, contacts and more.

Perhaps we should think about passing a dictionary to respond which would allow flexible response types. 

We also need to think about what happens if a connector doesn't support that media type.

Cadair:
So I think the most general way to do this would probably be based around the event type stuff proposed in #536 you could construct an `Image` class and then send that object into `respond`. Connectors could then also define their own custom event types which could be imported (now connectors are moving into core) and then instantiated and sent back in `respond`.

I did this like this because I need a way to send images without refactoring the whole of opsdroid ;)

jacobtomlinson:
I like the idea of classes like `Image` inheriting from `Message`. I still want to try and keep things generic, I get the feeling you are very focused on the matrix connector. Connectors could create their own classes but then the skills become much less flexible.

We would need to add a graceful fallback if a connector doesn't support the object it has been asked to respond with. That way connector specific response objects wouldn't cause issues if used in other connectors.

I appreciate you are keen to use this feature in opsdroid, but I would prefer not to make temporary changes to the response method to satisfy a single requirement. 

I would love to see this PR updates to include the `Image` class instead of an image kwarg.

Cadair:
> I like the idea of classes like Image inheriting from Message. I still want to try and keep things generic, I get the feeling you are very focused on the matrix connector. Connectors could create their own classes but then the skills become much less flexible.

I hope I am not focusing too much on the matrix connector, the things you listed in the previous comment "text, images, audio, video, documents, locations, contacts" for telegram are pretty generic, most chat services support most of them. It makes sense to me to have these be in core.

As far as connectors implementing their own Events / Message subtypes, I was actually thinking more along the lines of a an advanced GitHub bot where you could have events for a new PR, or a new PR review etc. Skills which are built to handle messages should probably ignore these and only focus on comments on issues / prs etc but you could have a skill which sets a commit status on a `Commit` event.

While I love the generic chat bot skills in opsdroid, and would not want to reduce their utility in anyway, I have encountered some frustration with using opsdroid in a connector(s) specific manner. I would love to see opsdroid support both the generic and the specific use cases.


> We would need to add a graceful fallback if a connector doesn't support the object it has been asked to respond with. That way connector specific response objects wouldn't cause issues if used in other connectors.

I agree, which is why I kept the message argument to the `respond` function required in this PR. It makes sense to me to always provide a plain text fallback, which in the case of an image could be just the URL or a description of the image with the URL". This could obviously be a little smarter by making use of a class hierarchy. (a `Image` could be a subclass of `Attachment` so if a connector only supports generic attachments it could just handle image like that [as an arbitrary example])

I do also think it's important for connectors to just be able to drop events if they can't support them in a nice way though. For instance if a future GitHub connector implemented a `PullRequest` event, which when sent to it's `respond` method opened a PR, it's pretty meaningless in any connector other than GitHub.

> I appreciate you are keen to use this feature in opsdroid, but I would prefer not to make temporary changes to the response method to satisfy a single requirement.

Sure, I didn't really expect this PR to get merged :wink: but I wrote it as a quick hack and thought it would be a good starting point for a discussion!

> I would love to see this PR updates to include the Image class instead of an image kwarg.

Hopefully I will have a little time to keep working on this, I will use Image as a good first step.

jacobtomlinson:
Thanks for the thorough response!

> I hope I am not focusing too much on the matrix connector

I'm not worried, but it is worth bearing in mind. Particularly as the most common connector seems to be Slack.

> While I love the generic chat bot skills in opsdroid ... I have encountered some frustration

I agree. These skills are mostly to demonstrate capability, rather than for actual use. So I take your point that people will likely write skills with a particular connector in mind. However my hope is that people will share their skills on GitHub and I want to avoid there being a specific connector required to use it.

> a `Image` could be a subclass of `Attachment`

Good idea!

> It makes sense to me to always provide a plain text fallback

That makes a lot of sense. Kind of like an `alt` tag in HTML images. However I would want people to be able to omit this and the bot simply do nothing if that makes most sense in that situation.

> For instance if a future GitHub connector implemented a `PullRequest` event, which when sent to it's respond method opened a PR, it's pretty meaningless in any connector other than GitHub.

I take your point. However there is no reason why we couldn't have Bitbucket or GitLab connectors which can handle the same events. There is definitely a balance between making things generic and bloating opsdroid with lots of niche message types. One possible solution would be to include an `AbstractVCSProviderConnector` which includes these message types that the version control provider connectors can inherit from.

Cadair:
> However I would want people to be able to omit this and the bot simply do nothing if that makes most sense in that situation.

Agreed, this makes sense. I think this ties in well with the class hierarchy fall back system as well, if you try and handle an `Image` as a plain `Message` but `Message.text` is `None` then you could ignore it.

> There is definitely a balance between making things generic and bloating opsdroid with lots of niche message types.

Yes, I think as you point out the GitHub example is not really niche enough (although it could be initially developed as GH specific and then generalized into core) but I am sure there are some things which will really deserve to be connector specific.

PR #573, opened 2018-07-14
Add support for ipython notebooks as skills

### LINKED ISSUE #571: Add support for IPython/Jupyter Notebooks
I've found myself using [Jupyter Notebooks](http://jupyter.org/) more and more and I think it would be a great way to write opsdroid skills. 

The notebooks are a mixture of code, output from code (print statements, matploptlib plots, etc) and markdown. They are stored as a json blob of the cells with the file extension `.ipynb`.

There is a tool called [nbconvert](https://github.com/jupyter/nbconvert) which can convert these files into python scripts. We could wrap this in the opsdroid loader so a user could set a skill path to an `.ipynb` file and when it is loaded it will get converted into a python script and loaded like a regular single file skill.

We could also add some debugging tools which would allow us to test the skill from within the notebook.

jacobtomlinson:
# Description

This PR updates the loader so that when a single file skill is specified using the `path` option it checks whether it is an iPython Notebook and if it is it converts it into a python script.

Fixes #571

## Status
 **READY** 

## Type of change

# Checklist:

### DIFF
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -6,7 +6,7 @@
    - [Database Modules](#database-modules)
    - [Logging](#logging)
    - [Installation Path](#installation-path)
-   - [Parsers](#parsers) 
+   - [Parsers](#parsers)
    - [Skills](#skills)
    - [Time Zone](#time-zone)
    - [Language](#language)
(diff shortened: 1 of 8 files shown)


jacobtomlinson:
I'm happy with this now but If @FabioRosado could have a look over that would be great!

FabioRosado:
Sorry, Jacob, it took me so long to check this PR, I am quite happy with it. I thought it would be hard to read an IPython/notebook and use it but following your code seems pretty easy 👍 

I'll merge the PR now 😄

jacobtomlinson:
Thanks!

PR #578, opened 2018-07-16
Fix no-else-return lint issues

FabioRosado:
# Description

Pylint is complaining about the use of return statements withint if/elsif/else statements. I get it where this error comes from since in most situations you would want to do something and only at the end of a function return a value.

We use control flow to return something some of our functions:
- warning levels
- Aiohttp logic on parsers 

We could either refactor all of these functions and only use if statements ( but that seems a silly way to control things) or disable the error in the parser files and the logging levels in `__main__.py`.

Not sure what is the preferable solution here, so this PR simply disables this error all around, would like to know what do you think about this option Jacob.

Fixes #<issue>

## Status
**READY**

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/opsdroid/__main__.py
+++ b/opsdroid/__main__.py
@@ -85,11 +85,13 @@ def get_logging_level(logging_level):
     """Get the logger level based on the user configuration."""
     if logging_level == 'critical':
         return logging.CRITICAL
-    elif logging_level == 'error':
+
+    if logging_level == 'error':
         return logging.ERROR
-    elif logging_level == 'warning':
+    if logging_level == 'warning':
         return logging.WARNING
-    elif logging_level == 'debug':
+
+    if logging_level == 'debug':
         return logging.DEBUG
 
     return logging.INFO
--- a/opsdroid/parsers/rasanlu.py
+++ b/opsdroid/parsers/rasanlu.py
@@ -114,10 +114,10 @@ async def train_rasanlu(config, skills):
                              int(time_taken))
                 await _init_model(config)
                 return True
-            else:
-                _LOGGER.debug(result)
-        else:
-            _LOGGER.error(_("Bad Rasa NLU response - %s"), await resp.text())
+
+            _LOGGER.debug(result)
+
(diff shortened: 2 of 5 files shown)


jacobtomlinson:
Having a look through the failures I think they are correct. We should refactor to fix the linting problems rather than ignoring the problems.

Here's an example of what it is complaining about:

```python
if some_truthy_variable:
    return "Hello world"
else:
    return "Goodbye world"
```

In this example if `some_truthy_variable` is true it will return and never execute the else. Therefore the else is unnecessary. `pylint` is suggesting you write it like this:

```python
if some_truthy_variable:
    return "Hello world"

return "Goodbye world"
```

This would have exactly the same functionality as the first one but with less code. 

I'm pretty sure this is something which `pylint` already checks for, and looking at the failures it seems to be mostly complaining about `elif` statements so I wonder if they added better checking into the new version of `pylint`.

So the [first failure it points out](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/__main__.py#L84) looks like this:

```python
def get_logging_level(logging_level):
    """Get the logger level based on the user configuration."""
    if logging_level == 'critical':
        return logging.CRITICAL

    elif logging_level == 'error':
        return logging.ERROR

    elif logging_level == 'warning':
        return logging.WARNING

    elif logging_level == 'debug':
        return logging.DEBUG

    return logging.INFO
```

However I think it would prefer us to write it like this:

```python
def get_logging_level(logging_level):
    """Get the logger level based on the user configuration."""
    if logging_level == 'critical':
        return logging.CRITICAL

    if logging_level == 'error':
        return logging.ERROR

    if logging_level == 'warning':
        return logging.WARNING

    if logging_level == 'debug':
        return logging.DEBUG

    return logging.INFO
```

Do you mind going through and making the changes?

FabioRosado:
No worries I'll go along and change the issues so lint passes 👍

jacobtomlinson:
Awesome stuff!

PR #586, opened 2018-07-24
Revert "Add gcc to docker"

jacobtomlinson:
Reverts opsdroid/opsdroid#584

Didn't work. Reverting.


FabioRosado:
Thanks for raising this Jacob unfortunately what seems an easy fix is not haha I’m merging it now

Issue #558, opened 2018-06-17
Dependency hell: opsdroid needs aiohttp 3.3.2 and slacker-asyncio needs 3.0.2


Cadair:
I just removed slacker-asyncio from requirements.txt and it worked. :confused:

FabioRosado:
Slack is now merged into the core of opsdroid, by removing slacker-asyncio it didn't break opsdroid for you?

Cadair:
It didn't seem to. 

Why did slack get merged into core ooi?

FabioRosado:
Due to issue #185 Jacob worked and merged slack and websocket into the core, other connectors will eventually be merged as well. Opsdroid didn't drop the connector class though so users can still create their own connectors

jacobtomlinson:
We should probably update the connectors to use a more flexible requirements file rather than pinning the version. However as @FabioRosado said this should be relevant to Slack as of the next release due to it being merged into the core.

FabioRosado:
I was looking at the [skill words](https://github.com/opsdroid/skill-words) and it contains a requirement.txt file, I didn't mention anywhere that you should install the requirements manually, does opsdroid get the dependencies of the file if it's in the folder?

I'm only asking because if opsdroid does that then perhaps we could extend the functionality to connectors as well (if it's not in place yet)

jacobtomlinson:
When you configure the skill opsdroid will install the requirements file as part of installing the skill. Same for connectors and databases. Those dependencies get installed into a separate `site-packages` directory which is added to the path dynamically when the skills are loaded.

tdavis:
I ran into this earlier too. It happens if you use opscode from source to get the included connectors. I ultimately had to just back out to the pypi version and copy-paste the source for the connector I wanted to modify. The entire process of which is quite obscure. I had to read the source code to find out that if I added a `path` key to the yaml pointing to my connector it would stop trying to install it from Github. I would have liked to simply extend the opsdroid connector, but that _isn't_ published to pypi and has no `setup.py` so no dice there.

More generally, the whole module loading thing is very foreign to me as a Python developer. Why can't I install a skill or connector with `pip` (or work on one locally with the convenient `python setup.py develop`), point to the class in the yaml like `foo.bar.Baz`, and opsdroid can just `import()` it? I'm sure there's a reason for rolling your own module loading stuff, but using it to make opsdroid load my own code is not intuitive. 

In any case, thanks for your work on the project!

jacobtomlinson:
[...]

PR #587, opened 2018-07-26
Remove tests and test deps from docker and minor docker improvements.

jacobtomlinson:
# Description

This PR removes the test dependencies from the docker container and also makes some minor changes to the docker image and process.

As highlighted in #574 test dependencies can grow rapidly. We probably don't need to run the tests inside the container, but only test that the container builds and runs successfully.

I also took the opportunity to streamline the container image a little: 
- Changed the deprecated `MAINTAINER` flag to the new `LABEL` symantics.
- Added a `.dockerignore` which is simply a symlink to `.gitignore` as I noticed my local build context was >100MB, this change reduced it to <10MB. 
- Changed the pip install command to install the local module using pip rather than installing the dependencies. This results in the docker image being more representative of doing a regular pip install from pypi.
- Instead of copying `.` to the image I have cherry picked the required parts, this may help with the final image size. But probably not by a huge amount.

## Status
**READY** 

## Type of change

# Checklist:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,20 @@
 FROM python:3.5-alpine
-MAINTAINER Jacob Tomlinson <jacob@tom.linson.uk>
+LABEL maintainer="Jacob Tomlinson <jacob@tom.linson.uk>"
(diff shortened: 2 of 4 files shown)


jacobtomlinson:
This has now highlighted #558. That needs to be resolved before this will pass.

PR #589, opened 2018-07-27
Safe load 'included' yaml files

### LINKED ISSUE #580: Potential vulnerability with yaml.loader
# Description
I was listening to Talk Python to Me podcast - the one where they talk about Python vulnerabilities and things that we shouldn't be doing in Python.  One of the things that they mention was using `yaml.load` instead of `yaml.safe_load`. 

I know that when I wrote the bit of code so we can use `!include` I use yaml.loader so perhaps this should be changed to `yaml.safe_load`. One thing I didn't understand was if the `yaml.loader` is now set as the default safe_load or not. (Mostly due to [this issue](https://github.com/yaml/pyyaml/issues/5)).

I tried to swap `yaml.load` for `yaml.safe_load` but tox complained and both the environmental variables and the `!include` tests are failing.

Should I dig deeper and try to make the safe_load work?

Also, this article in hackernoon might be interesting for reference about this issue - [10 common security gotchas in Python and how to avoid them](https://hackernoon.com/10-common-security-gotchas-in-python-and-how-to-avoid-them-e19fbe265e03)

FabioRosado:
# Description

This is the first baby-step to replace `yaml.load` with `yaml.safe_load`. Currently, I have changed how the command `!include` loads a new file into the configuration - mostly because this is the first bit where people could exploit the vulnerabilities of `yaml.load`.

When I tried to replace load with safe_load on lines `loader:229-232` the texts failed as everything was returning None, so I will need to do some further testing to see why safe_load is not loading the file properly.

Another thing I need to do will be testing opsdroid with an included file and see if it works without any issue.

Fixes #580 

## Status

## Type of change

# How Has This Been Tested?

- Tox - All Green
- Travis - All Green
- !Include doesn't load python code related to the vulnerability
- Tested with an `!include file` - returns NoneType (needs further investigation)

# Checklist:

### DIFF
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -221,7 +221,7 @@ class Loader:
                                          loader.construct_scalar(node))
 
             with open(included_yaml, 'r') as included:
-                return yaml.load(included)
+                return yaml.safe_load(included)
 
         yaml.add_constructor('!envvar', envvar_constructor)
         yaml.add_constructor('!include', include_constructor)
--- a/tests/configs/exploit.yaml
+++ b/tests/configs/exploit.yaml
@@ -1,0 +1,1 @@
+!!python/object/apply:os.system ["echo 'Oops!';"]
--- a/tests/configs/include_exploit.yaml
+++ b/tests/configs/include_exploit.yaml
@@ -1,0 +1,1 @@
+!include exploit.yaml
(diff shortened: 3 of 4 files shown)


FabioRosado:
I'm wondering why `safe_load` doesn't work when we run the yaml validator method, so I'll need to investigate further to try and figure out why this happens.

As for your suggestion on creating a test to see if the exploit can be abused, sounds like a good thing to have so I'll try to write something like that, thanks for the suggestion 👍

jacobtomlinson:
These changes look good.

I'm not entirely sure what to do about the other safe load. I would recommend merging this now and then investigating it further in a new PR.

PR #597, opened 2018-07-31
Switch slack library from slacker-asyncio to aioslacker

jacobtomlinson:
Given all the problems we've had with [slacker-asyncio](https://github.com/gfreezy/slacker-asyncio) I thought I'd try out [aioslacker](https://github.com/aio-libs/aioslacker) instead. 

The main benefits are:
 - Maintained by the [aio-libs](https://github.com/aio-libs) org
 - Doesn't override the synchronous slacker module
 - Imports slacker and overrides methods which should be more future compatible
 - Has more flexible dependency specifications

Main drawbacks are:
- Looks like is hasn't been maintained as recently
- Currently uses an older version of slacker

Would be interested in thoughts from @FabioRosado and @Cadair


FabioRosado:
To be fair our slack connector just needs to get and deliver messages from a slack channel (since we aren't supporting other types of message yet) so I think the fact that aioslacker lib is using an older version os slacker might not pose too much of an issue.

I also haven't received any answer on the issue that I created from the creator of slacker-asyncio so I'm not sure if this lib is actually being maintained actively. 

For what you said about the benefits of aioslacker seems like a good way for us to mitigate the issue with the dependency hell issue and if @Cadair  had problems with slacker-asyncio overwriting slacker I'm sure someone else will have the same issue in the future.

Basically, this is a big text to say that I might be a good idea to change from slacker-asyncio to aioslacker 👍

FabioRosado:
Should we merge this and see how everything works or would you like me to give some quick testing on my end? 
Also sorry been a bit busy with visiting family and friends - plus trying to survive the 45 degrees @.@

jacobtomlinson:
I want to do a bit of manual testing. Feel free to do some too. 

Don't worry, I just try and do things when I can.

PR #604, opened 2018-08-07
WIP hangouts connector

gabeduke:
Early stages of a Google Hangouts Connector

- [x] Connect to hangups client
- [ ] Receive msgs
- [ ] Respond to msgs

To test:

Add to configuration.yaml:
```yaml
connectors:
  - name: hangups
    no-cache: true
    path: [...repository path...]/opsdroid/opsdroid/connector/hangouts
```

After starting opsdroid comes up it will ask to authenticate with a google account and then you should see all of the chat events start streaming in opsdroid debug logs.


jacobtomlinson:
Looks like you've made a good start.

Did you specifically want feedback on this PR?

It looks like you've committed your `Pipfile`. While I think they are superior to `requirements.txt` we haven't made a conscious decision to move over, so best to update the `requirements.txt` for now.

jacobtomlinson:
Ok great. 

It looks like the following tasks need to be done:
- [ ] Implement listen for messages
- [ ] Implement respond to messages
- [ ] Add docs
- [ ] Add tests

Having a look through the docs for `hangups` I would imagine you need to use [`sync_all_new_events()`](https://hangups.readthedocs.io/en/latest/developer_guide/low_level.html#hangups.Client.sync_all_new_events) within a `while True:` loop to get messages. You'll need to create `Message` objects for each new message event and send them to `opsdroid.parse()`. I also recommend adding a `await asyncio.sleep(0.5)` or similar to stop it thrashing the API.

Then to respond you'll need to take the opsdroid `Message` which is passed into the method and call [`send_chat_message()`](https://hangups.readthedocs.io/en/latest/developer_guide/low_level.html#hangups.Client.send_chat_message) to send it back to hangouts.

jacobtomlinson:
Hey @gabeduke. Are you planning on picking this up again?

gabeduke:
@jacobtomlinson let's close this. I know what is needed so if I'm able to make progress I'll push up a real pr. It's likely that this is going to be outside the bounds of my abilities but I'll keep playing around in my down time.

jacobtomlinson:
Ok sure. I would be more than happy to provide more guidance if it would be useful to you.

PR #679, opened 2018-10-01
Move the sqlite database module into core #671

### LINKED ISSUE #671: Move the sqlite database module into core
This issue covers adding the [sqlite database module](https://github.com/opsdroid/database-sqlite) to core.

## Background
A while ago we began moving connectors and databases from external plugins into the core of the project (see #185 for more context). We started with [slack](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/slack/__init__.py) and [websockets](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/websocket/__init__.py) connectors but need to go through all the other existing plugins and move them into the core.

## Steps

 - Make a new submodule directory in [`opsdroid.database`](https://github.com/opsdroid/opsdroid/tree/master/opsdroid/database) and copy the database code over.
 - Update the [`requirements.txt`](https://github.com/opsdroid/opsdroid/blob/master/requirements.txt) with any dependencies from the database if necessary.
 - Write tests for the database. (See the [Slack connector tests](https://github.com/jacobtomlinson/opsdroid/blob/master/tests/test_connector_slack.py) for inspiration).
 - Copy the relevant information from the database `README.md` into a [new documentation page](https://github.com/opsdroid/opsdroid/tree/master/docs/connectors).
 - Add the new page to the [mkdocs.yml](https://github.com/opsdroid/opsdroid/blob/master/mkdocs.yml).
 - Add to the [list of databases](https://github.com/opsdroid/opsdroid/blob/master/docs/configuration-reference.md#database-modules).
 - Add a deprecation notice to the old database module. (See [the slack connector](https://github.com/opsdroid/connector-slack))

vshelke:
# Description

Includes moving of sqlite database module into the core of opsdroid  project.

Fixes #671 

## Status
**READY**

## Type of change

# How Has This Been Tested?

- sqlite connection test
- sqlite put and get test

# Checklist:


opsdroid/database/sqlite/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+from opsdroid.database import Database
+
+
+class DatabaseSqlite(Database):
+    """A module for opsdroid to allow persist in a sqlite database."""
+
+    def __init__(self, config):
+        """Start the database connection."""
+        super().__init__(config)
+        logging.debug("Loaded sqlite database connector")
```
FabioRosado: We currently use the following logging strategy: `_LOGGER.debug` which means that you should create a `_LOGGER` variable after imports that refers back to the logging module like such: `logging.getLogger(__name__)`, this strategy helps to make logging functions more noticeable. 

We also have the possibility to translate the logging messages into other languages so the string text should be encapsulated in: `_("string")`

As an example this logging function will look like this: `_LOGGER.debug(_("Loaded sqlite database connector"))`

opsdroid/database/sqlite/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+import json
+import datetime
+import aiosqlite
+
+from opsdroid.const import DEFAULT_ROOT_PATH
+from opsdroid.database import Database
+
+
+class DatabaseSqlite(Database):
+    """A module for opsdroid to allow persist in a sqlite database."""
```
FabioRosado: We are in the process of replacing all the doc strings with the google style doc strings - reference #532 

I would like to have all the new code to be added to opsdroid core following this style, could you change the docstrings so they follow the google styling?

vshelke:
@FabioRosado In the `DatabaseSqlite` class can i store the connection object in `self.client` property ?

jacobtomlinson:
I don't think `appdirs` should behave differently. Perhaps this is because the database exists locally for you but not in Travis?

vshelke:
Yes @jacobtomlinson . 
So i checked by deleting my `sqlite.db` file locally the test environment makes a new `sqlite.db` and runs the tests successfully. 
I am wondering why travis does not create the `sqlite.db` file in `user_data_dir` on running the tests ?

jacobtomlinson:
Perhaps the `user_data_dir` also doesn't exist on Travis?

vshelke:
So how should we get the travis build passed without the `user_data_dir` ? 
Do i have to keep the `sqlite.db` file in current directory at the time of build ?

jacobtomlinson:
I think you just need to create it as part of the test setup.

jacobtomlinson:
This is looking great. There are just a few lines which aren't being tested. If you could add a few more tests to cover everything that would be great.

FabioRosado:
Thank you so much for the time working on this, I will merge it now 👍 
We give stickers for contributors, send your home address by DM to [opsdroid twitter](https://www.twitter.com/opsdroid) and we will send some to you 😄

PR #685, opened 2018-10-02
[WIP] Add Rocket.Chat Connector

### LINKED ISSUE #672: Create a Rocket.Chat connector module
Let's add a [Rocket.Chat connector](https://rocket.chat/) as proposed in #498.

## Steps

 - Make a new submodule directory in [`opsdroid.connector`](https://github.com/opsdroid/opsdroid/tree/master/opsdroid/connector) called `rocketchat` and create an `__init__.py` file inside.
 - Implement the connector according to the [documentation on writing connectors](https://opsdroid.readthedocs.io/en/stable/extending/connectors/).
 - Update the [`requirements.txt`](https://github.com/opsdroid/opsdroid/blob/master/requirements.txt) with any new dependencies from the connector if necessary.
 - Write tests for the connector. (See the [Slack connector tests](https://github.com/jacobtomlinson/opsdroid/blob/master/tests/test_connector_slack.py) for inspiration).
 - Create a [new documentation page](https://github.com/opsdroid/opsdroid/tree/master/docs/connectors).
 - Add the new page to the [mkdocs.yml](https://github.com/opsdroid/opsdroid/blob/master/mkdocs.yml).
 - Add to the [list of connectors](https://github.com/opsdroid/opsdroid/blob/master/docs/configuration-reference.md#connector-modules).

### Resources to get you started
   - [Rockat.Chat REST API](https://rocket.chat/docs/developer-guides/rest-api/)
   - [Rocket.Chat Beta Realtime API](https://rocket.chat/docs/developer-guides/realtime-api/)
   - [jadlog/rocketchat_API](https://github.com/jadolg/rocketchat_API) (bear in mind my comments in #498)

### Asyncio
 Opsdroid uses [asyncio and an event loop](https://docs.python.org/3/library/asyncio.html), so here are some things to bear in mind:
   - Functions should be coroutines which are defined with `async def myfunc(): ...`
   - Coroutines are called with `await`. For example `await myfunc()`. This is how asyncio allows many functions to take it in turns as they swap over when the current thread calls `await`.
   - **Don't block the thread!** If you need to make a call to a blocking function such as `time.sleep()` or any kind of IO request such as `requests.get()` then you should `await` an async variant. For example  you can call `await asyncio.sleep(10)` to sleep without blocking the whole application and you can use `aiohttp` as an async url library.

FabioRosado:
# Description

Initial work for adding rocket.chat connector. I have encountered a few issues with this and some similarities to the telegram connector that already exists. The documentation for Rocket.Chat is not that explicit and a lot of things were made by trial and error - and some banging my head against the wall.

At the moment opsdroid will connect to the service and complain if something goes wrong like unauthorised. Opsdroid reads messages from the chat service and replies internally (but not to the chat service as its broken at the moment).

I tried to implement the same solution as the Telegram connector although opsdroid will just run an infinite loop, replying to the last message over and over again.

I have encountered a puzzling issue with this code as well.

In postman the request `https://open.rocket.chat/api/v1/groups.history?roomName=testchannel_fabio&oldest=2018-10-02T14:10:28.549Z` will return only the latest messages from that date/time onwards, but in opsdroid I keep getting all the messages of the channel even thought the url is exactly the same as the one used in postman.

**To Do List:**
- [x] Fix coverage issues
- [x] Add tests
- [x] Update documentation
- [x] Fix `oldest` param issue
- [x] Fix respond method

Fixes #672

## Status

## Type of change

# How Has This Been Tested?

- Tox
- Ran opsdroid with Rocket.Chat connector - opsdroid replied to messages.

# Checklist:


jacobtomlinson:
I'll try and have a look when I get a chance. Might not be until later in the week. The Slack tests are definitely not perfect, so maybe once we figure this out we can make improvements there too.

FabioRosado:
No worries my work week is starting tomorrow so I might not have much time either. I might try to just implement the same solution on the other methods and see if they look okay. When I get your feedback I will change things around since every method relies on `aiohttp.ClientSession`

FabioRosado:
I followed the example of @gopalkildoliya work on the facebook connector and was able to make all the tests pass and even bump the coverage quite a bit. I am unsure how to bump the coverage further as the issue seems to be the `resp.json()` bit.

mkdocs.yml
```diff
@@ -22,6 +22,7 @@ pages:
   - 'Databases':
       - 'MongoDB': 'databases/mongo.md'
       - 'SQLite': 'databases/sqlite.md'
+      - 'Rocket.Chat': 'connectors/rocketchat.md'
```
jacobtomlinson: This should be in connectors not databases 😄

opsdroid/connector/rocketchat/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+
+        Args:
+            message (object): An instance of Message
+            room (string, optional): Name of the room to respond to.
+
+        """
+        _LOGGER.debug("Responding with: %s", message.text)
+        async with aiohttp.ClientSession() as session:
+            data = {}
+            data['channel'] = "#{}".format(self.default_room)
```
jacobtomlinson: Should this always be responding to the default room? What about the room that is passed into this method?

opsdroid/connector/rocketchat/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+        super().__init__(config)
+        self.name = "rocket.chat"
+        self.config = config
+        self.default_room = config.get("default-room", "general")
+        self.group = config.get("group", None)
+        self.url = config.get("channel-url", "https://open.rocket.chat")
+        self.update_interval = config.get("update-interval", 1)
+        self.bot_name = config.get("bot-name", "opsdroid")
+        self.listening = True
+        self.latest_update = None
```
jacobtomlinson: Should we instantiate this to now? That way if you restart opsdroid it will not respond to historic messages.

opsdroid/connector/rocketchat/__init__.py
FabioRosado: The reason why I added the default is because this is changed depending on group/channel in the config. I can try to just use the room passed to this method, cant remember if I tried it

mkdocs.yml
FabioRosado: Ooops thats right haha

opsdroid/connector/rocketchat/__init__.py
FabioRosado: How should this be done? Just to check how to best alter this

FabioRosado:
Yeah getting a list of both the channels and groups might be the best way to do this, I will update the PR tomorrow 👍

opsdroid/connector/rocketchat/__init__.py
jacobtomlinson: The room that gets passed in will be the same room you set when you construct the `Message` object. So using that should be fine.

opsdroid/connector/rocketchat/__init__.py
jacobtomlinson: Using `datetime.datetime.now().strftime('some formatting string')`. You'll need to find the right formatting string in the docs to match the format used by the API (looks like full ISO 8601 to me).

FabioRosado:
I have updated most of the things in the review, the only thing left to do is work out a best way to get the room id (because we need to make two api calls to get all of the channels/groups ids) and then make another two calls to get the history for each id (channels and groups)

jacobtomlinson:
I would make the calls to list the rooms and groups in the `connect` method and set a priodic update (60 seconds parhaps) to check if this has changed. Then every second you will need to get the history on all the channels and groups the user is a member of. This will result in quite a lot of calls, but I can't really see a way around it without some kind of websocket API being available.

FabioRosado:
I'm not sure if a lot of people use rocket.chat, but perhaps a user will have a single channel where he will interact with opsdroid - like in telegram you get the private messages to interact with the bot.  Rocket.chat has a websockets api  which I was wondering if I should use but its an unofficial release one so I decided against it.

Since we would have to do a lot of api calls to keep checking new messages on every channel, perhaps we should leave it like this and if more users end up using rocket.chat and ask for it we update the connector?

jacobtomlinson:
Yeah ok I'm fine with that. It might be worth making it clear in the documentation though.

PR #692, opened 2018-10-05
Add parameter to configure priority in every regex skill (proposal)

### LINKED ISSUE #458: Change the way opsdroid prioritize regex / NLU skills
# Description
Actually opsdroid use a score system to decide what skill have to run when more than one skills match a message.
This score system is designed to prioritize NLU engines over regex (see #311).

The problem comes when you're working at the same time with some NLU engine and some regex skills. If the regex skill have some words of some NLU intent, it will be never executed, even though you write it exactly as in the regex.

## Steps to Reproduce
For example, you add a greetings intent to some NLU training with various sentences like 'hello', 'hello opsdroid', 'hi', 'hello there', 'whats up', 'good morning'...
And you have the min-score at 0.6.

Then you have a regex matcher with something like 'hello from (?P\w+)!'.

If a user writes "hello from Barcelona!" opsdroid will prioritize the NLU response, because it's almost secure that the NLU platform will detect the intent into the sentence.

## Expected Functionality
If a developer has the regex 'hello from (?P\w+)!' and it matches, it has to be prioritized over the NLU intent, because regex it's more specific.

## Proposed solution
My proposal is redesign the priority system by specificity. So I will prioritize skills in that order:
_parse (#450) > regex > NLU engines_

And, to decide what skill to execute if more than one match in the same level:
- In parse and regex, use the expression length. Longer expression normally it's more specific.
- In NLU engines, use the engine score (like now)

Any thought or discussion will be welcome :)

anxodio:
# Description

Here I am again! Hacktoberfest motivated me :)

So I'm recovering an [old discussion](https://github.com/opsdroid/opsdroid/issues/458) that was gone stale, I read it again, and I wrote this changes that I think would make all of us happy.

The point here is change the concept of `REGEX_MAX_SCORE` to `REGEX_SCORE_FACTOR` (in fact, it was a factor all the time, not a max score), add an optional keyword argument in `match_regex` to empower the developer to change this factor, but if not provided will be the configured `REGEX_SCORE_FACTOR` (stills in 0.6).

With this change, we let the developers prioritize skills that they want to be executed over the NLU skills.

It's a proposal, so if contributors are happy with it, I will update the docs before merging the PR.

Fixes #458

## Status
**UNDER DEVELOPMENT**

## Type of change

# How Has This Been Tested?

- `TestParserRegex.test_parse_regex_priority`: it registers the same regex with different factors and ensures the skill with higher priority is executed.

# Checklist:

### DIFF
--- a/docs/matchers/regex.md
+++ b/docs/matchers/regex.md
@@ -28,7 +28,7 @@ async def hello(opsdroid, config, message):
     await message.respond('Hey')
 ```
 
-The above skill would be called on any message which matches the regex `'hi'`, `'Hi'` or `'HI`. The `case_sensitive` kwarg is optional and defaults to `True`. 
+The above skill would be called on any message which matches the regex `'hi'`, `'Hi'`, `'hI'` or `'HI'`. The `case_sensitive` kwarg is optional and defaults to `True`. 
 
 
 ## Example 2
@@ -91,3 +91,22 @@ async def remember(opsdroid, config, message):
     await opsdroid.memory.put("remember", remember)
     await message.respond("OK I'll remember that")
 ```
+
+## Score factor
+
+In order to make NLU skills execute over regex skills, opsdroid always applies a factor of `0.6` to every regex evaluated score.
+
+Sometimes can be interesting for a developer to get a regex skill executed over a NLU skill, and `score_factor` keyword argument can be used to archieve this.
+
+
+### Example 
+
+```python
+from opsdroid.matchers import match_regex
+
+@match_regex('ping', score_factor=0.9)
+async def ping(opsdroid, config, message):
+    await message.respond('pong')
+```
+
+In this example, the evaluated score of `ping` skill will be multiplied by `0.9` instead of `0.6`.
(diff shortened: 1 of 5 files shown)


jacobtomlinson:
This is really great! Thanks for getting back into it. 

I'm happy with this approach, it should help us progress with matchers like the parser one you started previously.

anxodio:
Nice! So I added some explanation in docs, please review it (my English structures can fail sometimes) :smile:

docs/matchers/regex.md
```diff
@@ -91,3 +91,22 @@ async def remember(opsdroid, config, message):
     await opsdroid.memory.put("remember", remember)
     await message.respond("OK I'll remember that")
 ```
+
+## Score factor
+
+In order to make NLU skills execute over regex skills, opsdroid always applies a factor of `0.6` to every regex evaluated score.
+
+Sometimes can be interesting for a developer to get a regex skill executed over a NLU skill, and `score_factor` keyword argument can be used to archieve this.
```
FabioRosado: I would restructure this sentence like this:
"If a developer want to have a regex skill executed over a NLU one then the keyword argument `score_factor`  can be used to achieve this."

FabioRosado:
Hey @anxodio  thanks for updating the documentation so quickly. I am going to merge this PR now 😄 👍

Issue #709, opened 2018-10-23
Skills are being loaded twice


lethonomia:
# Description
Thanks for all the hard work on this. I'm testing out the framework and hope to be able to contribute back into it in the near future. 

I was investigating an issue with a cron skill running twice. I believe its related to the modules being loaded twice. Most don't trigger twice because of rank order. The easiest way to see this is to load the help module and run it. I altered the help command to dump the opsdroid.skill object and you can see two memory addresses for each module. 

{'regex': {'expression': 'when did you last see (?P<name>.*)\\?', 'case_sensitive': True}, 'skill': <function last_seen at 0x7f91386e9d90>, 'config': {'name': 'seen', 'type': 'skill', 'is_builtin': False, 'module_path': 'opsdroid-modules.skill.seen', 'install_path': '/home/atc/skills-cache/opsdroid-modules/skill/seen', 'branch': 'master'}}, {'always': True, 'skill': <function update_seen at 0x7f91386e9d08>, 'config': {'name': 'seen', 'type': 'skill', 'is_builtin': False, 'module_path': 'opsdroid-modules.skill.seen', 'install_path': '/home/atc/skills-cache/opsdroid-modules/skill/seen', 'branch': 'master'}}, 

{'regex': {'expression': 'when did you last see (?P<name>.*)\\?', 'case_sensitive': True}, 'skill': <function last_seen at 0x7f91386e9ea0>, 'config': {'name': 'seen', 'type': 'skill', 'is_builtin': False, 'module_path': 'opsdroid-modules.skill.seen', 'install_path': '/home/atc/skills-cache/opsdroid-modules/skill/seen', 'branch': 'master'}}, {'always': True, 'skill': <function update_seen at 0x7f91386e9e18>, 'config': {'name': 'seen', 'type': 'skill', 'is_builtin': False, 'module_path': 'opsdroid-modules.skill.seen', 'install_path': '/home/atc/skills-cache/opsdroid-modules/skill/seen', 'branch': 'master'}}

## Steps to Reproduce
load help and issue the help command. I added

`skills = str(opsdroid.skills)
 logging.debug(skills)`

to get the skills object easily 

## Expected Functionality
Each module/skill should only be loaded once. 

## Experienced Functionality
Modules are loaded into memory twice and are available to be run.

(atc) [atc@95672318d4ad ~]$ opsdroid
INFO opsdroid: ========================================
INFO opsdroid: Started application
INFO opsdroid: ========================================
[...]

Issue #655, opened 2018-09-25
Reload skills


colonelkrud:
Perhaps as part of #643  we should rebuild the reload function. 

When developing, sometimes I want to reload a skill many times to text each part of functionality. Currently the only way to reload skills is to restart the bot. My suggestion is a new method of loading skills that allows for a single skill or all skills to be reloaded while the bot is live. It could even recheck the config file and redownload non local skills when no single skill is named - meaning skills could be added or removed while the bot is running. This should be open to the skills themselves as a function of the core such that it can be called from within a skill. 

An example would be a dev skill which listens for various commands in a dev room, checks the user is on an approved user list, then runs the reload <config|(skill name)>. This would allow for rapid development and testing of skills.

FabioRosado:
I believe the issue with the reload and why we decided to remove it was because it caused some issues with the loader.

colonelkrud:
True.. but if we are looking at improving the loader, this should be something considered from the beginning.

jacobtomlinson:
The reload was removed because it was poorly implemented (by me) and caused more problems than it solved, so the decision was to take it back out again. However I would like to see this functionality in opsdroid, just done properly 😂.

One package to look at for inspiration would be [appdaemon](https://github.com/home-assistant/appdaemon). They have auto loading and 'auto reload on file changed' for their plugins.

Issue #643, opened 2018-09-13
Improve loader.py


FabioRosado:
# Description
The `loader.py` file is in need of refactoring as at the moment it is a bit of a mess. I remember from a few chats in issues/gitter/PRs that a few things that should be improved. This issue is just to formalize that need.

The loader could probably be refactored in order to make it fast. In previous versions of opsdroid, the loader would kick in and in seconds opsdroid was working, now it takes a bit to load - I'm not sure why this happens at the moment so it should be investigated.

I also remember Jacob mention that we should probably grab all the skills and just do one single check if any of them are out of data - and do the update from github at once.

@jacobtomlinson could you check this issue and give more feedback on to what else should be changed on loader? I will write up a list of things once I have your feedback 👍

jacobtomlinson:
I was going to tackle this in #711 but that PR has gotten so big already that I'm just going to put some thoughts here.

I think it would be a good idea to split the `Loader` class into `Loader` and `Installer` to simplify the class. `Loader` should handle loading config, loading modules and installing modules using `Installer`. I think `Installer` should not be instantiated but simple have class methods and static methods. 

We should also check that modules specified with `path` or `repo` in the config override the built in modules.

Issue #247, opened 2017-09-30
SIGINT (Ctrl+C) should be handled by opsdroid


gtseres:
When executing opsdroid in Ubuntu or in a Docker container, the following message appears:
`(Press CTRL+C to quit)`

When hitting `Ctrl+C` though, the following traceback is produced:

```
^CERROR asyncio: Task exception was never retrieved
future: <Task finished coro=<ConnectorShell.listen() done, defined at /root/.opsdroid/modules/opsdroid-modules/connector/shell/__init__.py:57> exception=ConnectionResetError('Connection lost',)>
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/asyncio/tasks.py", line 240, in _step
    result = coro.send(None)
  File "/root/.opsdroid/modules/opsdroid-modules/connector/shell/__init__.py", line 63, in listen
    user_input = await async_input('', opsdroid.eventloop)
  File "/root/.opsdroid/modules/opsdroid-modules/connector/shell/__init__.py", line 37, in async_input
    await writer.drain()
  File "/usr/local/lib/python3.5/asyncio/streams.py", line 333, in drain
    yield from self._protocol._drain_helper()
  File "/usr/local/lib/python3.5/asyncio/streams.py", line 204, in _drain_helper
    raise ConnectionResetError('Connection lost')
ConnectionResetError: Connection lost
ERROR asyncio: Exception in default exception handler
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/asyncio/base_events.py", line 1284, in call_exception_handler
    self.default_exception_handler(context)
  File "/usr/local/lib/python3.5/asyncio/base_events.py", line 1259, in default_exception_handler
    logger.error('\n'.join(log_lines), exc_info=exc_info)
  File "/usr/local/lib/python3.5/logging/__init__.py", line 1314, in error
    self._log(ERROR, msg, args, **kwargs)
  File "/usr/local/lib/python3.5/logging/__init__.py", line 1421, in _log
    self.handle(record)
  File "/usr/local/lib/python3.5/logging/__init__.py", line 1431, in handle
    self.callHandlers(record)
  File "/usr/local/lib/python3.5/logging/__init__.py", line 1493, in callHandlers
    hdlr.handle(record)
  File "/usr/local/lib/python3.5/logging/__init__.py", line 861, in handle
    self.emit(record)
  File "/usr/local/lib/python3.5/logging/__init__.py", line 1053, in emit
    self.stream = self._open()
[...]

PR #711, opened 2018-10-25
[WIP] Refactor skill discovery

### LINKED ISSUE #683: Update plugin discovery
Currently when you decorate a skill with a matcher it registers itself with opsdroid. It has become aparent for a couple of reasons that this isn't ideal:
 - In order for a matcher to self register opsdroid needs to attach a pointer to the main instance of opsdroid to an attribute of the core opsdroid class. This is a hack and should be avoided.
 - To add support for context opsdroid needs to be aware of all the constraints attached to a skill as well as the matchers. Currently if there are two matchers it will create two separate skills in opsdroid with pointers to the same function. Really we want one skill with all the decorated attributes.

My plan for implementation:
 - Update matchers to add attributes to skills rather than registering them.
 - Add a discovery section to the loader after importing the skills. This will search through all the functions in the imported module and look for the attributes, if it finds them it will register them at that point.
 - Remove the instance pointers from the opsdroid class as this is no longer necessary.

jacobtomlinson:
# Description
This is going to be quite a major PR where I am rewriting the way modules are loaded. This is partly in preparation for adding constraints and context and partly to tackle a number of issues.

Changes in this PR:
 - [x] Change the main entry point to simplify the opsdroid context manager
 - [x] Update matchers to annotate skill functions instead of registering them with opsdroid
 - [x] Update the opsdroid setup to discover annotated functions from loaded modules
 - [x] Update aiohttp server code to resolve deprecation warnings
 - [ ] Split `Loader` into `Loader` and `Installer` to simplify the class
 - [ ] Add unload function which reverses what `opsdroid.load` does
 - [ ] Reinstate the reload functionality

Fixes #683 

Also addresses #709, #655, #643, 

## Status
**UNDER DEVELOPMENT**

# How Has This Been Tested?

Tests have been added and updated as the changes are being made.

# Checklist:

### DIFF
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -13,6 +13,7 @@ from opsdroid.memory import Memory
 from opsdroid.connector import Connector
 from opsdroid.database import Database
 from opsdroid.loader import Loader
+from opsdroid.web import Web
 from opsdroid.parsers.always import parse_always
 from opsdroid.parsers.regex import parse_regex
 from opsdroid.parsers.dialogflow import parse_dialogflow
@@ -21,7 +22,6 @@ from opsdroid.parsers.recastai import parse_recastai
 from opsdroid.parsers.witai import parse_witai
 from opsdroid.parsers.rasanlu import parse_rasanlu, train_rasanlu
 from opsdroid.parsers.crontab import parse_crontab
-from opsdroid.const import DEFAULT_CONFIG_PATH
 
 
 _LOGGER = logging.getLogger(__name__)
@@ -47,6 +47,7 @@ class OpsDroid():
                 self.eventloop.add_signal_handler(sig, self.call_stop)
         self.skills = []
         self.memory = Memory()
+        self.modules = {}
         self.loader = Loader(self)
         self.config = {}
         self.stats = {
@@ -55,7 +56,7 @@ class OpsDroid():
             "total_response_time": 0,
             "total_responses": 0,
         }
-        self.web_server = None
+        self.web_server = Web(self)
         self.stored_path = []
 
     def __enter__(self):
@@ -117,24 +118,20 @@ class OpsDroid():
         print('')  # Prints a character return for return to shell
         _LOGGER.info(_("Keyboard interrupt, exiting."))
 
-    def load(self):
-        """Load configuration."""
-        self.config = self.loader.load_config_file([
-            "configuration.yaml",
-            DEFAULT_CONFIG_PATH,
-            "/etc/opsdroid/configuration.yaml"
-            ])
+    def load(self, config):
+        """Load modules."""
+        self.config = config
+        self.modules = self.loader.load_modules_from_config(self.config)
+        _LOGGER.debug(_("Loaded %i skills"), len(self.modules["skills"]))
+        self.setup_skills(self.modules["skills"])
+        self.setup_webhooks()
+        self.train_parsers(self.modules["skills"])
 
     def start_loop(self):
         """Start the event loop."""
-        connectors, databases, skills = \
-            self.loader.load_modules_from_config(self.config)
-        _LOGGER.debug(_("Loaded %i skills"), len(skills))
-        if databases is not None:
-            self.start_databases(databases)
-        self.setup_skills(skills)
-        self.train_parsers(skills)
-        self.start_connector_tasks(connectors)
+        if self.modules["databases"] is not None:
+            self.start_databases(self.modules["databases"])
+        self.start_connector_tasks(self.modules["connectors"])
         self.eventloop.create_task(parse_crontab(self))
         self.web_server.start()
         try:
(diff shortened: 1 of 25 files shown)


FabioRosado:
This is a pretty awesome job and can't imagine how long it must had taken you to rewrite some of the bits in the code.

I had a quick look on the PR and things seem okay, a few things I have notice is when we use the `_(string to be translated)` in the logging and in some places this isn't used - but we haven't yet received any info about how useful this is as we spoke in gitter.

I want to give this PR a try on my end and have a look at it better again when I'm not tired from a 13 hour day work haha

I will be free this Tuesday and I'll review it properly then if you are okay with it?

jacobtomlinson:
Yeah it has taken a while 😆. Good spot with the logging strings, I'll get them updated.

If you could take a good look on Tuesday that would be great.

opsdroid/core.py
```diff
@@ -97,54 +107,95 @@ def critical(self, error, code):
...
-    def call_stop(self):
-        """Signal handler to call disconnect and stop."""
-        future = asyncio.ensure_future(self.disconnect())
-        future.add_done_callback(self.stop)
-        return future
+    @staticmethod
+    def handle_async_exception(loop, context):
+        """Handle exceptions from async coroutines."""
+        _LOGGER.error("Caught exception")
+        _LOGGER.error(context)
```
FabioRosado: Do you think this logging error should be added into the same one?

Also, when I kill opsdroid this is being triggered by the `connector.listen` method (I'm assuming due to the infinite loop)

```
INFO opsdroid.core: Exiting application with return code 0
DEBUG asyncio: Using selector: KqueueSelector
ERROR opsdroid.core: Caught exception
ERROR opsdroid.core: {'message': 'Task was destroyed but it is pending!', 'task': <Task pending coro=<RocketChat.listen() running at /Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/connector/rocketchat/__init__.py:165> wait_for=<Future pending cb=[BaseSelectorEventLoop._sock_connect_done(10)(), <TaskWakeupMethWrapper object at 0x10e7a98e8>()]> cb=[gather.<locals>._done_callback(2)() at /usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/asyncio/tasks.py:609]>}
ERROR opsdroid.core: Caught exception
ERROR opsdroid.core: {'message': 'Task was destroyed but it is pending!', 'task': <Task pending coro=<ConnectorShell.listen() running at /Users/fabiorosado/Library/Application Support/opsdroid/opsdroid-modules/connector/shell/__init__.py:70> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x10e7a9978>()]> cb=[gather.<locals>._done_callback(3)() at /usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/asyncio/tasks.py:609]>}
```


This is probbaly outside the scope of this PR and more related to #247

opsdroid/core.py
```diff
@@ -97,54 +107,95 @@ def critical(self, error, code):
...
+    def run(self):
+        """Start the event loop."""
+        _LOGGER.info(_("Opsdroid is now running, press ctrl+c to exit."))
+        if not self.is_running():
+            self._running = True
+            while self.is_running():
+                pending = asyncio.Task.all_tasks()
+                try:
+                    self.eventloop.run_until_complete(asyncio.gather(*pending))
+                except asyncio.CancelledError:
```
FabioRosado: Could this be replaced with `with contextlib.suppress(asyncio.CancelledError)`?

opsdroid/core.py
```diff
@@ -97,54 +107,95 @@ def critical(self, error, code):
...
+        for skill in self.skills:
+            _LOGGER.info(_("Removed %s"), skill.config['name'])
+            self.skills.remove(skill)
 
-    async def disconnect(self):
-        """Disconnect all the connectors."""
         for connector in self.connectors:
+            _LOGGER.info(_("Stopping connector %s..."), connector.name)
             await connector.disconnect(self)
+            self.connectors.remove(connector)
```
FabioRosado: _Note: This is a thought._
Perhaps we could somehow change the infinite loop flag from `connector.listen` to false to avoid the exception `Task pending coro=<ConnectorShell.listen() running at (...)`? Could perhaps a simple connector.is_listening = False work?

opsdroid/core.py
```diff
@@ -156,6 +207,11 @@ def setup_skills(self, skills):
...
 
         """
+        for skill in skills:
+            for _, func in skill["module"].__dict__.items():
+                if hasattr(func, "skill"):
+                    func.config = skill['config']
+                    self.skills.append(func)
         with contextlib.suppress(AttributeError):
             for skill in skills:
                 skill["module"].setup(self, self.config)
```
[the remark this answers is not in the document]
jacobtomlinson: What skills do you have configured? There was a breaking change introduced in #700 which passes config to the setup too.

opsdroid/core.py
jacobtomlinson: The two connectors which are complaining `rocketchat` and `shell` are not honoring the `disconnect` method. Therefore I would expect to see this error. 

Happy to make it one line though.

opsdroid/core.py
jacobtomlinson: Probably :)

opsdroid/core.py
jacobtomlinson: The shell connector just needs to implement the disconnect method correctly.

jacobtomlinson:
Thanks for reviewing this @FabioRosado. I'll take a look at it this week and update it with your comments. I replied to your in line comments. Happy to continue discussing those.

opsdroid/core.py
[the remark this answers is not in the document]
FabioRosado: I did pull the changes from master before pulling this branch so it should be working okay. If you can run opsdroid fine then it must be an issue on my end. The skills that I currently have to try this branch is just the `hello` skill

opsdroid/core.py
FabioRosado: I gave this another thought and I think two line its fine as it makes it easier to fix the issue - it might add a massive block of text is both of them are in one line.  I also looked at rocket.chat and I kind off forgot to add that haha

FabioRosado:
I have replied to you in the comments as well, everything seems okay and thanks for shinning some light in the issues that I have found with the connectors

Issue #719, opened 2018-11-01
A Proposal for Handling Generic Events


Cadair:
This is a proposal on how to resolve #536. It came from a discussion with @SolarDrew 

## Event Classes

Currently opsdroid provides a `Message` class which is designed to represent a simple textual message. This proposal is for a hierarchy of different types of events decended from an `Event` abstract class.

For example a simple inheritance chain might be `Event` > `Message` > `Image`.

Other types of events may include `RoomCreation` `RoomInvite` or `PullRequest`.

### Connector Specific Event Classes

To fully support the features of all the disperate platforms, I anticipate that connectors might want to implement derived event specific to their platform. Although I hope this might be rather rare. One example I could think of is `MatrixStateEvent` for arbitrary state keys in matrix rooms. (I.e. what the
matrix state database plugin uses.)

An issue I can not see a nice solution to is now to import these classes into skills if the connectors are loaded via the plugin system.

# Matchers and Events

If a connector emits an event of type `Image` a skill may want to handle it differently to an event of type `Message`. Therefore an event type matcher should be provided. Given that `Image` may be a subclass of `Message` the matcher which is the most specialised (longest walk down the MRU?) should be matched.

I am strongly in favor of having the skills import the event types from opsdroid and pass them to `@event_matcher` rather than using strings or other way of specifying type.

I also suggest that for the first version of this we only match on event type and not other properties.

### Connector Specific Event Matching

It is possible a skill will want to only match `PullRequest` or even `Message` events emitted by a specific connector. To facilitate this I propose that on creation of an event a `source` parameter is given, which would normally take the name of the emitting connector. This should allow both matching on connector while retaining a more generic event hierarchy in opsdroid core.

## Responding to Events

Currently messages are responded to by calling `Message.respond` which in turn calls `connector.respond`. We propose that a new method `opsdroid.send` be implemented for dispatching events. This method would have the following signature: (type hints for API design only)

```python
def send(event: Event, connector: str, target: str)
```

This method would be called by `Event.respond` which would pre-populate the `connector` and `target` keyword arguments based on the properties given when it was created. (Respond here meaning "respond in the same room (and therefore connector) as the original message).

[...]

Issue #446, opened 2018-02-09
Add helper method for getting entities from a message


jacobtomlinson:
# Description
Depending on which matcher is used to match a skill the `message` object has additional attributes containing the raw response from the parser.

For example if matched with `match_regex` then the `re.match` object is available at `message.regex`, if matched with `match_rasanlu` then the response object from the Rasa NLU api is available at `message.rasanlu`, etc. 

This means that if you use multiple matchers on your skill then you need to include a ton of code checking which matchers was used and then extract entities in different ways before doing anything else. You also end up duplicating that code frequently.

Example with three matchers

```python
@match_regex(r'my name is (?P<name>\w+)')
@match_rasanlu('smalltalk.name')
@match_witai('smalltalk.name')
async def my_name_is(opsdroid, config, message):
    entity_name = 'name'
    name = None
    if hasattr(message, 'regex'):
        name = message.regex.group(entity_name)
    if hasattr(message, 'rasanlu'):
        [name] = [entity["value"] for entity in message.rasanlu["entities"] if entity["entity"] == entity_name]
    if hasattr(message, 'witai'):
        name = message.witai["entities"][entity_name][0]["value"]
    if name is None:
        return False
    await message.respond("Hello {name}".format(name=name))
```

As all language matchers have the capacity to return named entities we should add a `entities` `@property` to the `Message` class which basically wraps up the if statements from the above example (along with more to cover all matchers) and extracts all entities into a dictionary.

New example

```python
@match_regex(r'my name is (?P<name>\w+)')
@match_rasanlu('smalltalk.name')
@match_witai('smalltalk.name')
async def my_name_is(opsdroid, config, message):
    entity_name = 'name'
    if entity_name not in message.entities
        return False
    await message.respond("Hello {name}".format(name=message.entities[entity_name]))
```

jacobtomlinson:
@redfruitt have you made any progress on this?

redfruitt:
not much yet, will buck up this weekend.

redfruitt:
Hey @jacobtomlinson, wanted to know your thoughts on how best setup this @property within the Message class, I'm stuck a little with the regex clause.
And do you see any problems with the how I wrote rasa's entity fetch?

```python
    @property
    def entities(self):
        """Store entities created via various parsers."""
        entities = {}

        if self.regex:
            entities = [m.groupdict() for m in self.regex.finditer(self.text)]

        elif rasanlu:
            entities = rasanlu["entities"]
        return entities

[...]

PR #721, opened 2018-11-02
[WIP] First pass at `Event`s

### LINKED ISSUE #195: Allow sending images via connectors

### LINKED ISSUE #536: Add support for non-Message events

SolarDrew:
# Description

First pass at replacing the current `Message` class with a heirarchy of events inheriting from a generic  base `Event` class. Motivated by discussions at #719.

Fixes #536 
Fixes #195 

## Status
**UNDER DEVELOPMENT**

## Type of change

# How Has This Been Tested?

No new tests as yet, but existing tests still pass.

# Checklist:

- [ ] New and existing unit tests pass locally with my changes (they don't at the moment without my changes either, so I need to look into that...)


opsdroid/events.py
```diff
@@ -115,3 +123,13 @@ def __init__(self, text, user, room, connector, raw_message=None):
         if 'thinking-delay' in self.connector.configuration:
             await self._thinking_delay()
         return await self.connector.react(self, emoji)
+
+
+class Image(Message):
+    def __init__(self, image_url, user, room, connector):
+        self.created = datetime.now()
+        self.url = image_url
```
jacobtomlinson: We need to think carefully about this. Some connectors will want a url, some will want the image itself probably as a binary or base64 encoding.

opsdroid/events.py
```diff
@@ -115,3 +123,13 @@ def __init__(self, text, user, room, connector, raw_message=None):
         if 'thinking-delay' in self.connector.configuration:
             await self._thinking_delay()
         return await self.connector.react(self, emoji)
+
+
+class Image(Message):
+    def __init__(self, image_url, user, room, connector):
+        self.created = datetime.now()
```
jacobtomlinson: This is a godo idea. We should move this up into the `Event` as it makes sense to have it in everything.

opsdroid/events.py
```diff
@@ -82,8 +87,11 @@ def __init__(self, text, user, room, connector, raw_message=None):
         Logs response and response time in OpsDroid object stats.
         """
         opsdroid = get_opsdroid()
-        response = copy(self)
-        response.text = text
+        if isinstance(response_event, str):
+            response = copy(self)
```
Cadair: I think this should be `response = Message(response_event)` no? Are we not going to assume that a textual response is always a `Message`?

SolarDrew:
Yes, I think that was what I was going for in a more round about way.

opsdroid/events.py
SolarDrew: Agreed. So far I've just made `Event` basically empty and I figure we'll move overlapping functionality into it as we find it. So please keep pointing it out 😃

opsdroid/events.py
Cadair: It would be useful to have a method on the image connector to download a URL. i.e. convert a URL to a bytes string, would save all the connectors implementing it themselves.

SolarDrew: Would that not be an issue for the connector to deal with? I feel like the most sensible thing for the event to do is use a single general representation of the image and then each connector will have to decide the best way to use that, which could include going to the URL and downloading the image.

opsdroid/events.py
SolarDrew: Yes, I think that was what I was going for in a more round about way.

opsdroid/events.py
jacobtomlinson: I can think of many examples where there is no url. What if the bot generates a plot with matplotlib for example? Or a QR code with [qrcode](https://pypi.org/project/qrcode/). These will start as image objects in Python.

SolarDrew:
Nooooooooo! :'(

jacobtomlinson:
Let's give it another month before the bot closes it again :)

SolarDrew:
Yaaay! :D Thanks @jacobtomlinson!

SolarDrew:
FYI, I haven't forgotten about this. But I'd like to have the matrix connector in core before I do too much more on it.

jacobtomlinson:
Still pending on #731

opsdroid/events.py
Cadair: Having discussed this with @SolarDrew today, I think the base representation for a `File` should be a bytes object, this is the only way we can always represent an image.

I think that you should be able to construct an image with either a URL or a bytes object, and then the URL representation and the bytes representation should be accessible as properties, with the understanding that the URL can be `None`.

opsdroid/events.py
jacobtomlinson: If you go with this approach be sure to copy the other information like connector and room from `self`. The `copy` method is being used here to do that in a lazy way.

opsdroid/events.py
jacobtomlinson: This definitely highlights one of the challenges of opsdroid. It is trying to support "all" chat clients, and so we are going to get issues like this. 

I would be happy for url to default to `None` and have a property for the bytes object which can download the url if accessed. This would result in clients which only accept urls not be able to take bytes only images. But I think thats fine, I see no solution without hosting an image CDN 😆.

opsdroid/events.py
```diff
@@ -115,3 +123,13 @@ def __init__(self, text, user, room, connector, raw_message=None):
...
+
+
+class Image(Message):
+    def __init__(self, image_url, user, room, connector):
+        self.created = datetime.now()
+        self.url = image_url
+        self.user = user
+        self.room = room
+        self.connector = connector
+        self.responded_to = False
```
jacobtomlinson: We should bump this up to the event object.

opsdroid/events.py
jacobtomlinson: Gitter?

Cadair: Really??? 🤦🏾‍♀️

opsdroid/events.py
```diff
@@ -115,3 +123,13 @@ def __init__(self, text, user, room, connector, raw_message=None):
         if 'thinking-delay' in self.connector.configuration:
             await self._thinking_delay()
         return await self.connector.react(self, emoji)
+
+
+class Image(Message):
```
Cadair: I think `Image` should be a subclass of `File`

tests/test_connector_facebook.py
```diff
@@ -7,8 +7,7 @@
 from opsdroid.__main__ import configure_lang
 from opsdroid.core import OpsDroid
 from opsdroid.connector.facebook import ConnectorFacebook
-from opsdroid.message import Message
-from opsdroid.__main__ import configure_lang
+from opsdroid.events import Message
```
Cadair: Are the changes in this file a rebase fail?

opsdroid/events.py
Cadair: Exasperation at Gitter aside, opsdroid can only abstract away the differences between chat platforms so far. I think this is just the limit of how far we can go?

opsdroid/events.py
SolarDrew: Agreed, that change will be in the next push.

tests/test_connector_facebook.py
SolarDrew: Very likely. I've actually also realised that it may be necessary to change the ordering of arguments in the `Event` subclasses' call signatures, so everything else should carry on using `message.Message` for now anyway because otherwise stuff will break.

jacobtomlinson:
Could someone please give me a status update on this PR? I'm keen to get it merged as it opens up some awesome new functionality. I'm planning on doing a release asap it it would be great to get this into it.

SolarDrew:
At this point I think all the functionality is there, so it's just waiting on tests passing and proper coverage again. I've got a coupld of long train journeys coming up this week so I'll see if I can finish this off then.

jacobtomlinson:
Awesome! I'll aim to do a release on the weekend then so if you can get it ready before that would be excellent.

opsdroid/events.py
```diff
@@ -0,0 +1,145 @@
...
+        raw_message: Raw message as provided by chat service. None by default
+
+    Attributes:
+        created: Local date and time that message object was created
+        user: String name of user sending message
+        room: String name of the room or chat channel in which message was sent
+        connector: Connector object used to interact with given chat service
+        text: Text of message as string
+        raw_message: Raw message provided by chat service
+        regex: A re match object for the regular expression message was matched
```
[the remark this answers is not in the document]
SolarDrew: Ok, no problem. Perhaps changing it to something generic like `match`?

SolarDrew:
> Thanks for adding the deprecation warning. It would be good to ensure that opsdroid core doesn't trigger this warning itself. This means all the connectors will need updating. Could you please do that too?

Damnit I did that and then changed it back... :P No problem, I can change it back again. I just hesitated because I thought it might be easier for the maintainers of those connectors to switch over gradually, rather than me going in and changing their stuff.

opsdroid/events.py
```diff
@@ -0,0 +1,166 @@
...
+        user: String name of user sending event
+        room: String name of the room or chat channel in which event was sent
+        connector: Connector object used to interact with given chat service
+
+    Attributes:
+        created: Local date and time that message object was created
+        user: String name of user sending message
+        room: String name of the room or chat channel in which message was sent
+        connector: Connector object used to interact with given chat service
+        responded_to: Boolean initialized as False. True if message has been
```
Cadair:
```suggestion
        responded_to: Boolean initialized as False. True if event has been
```

opsdroid/events.py
```diff
@@ -0,0 +1,166 @@
...
+        raw_message: Raw message provided by chat service
+        regex: A re match object for the regular expression message was matched
+            against
+        responded_to: Boolean initialized as False. True if message has been
+            responded to
+
+    """
+
+    def __init__(self, user, room, connector,
+                 text, raw_message=None):  # noqa: D107
```
Cadair: I think it might be worth moving the `raw_message` kwarg up into `Event`. Certainly from the matrix perspective the raw event would be useful for much more than just messages. I would imagine most connectors will have a json or similar event structure that might be worth carrying around?

opsdroid/events.py
```diff
@@ -0,0 +1,166 @@
...
+        Logs response and response time in OpsDroid object stats.
+        """
+        opsdroid = get_opsdroid()
+        if isinstance(response_event, str):
+            response = copy(self)
+            response.text = response_event
+        else:
+            response = response_event
+
+        if 'thinking-delay' in self.connector.configuration or \
```
Cadair: style nit: why not `()` rather than line continuation?

opsdroid/events.py
```diff
@@ -0,0 +1,166 @@
...
+        if not self.responded_to:
+            now = datetime.now()
+            opsdroid.stats["total_responses"] = \
+                opsdroid.stats["total_responses"] + 1
+            opsdroid.stats["total_response_time"] = \
+                opsdroid.stats["total_response_time"] + \
+                (now - self.created).total_seconds()
+            self.responded_to = True
+
+    async def react(self, emoji):
```
Cadair: This should emit a `Reaction` event!!

opsdroid/events.py
```diff
@@ -0,0 +1,166 @@
...
+        return await self.connector.react(self, emoji)
+
+
+class File(Event):
+    """Event class to represent arbitrary files as bytes."""
+
+    def __init__(self, user, room, connector,
+                 file_bytes, url=None):  # noqa: D107
+        super().__init__(user, room, connector)
+        self.file_bytes = file_bytes
```
Cadair: I would suggest we make `file_bytes` a property which can either return `self._file_bytes` if it is set or call a method to download the bytes from the url.

opsdroid/events.py
Cadair: Arguably, this whole method should be deprecated in favour of some type of call to `Message.send` i.e.:

```python
Message.send(Reaction("😼"))
```

opsdroid/events.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        seconds = self.connector.configuration.get('typing-delay', 0)
+        char_count = len(text)
+
+        if isinstance(seconds, list):
+            seconds = randrange(seconds[0], seconds[1])
+
+        await asyncio.sleep(char_count*seconds)
+
+    async def respond(self, response_event, room=None):
```
Cadair: Should we rename this to `Event.send` as described in #719 in this PR?

opsdroid/events.py
```diff
@@ -0,0 +1,166 @@
...
+            response.text = response_event
+        else:
+            response = response_event
+
+        if 'thinking-delay' in self.connector.configuration or \
+           'typing-delay' in self.connector.configuration:
+            await self._thinking_delay()
+            await self._typing_delay(response.text)
+
+        await self.connector.respond(response, room)
```
Cadair: If we are refactoring as I propose in #719 this should be `opdroid.send`.

SolarDrew:
> I don't know how much of the wider changes in #719 we want to do here, specifically the whole generic opsdroid.send code path stuff. I would be happy to open a new PR for that once this one is done if people agree that's a better idea.

Sounds good. I figure this PR is likely to end up being just a fairly basic implementation, especially if it wants to be merged before the next release at the weekend.

> What's the motivation for rearranging the call signature of Message? That's a large API breakage no?

Yes, it is. But it made more sense to me to group `user`, `room` and `connector` that all events need, and then the subclass specific ones. And I'm sure there was a more practical reason when I decided this, but it escapes me for the moment. In any case, the old API is supported if you're using `opsdroid.message.Message`, so in theory nothing should break immediately.

opsdroid/events.py
SolarDrew: I agree on both points, but I think it's worth leaving for now and putting in the next PR.

opsdroid/events.py
SolarDrew: Makes sense. Again, I might leave this for the upgrade rather than including it now, unless you have suggestions for a quick and general way of implementing the download bit.

opsdroid/events.py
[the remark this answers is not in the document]
jacobtomlinson: Perhaps something like `raw_match`. I think we should make the object available but I'm also keen to abstract the useful bits into better properties. For example #446.

opsdroid/events.py
jacobtomlinson: Maybe it should be `raw_event`?

opsdroid/events.py
jacobtomlinson: Agree and agree that it can wait until the next PR.

opsdroid/events.py
jacobtomlinson: Happy to bump for now.

opsdroid/events.py
jacobtomlinson: Yes probably. We should make sure that `Message.respond` still exists for backward compatibility, but can just refer to `send`.

opsdroid/events.py
jacobtomlinson: I agree.

jacobtomlinson:
Looks like the docker failure in Travis in unrelated. Looking into it now in #815.

opsdroid/events.py
SolarDrew: Dunno, I didn't write that bit, but I'll change it.

opsdroid/events.py
SolarDrew: That's what I've gone with, yes.

opsdroid/events.py
SolarDrew: I agree we should do this but I'm actually not sure  it wants to be in this PR, unless you have a pressing desire for it to be in the release this weekend @jacobtomlinson . With the number of different parts of opsdroid that would be involved in the refactor I think it's worth dealing with in its own dedicated PR.

opsdroid/events.py
```diff
@@ -0,0 +1,169 @@
...
+            responded to
+
+    """
+
+    def __init__(self, user, room, connector,
+                 text, raw_event=None):  # noqa: D107
+        """Create object with minimum properties."""
+        super().__init__(user, room, connector)
+        self.text = text
+        self.raw_event = raw_event
```
jacobtomlinson: Could we also reinstate `raw_message` here to avoid this being a breaking change. It can just point to `self.raw_event`.

opsdroid/events.py
```diff
@@ -0,0 +1,169 @@
+"""Classes to describe different kinds of possible event."""
+
+from abc import ABC
```
jacobtomlinson: Can we sort these import nicely? I'm surprise the linter didn't pick it up.

E.g

```
import asyncio
from abc import ABC
from copy import copy
from datetime import datetime
from random import randrange

from opsdroid.helper import get_opsdroid
```

SolarDrew:
No problem, thanks for the feedback on it. Hope it doesn't all break straight away now :P

PR #729, opened 2018-11-07
Cancel pending tasks

### LINKED ISSUE #723: Application hangs on kill if connector doesn't disconnect
If a connector doesn't honour the disconnect method opsdroid can hang on shutdown. We should force kill connectors after 30 seconds after disconnect is called.

FabioRosado:
# Description

This is not the right way that the issue 723 suggests we should approach this issue. But I was unsure how to get running tasks because `asyncio.Task.all_tasks` seems to return all tasks not only pending but also the ones that have been set as done or finished.

I tried to search how to get only the pending tasks but couldn't figure out the best way to do this, since I couldn't figure out I couldn't create a condition to make opsdroid sleep for `x` amount of seconds before cancelling the task.

This solution does cancel the task but I'm not sure if this is the best way to do things like I said before. Would like to get some more feedback on how to approach the issue in a better way 👍 

Fixes #723

## Status
~~**READY**~~ | ~~**UNDER DEVELOPMENT**~~ | **ON HOLD**

## Type of change

# How Has This Been Tested?

- Tox passes
- Ran opsdroid, killed it and no exception was raised

# Checklist:

### DIFF
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -187,10 +187,9 @@ class OpsDroid():
 
         _LOGGER.info(_("Stopping pending tasks..."))
         tasks = asyncio.Task.all_tasks()
-        for task in tasks:
-            task.cancel()
-            with contextlib.suppress(asyncio.CancelledError):
-                await task
+        for task in list(tasks):
+            if not task.done() and task is not asyncio.Task.current_task():
+                task.cancel()
         _LOGGER.info(_("Stopped pending tasks"))
 
     async def reload(self):
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -185,6 +185,12 @@ class OpsDroid():
         self.cron_task = None
         _LOGGER.info(_("Stopped cron"))
 
+        _LOGGER.info(_("Stopping pending tasks..."))
+        tasks = asyncio.Task.all_tasks()
+        for task in tasks:
+            task.cancel()
+        _LOGGER.info(_("Stopped pending tasks"))
+
     async def reload(self):
         """Reload opsdroid."""
         await self.unload()
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -189,6 +189,8 @@ class OpsDroid():
         tasks = asyncio.Task.all_tasks()
         for task in tasks:
             task.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
+                await task
         _LOGGER.info(_("Stopped pending tasks"))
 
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
Yeah I also struggled with this. We need to cancel all tasks after a certain timeout (maybe 30 seconds), however this function is being called from within a task so we need to ignore that one. Sleeping for 30 seconds isn't great either as it would keep the thread open.

I think we should be counting the number of running tasks (not including the current task) every 0.1 seconds. If we get to 300 checks then they should be cancelled, if they all complete on their own then we should break out of the loop.

As you said counting the number of running tasks is tricky though. Might be worth a stack overflow question?

FabioRosado:
I was giving some thought about this issue and started looking into how other projects seem to close/cancel tasks, do you think is it possible to do something like in the  `aioslacker`? 

It seems they are adding all tasks(`ensure_future` creates tasks right?) into a set, with [lines 103-105](https://github.com/opsdroid/aioslacker/blob/26fb78de6e9fedfaab650ea5c3b87b33c4fc9657/aioslacker/__init__.py#L105)

```python
fut = ensure_future(coro, loop=self.loop)

self.futs.add(fut)
```

Then when we try to close the slacker the method `close` uses:

```python
        if self.futs:
            yield from asyncio.gather(*self.futs, loop=self.loop)

        yield from self.session.close()
```

Perhaps we could add the `connector.listen` method (this is where most of the issues happen) into a sort of set/list and on the `disconnect` just disconnect that task?

Another thing we could do is change the `connector.disconnect` method from being passed to `raise NotImplementedError` and then when we try to disconnect, if that exception is raised we can just cancel that task?

jacobtomlinson:
That was my original plan @petri, however the function is called within a task and it cancels itself. I haven't managed to find a way of not cancelling itself.

petri:
@jacobtomlinson recheck the code. It's not canceling itself due to checking for `p is not asyncio.current_task()`.

jacobtomlinson:
Totally missed that! Nice!

jacobtomlinson:
@FabioRosado could you please update this?

FabioRosado:
Just updated the PR, we will probably have an issue with testing as I need to test for `task.cancel()` but I haven't managed to figure out how to do this properly. I attempted to mock this method in order to see if it was called, but that didn't work out

FabioRosado:
So in the end testing this was way easier than I was thinking. Should I merge this now @jacobtomlinson ?

PR #734, opened 2018-11-13
Make it possible to create class-based skills

gergelypolonkai:
# Description

Make it possible to create class-based skills.  This can be very useful when the skill needs a helper object, like an SDK for an external API.

## Status
**UNDER DEVELOPMENT**

## Type of change

# How Has This Been Tested?

I created a very basic skill class for initial testing, but generic tests are still yet to come.

```python


class TestSkill(Skill):
    @match_regex(r'number [0-9]')
    async def describe_project(self, opsdroid, config, message):
        await message.respond('Yes, that’s a number.')
```

# Checklist:

### DIFF
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -215,14 +215,18 @@ class OpsDroid():
         """
         for skill in skills:
             for func in skill["module"].__dict__.values():
-                if isinstance(func, type) and issubclass(func, Skill) and func != Skill:
+                if (isinstance(func, type) and
+                        issubclass(func, Skill) and
+                        func != Skill):
                     skill_obj = func(self, skill['config'])
 
                     for name in skill_obj.__dir__():
-                        # pylint: disable=broad-except Getting an attribute of an object might
-                        # raise any type of exceptions, for example within an external library
-                        # called from an object property.  Since we are only interested in skill
-                        # methods, we can safely ignore these.
+                        # pylint: disable=broad-except
+                        # Getting an attribute of
+                        # an object might raise any type of exceptions, for
+                        # example within an external library called from an
+                        # object property.  Since we are only interested in
+                        # skill methods, we can safely ignore these.
                         try:
                             method = getattr(skill_obj, name)
                         except Exception:
--- a/docs/constraints.md
+++ b/docs/constraints.md
@@ -11,13 +11,15 @@ The constraint `constrain_rooms(rooms)` allows you to restrict a skill to a spec
 The following skill will respond with 'Hey' if a user says 'hi' in either the `#general` or `#random` rooms but not in `#public`.
 
 ```python
+from opsdroid.skill import Skill
 from opsdroid.matchers import match_regex
 from opsdroid.constraints import constrain_rooms
 
-@match_regex(r'hi')
-@constrain_rooms(['#general', '#random'])
-async def hello(opsdroid, config, message):
-    await message.respond('Hey')
+class MySkill(Skill):
+    @match_regex(r'hi')
+    @constrain_rooms(['#general', '#random'])
+    async def hello(self, message):
+        await message.respond('Hey')
 ```
 
 ## Constrain users
@@ -29,13 +31,15 @@ The constraint `constrain_users(users)` allows you to restrict a skill to a spec
 The following skill will respond with 'Hey' if the users `alice` or `bob` say 'hi' but not if `charlie` says it.
 
 ```python
+from opsdroid.skill import Skill
 from opsdroid.matchers import match_regex
 from opsdroid.constraints import constrain_users
 
-@match_regex(r'hi')
-@constrain_users(['alice', 'bob'])
-async def hello(opsdroid, config, message):
-    await message.respond('Hey')
+class MySkill(Skill):
+    @match_regex(r'hi')
+    @constrain_users(['alice', 'bob'])
+    async def hello(self, message):
+        await message.respond('Hey')
 ```
 
 ## Constrain connectors
@@ -47,11 +51,13 @@ The constraint `constrain_connectors(connectors)` allows you to restrict a skill
 The following skill will respond with 'Hey' via the `websocket` connector but not the `slack` connector.
 
 ```python
+from opsdroid.skill import Skill
 from opsdroid.matchers import match_regex
 from opsdroid.constraints import constrain_connectors
 
-@match_regex(r'hi')
-@constrain_connectors(['websocket'])
-async def hello(opsdroid, config, message):
-    await message.respond('Hey')
+class MySkill(Skill):
+    @match_regex(r'hi')
+    @constrain_connectors(['websocket'])
+    async def hello(self, message):
+        await message.respond('Hey')
 ```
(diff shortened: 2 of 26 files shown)


opsdroid/parsers/regex.py
```diff
@@ -32,7 +32,7 @@
                         "score": await calculate_score(
                             opts["expression"], opts["score_factor"]),
                         "skill": skill,
-                        "config": skill.config,
+                        "config": getattr(skill, 'config', getattr(skill, '__skill_attrs__', {}).get('config', None)),
```
jacobtomlinson: Could you explain a little what is going on here?

opsdroid/skill.py
```diff
@@ -0,0 +1,37 @@
...
+
+            for decor_params in getattr(method, '__matchers__', []):
+                decor_func, args, kwargs = decor_params
+
+                method.__skill_attrs__['config'] = self.config
+                decorator = decor_func(*args, **kwargs)
+                setattr(self, attr, decorator(method))
+
+    @classmethod
+    def set_matcher(cls, decor_func, *args, **kwargs):
```
jacobtomlinson: Could you explain why you have taken this approach?

I had imagined you would be able to still use the regular decorators on the methods. It may require updating the decorators but I would prefer that to forking the api.

opsdroid/skill.py
```diff
@@ -0,0 +1,37 @@
...
+        self.opsdroid = opsdroid
+        self.config = config
+
+        for attr in self.__dir__():
+            try:
+                method = getattr(self, attr)
+            except Exception:
+                continue
+
+            for decor_params in getattr(method, '__matchers__', []):
```
jacobtomlinson: Could you explain why this is necessary.

opsdroid/parsers/regex.py
gergelypolonkai: Unfortunately, you can’t modify a bound method’s attributes (in case of a class-based skill, `skill` here is a bound method of the instance).

So i try get the `config` attribute of `skill` with `getattr` (equals to `skill.config`) and if that fails, i try to get the `__skill_attrs__` of the function (which is set by `Skill.set_matcher` and updated by `Skill.__init__` with the actual config).

opsdroid/skill.py
gergelypolonkai: If you decorate a class method the decorated function is a class method (not bound to an instance yet). But when you activate a skill, you have to work on an object instance, not the unbound class method.

So either you store the skill instance(s) and for each unbound method call all the relevant instance methods, or do it this way :(

opsdroid/skill.py
gergelypolonkai: There can be extreme cases when `getattr` will throw an exception, like properties that need some initialisation.

opsdroid/parsers/regex.py
gergelypolonkai: Yes, but at this specific place you don’t know the instance, only the instance method that was decorated. And while you can get the relevant class method from an instance method, you can’t get neither the instance nor the class object from that.

opsdroid/skill.py
gergelypolonkai: I give it a go, but i’m afraid that requires bigger changes where we actually call the skill functions. We shall see.

opsdroid/parsers/regex.py
gergelypolonkai: And it immediately turned out i was wrong; `skill.__self__` holds the instance *if* the skill is from a class.  Let me rework this thing :)

opsdroid/skill.py
gergelypolonkai: Now that i think of it, that might require a huge overhaul of parsers, including external ones (ie. not written by opsdroid maintainers). But that might also apply to this solution, so i definitely give it a go now.

opsdroid/parsers/regex.py
jacobtomlinson: Ah you beat me to it! I just suggested this in the review below.

opsdroid/skill.py
jacobtomlinson: There are no external parsers. They are all core modules.

gergelypolonkai: Is it not possible to add external parsers, like it is with skills, connectors, and databases? If not, i can do the modifications, of course.

jacobtomlinson: Not currently.

gergelypolonkai:
I have reworked the whole thing (and learned a bit about Python internals while there). Now it works fine with the original decorators, without modifying the decorators themselves; it was much easier than i initially thought.

opsdroid/core.py
```diff
@@ -207,9 +208,22 @@ def setup_skills(self, skills):
         """
         for skill in skills:
             for _, func in skill["module"].__dict__.items():
+                if isinstance(func, type) and issubclass(func, Skill) and func != Skill:
+                    skill_obj = func(self, skill['config'])
+
+                    for name in skill_obj.__dir__():
+                        try:
+                            method = getattr(skill_obj, name)
+                        except Exception:
```
jacobtomlinson: Can you make this less broad?

gergelypolonkai: Actually, no. Imagine this code:

```python
class MySkill(Skill):
    @property
    def erroneous_property(self):
        return self.sdk_this_skill_uses.get_erroneous_property()
```

Where `get_erroneous_property()` can raise any type of exception. Been there, done that :disappointed:

We can either remove the try/except altogether and ask skill writers to look out for this corner case, or leave the except block this broad.

gergelypolonkai:
Should the deprecation warning be an actual [`DeprecationWarning`](https://docs.python.org/3.7/library/exceptions.html#DeprecationWarning) or just a log message like [the one for the Api.ai parser](https://github.com/opsdroid/opsdroid/blob/24eee6e7b67749113db46b34ca8c9ce8d92e5dde/opsdroid/core.py#L297)?

gergelypolonkai:
Also, should the deprecation message appear if a function based skill is used, even if there is no `setup` function?

opsdroid/core.py
jacobtomlinson: Fair enough! Can you please add a pylint disable with this justification in a commend. See [here](https://github.com/opsdroid/opsdroid/blob/7df9dc03b47bbb5b124ad6b7f74d02671058c60f/opsdroid/core.py#L272) for an example of this.

jacobtomlinson:
We should probably be raising deprecation warnings. No we should only show it if the skill uses it.

jacobtomlinson:
@gergelypolonkai I'm really keen to get this merged. Any chance you can pick it up again?

gergelypolonkai:
I’m a bit lost in the tests, but expect to finish soon.

jacobtomlinson:
Hi @gergelypolonkai. Are you able to pick this up again?

gergelypolonkai:
Most definitely! I really want to finish this, it’s just my $job that intervened.

gergelypolonkai:
Given i just got my second “stale” message, i hereby ask for some help. I tried to create a test for the class based skills, and for that i have created a `mockedmodule` with a skill in it. However, i can’t seem to figure out how to load it. Is there anyone who can give me some pointers?

jacobtomlinson:
Well it has taken a couple of months and some back and forth but it looks like this is ready to merge. Thanks for all the hard work @gergelypolonkai. I've managed to find some time today to sit down and sort out the tests and documentation, I understand how hard it can be to find time to pick these things back up again.

PR #747, opened 2018-11-21
Make shell connector optional

jacobtomlinson:
# Description
To avoid the current issues with the shell connector I've made it optional.

## Status
**READY**

## Type of change


jacobtomlinson:
@FabioRosado what do you think?

Issue #661, opened 2018-09-29
Specify modules by Python module name


jacobtomlinson:
Currently to configure modules for opsdroid (connectors, databases and skills) you specify a `path` or `repo` config option which tells opsdroid where it can find that specific module. Then at run time opsdroid will pull all of those modules together into an internal location, append that to the Python path and import all the modules.

It would be sensible to also include support for using modules from the regular Python path. These could be specified with `module` key and take the form of `module.submodule` (e.g the same thing you would `import`. These would bypass the custom loader intirely and directly import the package.

PR #750, opened 2018-11-22
no-install configuration option.

drmorr0:
# Description

We're operating in an environment where we install opsdroid in a virtualenv and don't want it to "install" our modules, it should just import them directly from the virtualenv.  This change allows this to work.

## Status

## Type of change

# How Has This Been Tested?

It's working locally in our environment.  I can update docs and write some tests if you want to merge this upstream.

# Checklist:

### DIFF
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -51,14 +51,17 @@ class Loader:
 
         # Proceed only if config.name is specified
         # and parent module can be imported
-        if config["name"] and importlib.util.find_spec(config["module_path"]):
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Thanks for raising this! We do have an open issue #661 which is also related to this problem. 

I would be interested to know if that solves your problem better and whether you would be interested in implementing it?

drmorr0:
Sure!  That seems a little more sensible than my approach, I'll see what I can do.

Issue #717, opened 2018-11-01
Implement disconnect method to connectors


FabioRosado:
# Description
Opsdroid can now exit in a clean way after #711 was merged. But most of the connectors (with the exception of websockets) don't implement the `disconnect` method and when opsdroid tries to exit, an exception occurs, because the connector task was destroyed while pending.

Implementing the disconnect method on every connector will avoid this exception to occur.

---
_As a work around until every connector is addressed should we just get all tasks and cancel every single one within the `core.unload` method?_

## Steps to Reproduce
Start opsdroid, wait for load and then kill it with `ctrl+c` the shell connector will throw the exception.

## Expected Functionality
Opsdroid should be able to exit without any exception being thrown.

## Experienced Functionality
Every single thing is cancelled but the connectors that don't implement the disconnect method.

```python
Caught exception

ERROR opsdroid.core: {'message': 'Task was destroyed but it is pending!', 'task': <Task pending coro=<ConnectorShell.listen() done, defined at /Users/fabiorosado/Library/Application Support/opsdroid/opsdroid-modules/connector/shell/__init__.py:64> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x10dbfd9a8>()]> cb=[gather.<locals>._done_callback(0)() at /usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/asyncio/tasks.py:609]>}
```

jacobtomlinson:
Related to #723

jacobtomlinson:
This still needs fixing.

jacobtomlinson:
Only the Rocket Chat connector needs this implementing now. @FabioRosado is this something you could pick up?

PR #754, opened 2018-11-23
fix slack error on connect

drmorr0:
# Description

When the slack client has an error on connecting, the `close` method on slacker is never called, leading to a very long string of tracebacks reporting that `open` is undefined and `_` cannot be found.  The actual issue is hidden several thousand lines at the top.

This PR fixes this.

## Status
**READY**

## Type of change

# How Has This Been Tested?

Tested locally: try starting opsdroid with an invalid token for Slack, for example.  Before this change, it spews out a thousand lines of tracebacks; after this change, it prints the error and shuts down cleanly.

# Checklist:

### DIFF
--- a/opsdroid/connector/slack/__init__.py
+++ b/opsdroid/connector/slack/__init__.py
@@ -63,7 +63,7 @@ class ConnectorSlack(Connector):
             _LOGGER.error("Failed to connect to Slack, retrying in 10")
             await self.reconnect(10)
         except Exception:
-            await self.slacker.close()
+            await self.disconnect()
             raise
 
     async def reconnect(self, delay=None):
@@ -75,6 +75,10 @@ class ConnectorSlack(Connector):
             await self.connect()
         finally:
             self.reconnecting = False
+
+    async def disconnect(self, opsdroid=None):
+        """Disconnect from Slack."""
+        await self.slacker.close()
 
     async def listen(self, opsdroid):
         """Listen for and parse new messages."""
--- a/opsdroid/connector/slack/__init__.py
+++ b/opsdroid/connector/slack/__init__.py
@@ -62,6 +62,9 @@ class ConnectorSlack(Connector):
             _LOGGER.error(error)
             _LOGGER.error("Failed to connect to Slack, retrying in 10")
             await self.reconnect(10)
+        except Exception:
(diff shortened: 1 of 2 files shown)


opsdroid/connector/slack/__init__.py
```diff
@@ -62,6 +62,9 @@ def __init__(self, config):
             _LOGGER.error(error)
             _LOGGER.error("Failed to connect to Slack, retrying in 10")
             await self.reconnect(10)
+        except Exception:
```
jacobtomlinson: Perhaps instead of catching a broad `Exception` and reraising it you could use a `finally` block.

tests/test_connector_slack.py
```diff
@@ -78,6 +78,17 @@ class TestConnectorSlackAsync(asynctest.TestCase):
...
+    async def test_abort_on_connection_error(self):
+        import aiohttp
+        connector = ConnectorSlack({"api-token": "abc123"})
+        connector.slacker.rtm.start = amock.CoroutineMock()
+        connector.slacker.rtm.start.side_effect = Exception()
+        connector.slacker.close = amock.CoroutineMock()
+
+        with self.assertRaises(Exception):
+            await connector.connect()
+        self.assertTrue(connector.slacker.close)
```
jacobtomlinson: Shouldn't this be `self.assertTrue(connector.slacker.close.called)`

opsdroid/connector/slack/__init__.py
drmorr0: I don't think `finally` will work here, because `finally` will get called even if no exception is raised, which we don't want.  I don't think the broad `Exception` here is necessarily an anti-pattern, since we do want do to some cleanup if something goes wrong, and we re-raise the Exception after doing that cleanup.

It does seem like #717 is semi-related here, and it seems like having a `disconnect` method here is the "right" approach for handling the cleanup, so I slightly modified the PR; but I don't have a better idea for how to actually handle arbitrary exceptions.  Happy to hear suggestions, though :)

tests/test_connector_slack.py
drmorr0: Whoops.  This is me being sloppy, sorry.

PR #755, opened 2018-11-24
load module from pythonpath

### LINKED ISSUE #661: Specify modules by Python module name

drmorr0:
# Description

This allows users to specify a skill module to import from your PYTHONPATH, by adding the "module" key to their configuration.yaml.

Fixes #661 

## Status
**READY**

## Type of change

# How Has This Been Tested?

Added new tests, and tested locally.

# Checklist:

### DIFF
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -11,7 +11,7 @@ import subprocess
 import sys
 import tempfile
 import urllib.request
-from collections import Mapping
+from collections.abc import Mapping
 
 import yaml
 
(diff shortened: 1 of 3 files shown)


opsdroid/loader.py
```diff
@@ -47,30 +47,35 @@ def import_module_from_spec(module_spec):
...
+            config["module"],
+            config["module_path"] + '.' + config["name"],
+            config["module_path"],
+        ]
+        for namespace in namespaces:
+            try:
+                module_spec = importlib.util.find_spec(namespace)
+                if module_spec:
+                    break
+            except (ModuleNotFoundError, AttributeError):
```
jacobtomlinson: The linter seems to be complaining that `ModuleNotFoundError` is not defined.

opsdroid/loader.py
```diff
@@ -47,30 +47,35 @@ def import_module_from_spec(module_spec):
...
             module = Loader.import_module_from_spec(module_spec)
             _LOGGER.debug(_("Loaded %s: %s"),
                           config["type"], config["module_path"])
             return module
-
-        _LOGGER.error(_("Failed to load %s: %s"),
-                      config["type"], config["module_path"])
-
-        return None
+        else:
```
jacobtomlinson: This `else` is unnecessary as it is after a return.

opsdroid/loader.py
drmorr0: Whoops, yea, missed this.  `ModuleNotFound` is a subclass of `ImportError` (https://airbrake.io/blog/python-exception-handling/class-hierarchy) so just changed it.

Issue #767, opened 2018-11-28
Support setuptools entry points for loading contributed extensions


petri:
It is not desirable to have every contributed extension in the opsdroid core.

By supporting a loader mechanism that uses setuptools entry points, it becomes easy for contributors to maintain extensions as proper Python packages and publish them in PyPI.

The existing existing GitHub loader mechanism can of course be kept as-is; the two mechanisms are complementary.

petri:
The entry points could be called simply `opsdroid_connectors`, `opsdroid_skills` and `opsdroid_databases`.

Someone contributing eg. a Skype connector in a package called eg. `opsdroid.contrib.skype` with a `connector.py` top-level module that has a `SkypeConnector` class, would then just need to add the following to setup.py:

```python
    entry_points = {
        'opsdroid_connectors': [
            'skype = opsdroid.contrib.skype.connector:SkypeConnector'
        ]
    },
```

FabioRosado:
Just to clarify but this will add another thing that we will have to add to configure out opsdroid right?
So we would have to change `config.yaml` plus add the `entry_points` dict into setup.py as well?

petri:
@FabioRosado no, the `entry_points` dict needs to be added just to the contributed proper Python packages.

FabioRosado:
Oh that's great then, I didn't know much about that. Thanks for clarifying this. 

It sounds like a great way to keep opsdroid "small" again 👍

jacobtomlinson:
This does sound good. I'm very keen to keep official supported modules in the core of opsdroid, they are not particularly big, many do not invoke additional dependencies and it makes testing and maintenance easier.

This could be a great way to support 3rd party connectors. However I'm not sure I see much difference between simply releasing a connector package to PyPI and configuring it with the new `module` config option added in #755.

petri:
> > It is not desirable to have every contributed extension in the opsdroid core.
> 
> This is a broad statement. Could you give your reasoning for this?
> 
More code means more maintenance and more dependencies that may have platform-specific installation issues and so on. Some extensions may be obscure. 

The "batteries included" slogan comes at a price.

> I'm very keen to keep official supported modules in the core of opsdroid, they are not particularly big, many do not invoke additional dependencies (especially after #766) and it makes testing and maintenance easier.
> 

[...]

PR #771, opened 2018-11-28
setuptools entry points support for loading contributed extensions (issue #767)

petri:
# Description

Implements #767 - setuptools entry points support for loading contributed extensions.

## Notes

 - entry point group name must be either `opsdroid_databases`, `opsdroid_skills` or `opsdroid_connectors`
- for an example definition, see [the example](https://github.com/koodaamo/opsdroid-zodb/blob/master/setup.py#L42-L46).

- The implementation currently prioritizes entry point based extensions.

# How Has This Been Tested?

A test was added to verify opsdroid loader now supports entry point mechanism. Tested also manually as working, using [opsdroid-zodb](https://pypi.org/project/opsdroid-zodb/).

To try it out, pull the implementation, do `pip install opsdroid-zodb`, create a skill that uses opsdroid memory, configure a 'zodb' database (yes just the name is enough) and try it out.

### DIFF
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -5,7 +5,6 @@ import importlib.util
 import json
 import logging
 import os
-import pkg_resources
 import re
 import shutil
 import subprocess
@@ -13,7 +12,7 @@ import sys
 import tempfile
 import urllib.request
 from collections.abc import Mapping
-
+from pkg_resources import iter_entry_points
 import yaml
 
 from opsdroid.helper import (
@@ -57,7 +56,8 @@ class Loader:
         # 4. try to import the module_path itself
 
         if config.get("entrypoint"):
-            _LOGGER.debug("Loading entry point-defined module for %s" % config["name"])
+            _LOGGER.debug(_("Loading entry point-defined module for %s"),
+                          config["name"])
             return config["entrypoint"].load()
 
         module_spec = None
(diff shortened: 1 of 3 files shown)


opsdroid/loader.py
```diff
@@ -304,6 +311,14 @@ def _load_modules(self, modules_type, modules):
             os.makedirs(DEFAULT_MODULE_DEPS_PATH)
         sys.path.append(DEFAULT_MODULE_DEPS_PATH)
 
+        # entry point group naming scheme: opsdroid_ + module type plural,
+        # eg. "opsdroid_databases"
+        epname = "opsdroid_%ss" % modules_type
```
jacobtomlinson: Just for code style consistency could you change this line to use `.format` rather than `%s`. We only use `%s` for logging.

opsdroid/loader.py
```diff
@@ -321,14 +336,18 @@ def _load_modules(self, modules_type, modules):
...
+                config["entrypoint"] = entry_points[config["name"]]
             config["module_path"] = self.build_module_import_path(config)
             config["install_path"] = self.build_module_install_path(config)
             if "branch" not in config:
                 config["branch"] = DEFAULT_MODULE_BRANCH
 
             # If the module isn't builtin, or isn't already on the
             # python path, install it
-            if not (config["is_builtin"] or config["module"]):
+            if not (config["is_builtin"]
```
jacobtomlinson: This statement seems to have a mix of `.get` and `[]` notations for the config dictionary. Could you please make it consistent (probably with `.get`).

petri: The mix is intentional to accommodate missing config `entrypoint` key, since it only exists when evaluating a packaged extension hooked in via entry point. But it just now occurred to me that perhaps we can count on `is_builtin` or `module` key being truthy in all such cases? Meaning that `config[entrypoint]` would never have a chance to fail.

petri: Duh, of course we cannot, given the whole point of this check... I guess we could just always set `config["entrypoint"]` then.

petri:
Lint fails because `_load_modules` has too complex logic it seems. Either lint has to relax or the said method needs to be refactored.

opsdroid/loader.py
jacobtomlinson: I would prefer to set it to be explicit and consistent.

jacobtomlinson:
I agree that the method should be refactored, but that is outside of the scope of this PR so don't worry. We will address it in #643 eventually.

I'm happy for you to update the pylint settings to relax the complexity rule enough for this to pass.

jacobtomlinson:
Also yes I don't think defaulting the config to `{}` makes sense any more.

jacobtomlinson:

PR #749, opened 2018-11-21
Refactor connector opsdroid pointer

zonagilreath:
# Description

Slight update to previous request

Fixes #725 

## Status
**READY** 

## Type of change

_Please delete options that are not relevant._

- Bug fix (non-breaking change which fixes an issue)
- Documentation (fix or adds documentation)

# How Has This Been Tested?

Tox tests passing

# Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #777, opened 2018-12-05
Refactor database opsdroid pointer

FabioRosado:
# Description

I'm not sure if this was something that we should have implement when #749 was merged into opsdroid, but @petri mentioned in gitter that _"There seems to be inconsistency in how opsdroid calls the extensions. Sometimes opsdroid is passed along and sometimes not when various methods of database or connector extensions can be called."_

This reminded me that on calling `disconnect()` on the databases it would break opsdroid after I merged #775 I thought that we were changing the pointer on the connectors and databases that's why I removed self from the disconnect, but I was wrong.

Perhaps we should apply the same work as we did on the connectors since opsdroid is only used on the `connect` method anyway.

If this is something that we don't really need to implement I will make another PR to restore the `database.disconnect(self)` so everything works okay.

Finally, I have updated the database documentation as I have encountered issues when using the `sqlite` database without any optional options. Like I said on gitter, I'm not sure if this should be changed or if I'm just missing something 😄 

## Status
**READY** |~~ **UNDER DEVELOPMENT**~~ | ~~**ON HOLD**~~

## Type of change

# How Has This Been Tested?

- Tox - all green

# Checklist:

### DIFF
--- a/docs/databases/mongo.md
+++ b/docs/databases/mongo.md
@@ -10,7 +10,7 @@ None.
 
 ```yaml
 databases:
-  mongo:
+  - name: mongo
     host:       "my host"     # (optional) default "localhost"
     port:       "12345"       # (optional) default "27017"
     database:   "mydatabase"  # (optional) default "opsdroid"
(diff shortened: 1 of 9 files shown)


jacobtomlinson:
This is good stuff. Happy for this to go ahead.

FabioRosado:
Thank you, as soon as requests get updated and the tests pass I'll merge this PR 👍

petri:
Lovely!

PR #778, opened 2018-12-05
Remove requests_mock

FabioRosado:
# Description

Travis seems to be broken because `requests_mock` couldn't be loaded. I checked our requirements and tests and it seems that this module is being loaded into `requirements_tests.txt` file but it seems we don't ever use this module on any of the tests since we don't use `requests`. 

This PR attempts to remove this module and see if anything breaks down on Travis since tox was happy with the changes.

## Status
**READY** 

## Type of change

# How Has This Been Tested?

- Tox - all green
- Test B

# Checklist:

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -7,5 +7,4 @@ pytest-cov==2.6.0
 pytest-timeout==1.3.3
 pydocstyle==3.0.0
 asynctest==0.12.2
-requests_mock==1.5.2
 mypy-lang==0.5.0


FabioRosado:
Now only the tests in docker are failing. I have looked around and the cause seems like this:

```python
pkg_resources.ContextualVersionConflict: (idna 2.8 (/usr/local/lib/python3.6/site-packages), Requirement.parse('idna<2.8,>=2.5'), {'requests'})
```

This took a while to figure out where are we using requests. After a bit of a search it seems that requests are only shown in `Pipfile.lock` - perhaps this file needs to be updated? 

I'm not too sure about what to do here at the moment so I'll wait for your comment @jacobtomlinson  👍

PR #784, opened 2018-12-18
Update docs with new path. Adds docker service example.

### LINKED ISSUE #781: Documentation is not clear about the docker volume configuration.
# Description
Using the documentation steps for launch opsdroid in Docker, can't load my custom configuration.

## Steps to Reproduce
1. Create a configuration file named `configuration.yaml`.
2. Run `docker run --rm -it -v $(pwd)/configuration.yaml:/etc/opsdroid/configuration.yaml:ro opsdroid/opsdroid:dev`

## Expected Functionality
The documentation should explain that `/path/to/configuration.yaml` must be the folder where the config file is, not the file path. Docker volumes maps folders, not files.

## Experienced Functionality
The docker volume option creates a folder called `configuration.yml` inside `/etc/opsdroid` so the default configuration is loaded.

## Versions
- **Opsdroid version:** opsdroid 0+unknown (inside the container)
- **Python version:** Python 3.6.7 (inside the container)
- **OS/Docker version:** Docker version 18.06.1-ce, build e68fc7a

## Configuration File
```yaml
welcome-message: true
web:
  host: '0.0.0.0'
  port: 8080
```

## Additional Details
Maybe adding a `docker-compose.yml` using docker configs can avoid problems. I created the file, so I can do a PR if you want.

AndresCidoncha:
# Description

Change docker run documentation for matching the new config path `/root/.config/opsdroid`.
Also adds a docker service deployment example.

Fixes #781 

## Status
**READY**

## Type of change

# How Has This Been Tested?

1. Run the changed/added commands in bash snippets.
2. Review that the configuration loaded is your custom config.

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,16 @@ docker pull opsdroid/opsdroid:latest
 docker run --rm -it -v /path/to/configuration.yaml:/root/.config/opsdroid/configuration.yaml:ro opsdroid/opsdroid:latest
 ```
 
+### Docker Service
+
+```bash
+# Create the opsdroid config file
+docker config create OpsdroidConfig /path/to/configuration.yaml
+
+# Create the service
+docker service create --name opsdroid --config source=OpsdroidConfig,target=/root/.config/opsdroid/configuration.yaml opsdroid/opsdroid:latest
+```
+
 ### Ubuntu 16.04 LTS
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Check out the [Getting Started](https://www.youtube.com/watch?v=7wyIi_cpodE&list
 docker pull opsdroid/opsdroid:latest
 
 # Run the container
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
This is great! 

Thinking more about your comments in #781 perhaps we should be encouraging people to be linking directories rather than files. Partly because opsdroid will make changes to the contents of `/root/.config/opsdroid` at runtime and we probably want to capture that in the volume.

I would be very happy if you want to change it to be at the directory level.

jacobtomlinson:
Thanks! I'm not up to date with docker services, is that still correct given this change?

AndresCidoncha:
Yes, I tested again after the changes. 

The service example is using docker external volume instead of a local path (docker handles the mount point of the external volume) and keeps the data between container executions.

Docker service will handle that the service always have X containers running (X defined by you, 1 by default). You can see it as supervisord for containers.

jacobtomlinson:

PR #803, opened 2019-01-09
Bump Pyyaml to avoid CVE-2017-18342

jacobtomlinson:
GitHub has alerted to CVE-2017-18342. This is related to #580.

The GitHub alert included a suggestion of updating to 4.2b1. Not sure why pyup-bot hasn't done this already.

### DIFF
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -249,7 +249,7 @@ class Loader:
         try:
             with open(config_path, 'r') as stream:
                 _LOGGER.info(_("Loaded config from %s."), config_path)
-                return yaml.load(stream)
+                return yaml.load(stream, Loader=yaml.SafeLoader)
         except yaml.YAMLError as error:
             _LOGGER.critical(error)
             sys.exit(1)
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ emoji==0.5.1
 nbconvert==5.4.0
 nbformat==4.4.0
 pycron==1.0.0
-pyyaml==3.13
+pyyaml==4.2b1
 websockets==7.0
 appdirs==1.4.3
 multidict==4.5.2
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -1,4 +1,4 @@
-
+import sys
 import os
 import shutil
 import subprocess
@@ -9,6 +9,7 @@ import unittest.mock as mock
 from types import ModuleType
 
 import pkg_resources
+from yaml import YAMLError
 from opsdroid.__main__ import configure_lang
 from opsdroid import loader as ld
 from opsdroid.loader import Loader
@@ -44,6 +45,24 @@ class TestLoader(unittest.TestCase):
         config = loader.load_config_file(
             [os.path.abspath("tests/configs/minimal_2.yaml")])
         self.assertIsNotNone(config)
+
+    def test_load_exploit(self):
+        """This will test if you can run python code from
+        config.yaml. The expected result should be:
+        - Yaml.YAMLError exception raised
+        - _LOGGER.critical message
+        - sys.exit
+
+        Note: the unittest.main(exit=False) is important so
+        other tests won't fail when sys.exit is called.
+        """
+        opsdroid, loader = self.setup()
+        with self.assertRaises(SystemExit):
+            config = loader.load_config_file(
+                [os.path.abspath("tests/configs/include_exploit.yaml")])
+            self.assertLogs('_LOGGER', 'critical')
+            self.assertRaises(YAMLError)
+            unittest.main(exit=False)
 
     def test_load_config_file_with_include(self):
         opsdroid, loader = self.setup()


jacobtomlinson:
@FabioRosado I know you were looking at this last. Any chance you can find some time to explore it again? We may have to remove the `!include` functionality to resolve this.

FabioRosado:
Hey sorry for not replying yesterday I was meant to and slip my mind.
Im off to work now and will finish around 4pm but I will have a look at it then.

I remember I changed something to safe_load and whatnot so that might be the issue. I don’t remember if I changed the envvar function but that bit is failing again, so I’ll have a look at the changes implemented on this new version of pyyaml and see what can be done 😀👍

FabioRosado:
I have spent some time trying to figure out how to solve this issue but I have been unsuccessful so far.  It seems that the pyyaml 4.2b1 is just an attempt to fix the security issue that was found - basically they replace `yaml.load` with `yaml.safe_load`, so the safe_load is the default. This change in itself shouldn't cause so many issues with the `!include` and `$envvar` constructors. But somehow it does.

Both tests are failing because pyyaml isn't recognising the !include constructor - probably because it starts with `!` and perhaps the safe_load will just ignore or throw an error if it finds this in the original yaml file.

The envvar constructor is working fine, but pyyaml is not able to convert `$ENVVAR` into something like it used to do in 3.13. Since homeassistant is currently using `!include` I had a peek to see if they fixed this issue somehow, but they are still using the version 3.13 so I am currently unsure how to fix this issue 👎

jacobtomlinson:
I guess one thing to bear in mind here is that this vulnerability allows you to execute arbitrary code included in a yaml file. This would be a concern if you are using yaml files for data storage or some other similar purpose, but in this case we are using it for configuration. Therefore there can be some protection by setting permissions on the config.

For example if you are running opsdroid on linux and you have a dedicated `opsdroid` user then anyone that can edit the config will be able to have their nasty code executed as the `opsdroid` user. However I think it is fair to assume that any user who can edit the config is likely a privileged user anyway.

FabioRosado:
**Sorry for this massive post!**

I agree with that. Also, it seems that there has been some confusion about this release and GitHub notifying users, I have been following the chat on it and yeah some people wonder why this hasn't been implemented and what not.

I think the reason why pyup-bot didn't notify us of this version is because this was just a tag that was added, it's not a full release - the confusing thing is that they have two branches `release/4.1` and `release/4.2`  but I am not sure if these changes have been merged into master.

I read a comment of one of the maintainers and it seems the team is struggling with a few issues with the library since this is an old library. I know that the changes added in 4.2b1 were merged but then reverted - perhaps because things didn't work out as expected like what is happening with us with $ENVVARS and !include.

When I heard about the security vulnerability of pyyaml I was more worried about my bit of code - the `!include` then the rest of the code because I thought if someone really wanted to do some harm then the include command would be the fastest and easiest way to do so. 

Currently, I'm happy with my attempt to fix the issue since I used `safe_load` for included yaml files and users can't run python code on an included file. 

Obviously, there is still the risk of someone being able to escalate their privileges to be able to change the `config.yaml` and run python code from them - but if they are able to do that they could as well just write an opsdroid skill and use python directly and just inject it in the `config.yaml` so the risk is still there even if we fix the pyyaml one.

One way that we could perhaps fix the pyyaml side of things would be to check [home assistant yaml](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/util/yaml.py) file. I was looking at it yesterday and they build a lot of things in order to use yaml and it seems that they use `SafeLoader` everywhere which fixes the security vulnerability.

At the moment, I am unable to implement something like that because I will have to spend more time in the yaml documentation, but might be a good thing to look at in the future?

FabioRosado:
I have been trying to play around with it and I don't think home assistant will be able to help us much as it is. Currently they are using `pyyaml>=3.13,<4` and what I have gathered so far is that 4.2b1 will break any work done under 4.2, the reason for this is how safe_load is coded, it won't accept much and you can pass arguments to it, so you can do something like: `yaml.load(file, yaml.SafeLoader)`  and it seems to be breaking a lot of things - my `!include` constructor is broken because pyyaml can't find that constructor.

I have found  this issue in stackoverflow [How to deserialise and object with pyyaml](https://stackoverflow.com/questions/2627555/how-to-deserialize-an-object-with-pyyaml-using-safe-load) and perhaps the trick might be to use `yaml.YAMLObject` in order to get safe_load to accept our constructors and load things as before.

The documentation of pyyaml has been nothing but frustrating and since 4.2b1 isn't yet an official release there is no documentation related to it (so I am going to spend time reading the code and try to understand what's going on).  But the documentation mentions this:

*A python object can be marked as safe and thus be recognized by yaml.safe_load. To do this, derive it from yaml.YAMLObject (as explained in section Constructors, representers, resolvers) and explicitly set its class property yaml_loader to yaml.SafeLoader.*

This together with the stackoverflow question gives me a feeling that perhaps marking something with `yaml.YAMLObject` might work but I have wrong many times before haha

FabioRosado:
I am just pushing this changes because locally this seems to be working. I re-tested it over and over again just in case my tox and pycharm seem happy o.O

FabioRosado:
I have added a test to this commit as well. It doesn't improve coverage but seemed like a good idea to have in case something changes in pyyaml and things stop working as we expect

jacobtomlinson:
Awesome work!

FabioRosado:
I think I broke the GitHub build? Travis is showing green here but on GitHub it still showing pending o.o

jacobtomlinson:
Weird! Everything is green so I'm going to merge.

PR #812, opened 2019-01-16
Add tweet button with text below the badges

Yash-Garg:


jacobtomlinson:
Thanks for raising the PR! Could you add some justification about why you think this is an improvement?

Yash-Garg:
@jacobtomlinson Because if feltt hat the tweet button was giving an ugly look being in the right of OpsDroid logo

jacobtomlinson:
What makes you say it is ugly? I'm just curious. I feel like this is subjective.

Yash-Garg:
@jacobtomlinson I mean most of the people won't be able to see it, so it is better to be down with a line 👍

jacobtomlinson:
Ok I think you've convinced me. I'm happy for you to remove it from next to the logo. However could we change the text a little, perhaps to something a bit shorter and snappier. Perhaps like

> Help support opsdroid in one click by pressing <tweet button>

Yash-Garg:
@jacobtomlinson Updated the message 👍

jacobtomlinson:
Many thanks!

PR #826, opened 2019-01-30
Event dispatch system

Cadair:
# Description

This is a follow on for #721 which addresses the second stage of #719. This provides the foundation for connectors to be able to respond to different types of events, via a `send` method.

~*No allowances for backwards compatibility have been made yet.*~

## Status
**READY**

## Type of change

# Checklist:

### DIFF
--- a/docs/extending/connectors.md
+++ b/docs/extending/connectors.md
@@ -16,14 +16,27 @@ Connectors are a class which extends the base opsdroid Connector. The class has 
 #### user_typing
 *user_typing* triggers the event message *user is typing* if the connector allows it. This method uses a parameter `trigger` that takes in a boolean value to trigger the event on/off.
 
-#### respond
-*respond* will take a Message object and return the contents to the chat service.
-
-#### react
-*react* will take a Message object and an emoji name and add this emoji as a Message reaction (if the chat service supports this feature).
-
 #### disconnect
 *disconnect* there is also an optional disconnect method that will be called upon shutdown of opsdroid. This can be used to perform any disconnect operations for the connector.
+
+
+### Handling Events
+
+Opsdroid supports different types of events, which can both be sent and received via connectors, for more information on the different types of events see the [events documentation](./events.md).
+
+
+Connectors can implement support for sending different types of events using the `opsdroid.connector.register_event` decorator.
+This decorator is used to define a method (coroutine) on the connector class for each different event type the connector supports, for instance to add support for `Message` events you may define a method like this:
+
+```python
+
+@register_event(Message)
+async def send_message(self, message):
+    await myservice.send(message.text)
+    
+```
+
+This method will be called when `Connector.send` is called with a `Message` object as its first argument. Methods such as this can be defined for all event types supported by the connector.
 
 ```python
 
@@ -34,7 +47,7 @@ import time
 import chatlibrary
 
 # Import opsdroid dependencies
-from opsdroid.connector import Connector
+from opsdroid.connector import Connector, register_event
 from opsdroid.events import Message
 
 
@@ -66,10 +79,11 @@ class MyConnector(Connector):
       # Parse the message with opsdroid
       await opsdroid.parse(message)
 
-  async def respond(self, message):
+  @register_event(Message)
+  async def send_message(self, message):
     # Send message.text back to the chat service
-    await self.connection.send(raw_message.text, raw_message.user,
-                               raw_message.room)
+    await self.connection.send(message.text, message.user,
+                               message.target)
 
   async def disconnect(self, opsdroid):
     # Disconnect from the chat service
--- a/docs/extending/connectors.md
+++ b/docs/extending/connectors.md
@@ -22,7 +22,7 @@ Connectors are a class which extends the base opsdroid Connector. The class has 
 
 ### Handling Events
 
-Opsdroid supports different types of events, which can both be sent and received via connectors, for more information on the different types of events see the [events documentation](./events.md).
+Opsdroid supports different types of events, which can both be sent and received via connectors, for more information on the different types of events see the [events documentation](/events.md).
 
 
 Connectors can implement support for sending different types of events using the `opsdroid.connector.register_event` decorator.
--- a/docs/extending/connectors.md
+++ b/docs/extending/connectors.md
@@ -44,7 +44,7 @@ class MyConnector(Connector):
     # Init the config for the connector
     self.name = "MyConnector" # The name of your connector
     self.config = config # The config dictionary to be accessed later
-    self.default_room = "MyDefaultRoom" # The default room for messages to go
+    self.default_target = "MyDefaultRoom" # The default room for messages to go
 
   async def connect(self, opsdroid):
     # Create connection object with chat library
--- a/docs/extending/events.md
+++ b/docs/extending/events.md
@@ -1,0 +1,80 @@
+# Events
+
+Events in opsdroid represent things that happen on the services that opsdroid connects to.
+They are both sent and received by the connectors.
+
+All events are a subclass of the base `Event` class. Each event also defines its own set of attributes which are specific to that type of event.
+
+
+### `Event`
+
+The base `Event` class has the following attributes and methods:
+
+
+#### Attributes
+* `user`: A string containing the username of the user who wrote the message.
+
+* `target`: A string normally containing the name of the room or chat channel the message was sent in.
+
+* `connector`: A pointer to the opsdroid _connector object_ which receieved the message.
+
+* `linked_event`: Another event linked to this one, for example the event that a `Message` replies to.
+
+* `created`: A timestamp of when this event was instantiated.
+
+* `event_id`: A unique identifier for this event as provided by the connector.
+
+* `raw_event`: The raw event received by the connector (may be `None`).
+
+* `responded_to`: A boolean (True/False) flag indicating if this event has already had its `respond` method called.
+
+
+#### Methods
+* `respond(event)`: A method which takes another `Event` object responds using any attributes of the event which are not overridden in the response `event`. i.e. if the `target` attribute on the `event` argument is `None` then the `target` of the event being responded to will be used.
+
+### `Message`
+
+The `Message` object is the most common event type, it represents textual events, and has the following attributes in addition to the base `Event` attributes.
+
+
+#### Attributes
+
+* `text`: The textual content of the message.
+
+#### Methods
+* `respond(text)`: The same as the `Event.respond` method, with the addition of accepting a `str` argument, which will be converted into a `Message` event.
+
+### `Typing`
+
+An event to represent a typing indication.
+
+#### Attributes
+
+* `trigger`: Trigger typing on or off.
+* `timeout`: Timeout on typing event.
+
+### `Reaction`
+
+A reaction to a message, normally an emoji or a string representing one.
+
+#### Attributes
+
+* `emoji`: The emoji to use as a reaction.
+
+### `File`
+
+A file event. All files should be representable as bytes. A `File` can be constructed using either a url or a `bytes` object, if a url is specified the `File.file_bytes` property will retrieve the url and store the response as `bytes`.
+
+#### Attributes
+
+* `url`: The url of the file, can be `None`.
+* `file_bytes`: The file as a `bytes` object.
+
+### `Image`
+
+The `Image` event is a subclass of `File`, so in the same way `File` can an `Image` can be either a url or bytes, but should always be representable as `bytes`.
+
+#### Attributes
+
+* `url`: The url of the file, can be `None`.
+* `image_bytes`: The image as a `bytes` object.
--- a/docs/extending/events.md
+++ b/docs/extending/events.md
@@ -16,7 +16,7 @@ The base `Event` class has the following attributes and methods:
 
 * `target`: A string normally containing the name of the room or chat channel the message was sent in.
 
-* `connector`: A pointer to the opsdroid _connector object_ which receieved the message.
+* `connector`: A pointer to the opsdroid _connector object_ which received the message.
 
 * `linked_event`: Another event linked to this one, for example the event that a `Message` replies to.
 
--- a/docs/extending/skills.md
+++ b/docs/extending/skills.md
@@ -39,14 +39,14 @@ In opsdroid when events are received the connector emits `Event` objects which c
 _Note: Not all connectors support all event types. To find out which events a connector will emit you can access the `.events` attribute of a connector._
 
 
-The base `Event` class has the following attributes. Also depending on the matcher it may have parser specific properties too. See the [matchers documentation](../tutorials/introduction.md#matchers-available) for more details.
+The base `Event` class has the following attributes. Also depending on the matcher it may have parser specific properties too. See the [matchers documentation](tutorials/introduction.md#matchers-available) for more details.
 
 
 * `user`: A _string_ containing the username of the user who wrote the message.
 
 * `target`: A _string_ normally containing the name of the room or chat channel the message was sent in.
 
-* `connector`: A pointer to the opsdroid _connector object_ which receieved the message.
+* `connector`: A pointer to the opsdroid _connector object_ which received the message.
 
 * `raw_event`: The raw event received by the connector (may be `None`).
 
--- a/docs/extending/skills.md
+++ b/docs/extending/skills.md
@@ -28,32 +28,41 @@ If the message matches the regular expression then the decorated function is cal
 
 To ensure the bot is responsive the concurrency controls introduced in Python 3.5 are used. This means that all functions which will be executed should be defined as an `async` function, and calls to functions which may require IO (like a connector or database) should be awaited with the `await` keyword. For more information see [asyncio](https://docs.python.org/3/library/asyncio.html) and [event loops](https://docs.python.org/3/library/asyncio-eventloop.html).
 
-## Message object
+## Events
 
-The message object passed to the skill function is an instance of the opsdroid `Message` class which has the following properties and methods.
+In opsdroid when events are received the connector emits `Event` objects which can be matched by skills and processed. The most common event is the `Message` event but a number of other events are implemented including:
 
-Also depending on the matcher it may have parser specific properties too. See the [matchers documentation](../tutorials/introduction.md#matchers-available) for more details.
+* Reaction
+* File
+* Image
 
-### `text`
+_Note: Not all connectors support all event types. To find out which events a connector will emit you can access the `.events` attribute of a connector._
 
-A _string_ containing the message from the user.
 
-### `user`
+The base `Event` class has the following attributes. Also depending on the matcher it may have parser specific properties too. See the [matchers documentation](../tutorials/introduction.md#matchers-available) for more details.
 
-A _string_ containing the username of the user who wrote the message.
 
-### `room`
+* `user`: A _string_ containing the username of the user who wrote the message.
 
-A _string_ containing the name of the room or chat channel the message was sent in.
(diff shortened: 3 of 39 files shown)


Cadair:
Ok, so this is spiralling into both the API for connectors (which I am quite happy about) and now also the API for responding to Events, in the land of many event types.

The latter came about as I was trying to work out what the best API for replacing `Message.react` was. What I have settled on is the following:

```
message.respond(event=Reaction("🌮"), target=message)
```
i.e to react to or reply to (or link in any other conceptual way) you set the target for the event as the previous event you want to link it to.

This necessitated changing the constructor of `Event` quite a lot to allow it to take only one argument (and then inherit all the metadata from the event you are responding to).

Finally I then thought why not shortcut this a little and implement a generic `reply` method on `Event` for when you want to link the event you are responding with to the event you are responding to which is implemented in the this commit https://github.com/opsdroid/opsdroid/pull/826/commits/5162309f49bd1039f1c364ab5fca4a26e7f5616c but is just:

```python
async def reply(self, event):
    self.respond(event=event, target=self)
```

opsdroid/connector/__init__.py
```diff
@@ -56,60 +103,38 @@ def configuration(self):
...
+            event (Event): A message received by the connector.
+            target (string): The name of the room or other place to send the
+            event.
 
-        Triggers the "user is typing" event if the chat service that
-        opsdroid is connected to accepts it.
+        Returns:
+            bool: True for event successfully sent. False otherwise.
         """
+        target = target if target else event.target if event.target else self.default_target
```
SolarDrew: You're a monster.

jacobtomlinson: This is some of the least readable code that I've ever seen 😆. Please don't do things like this, it makes debugging a chore.

Cadair: I agree that it's insane. However, the other options I can think of are pretty much equally insane:

```python
        target_arg = target
        target = self.default_target
        target = event.target if event.target else target
        target = target_arg if target_arg else target
```

```python
        target_arg = target
        target = self.default_target
        if event.target:
            target = event.target
        if target_arg:
            target = target_arg
```

I am open to other suggestions.

Cadair: of course there is always:

```python
list(filter(bool, (target, event.target, self.default_target)))[0]
```

:rofl:

jacobtomlinson: How about

```python
target_arg = target
target = target_arg or event.target or self.default_target
```

Cadair: :scream_cat: I did not know you could do that. That's clearly better.

jacobtomlinson: When using an `or` in assignment it evaluates the truthiness of each thing being assigned and chooses the first truthy one. So the `None` type variables will just pass to the right.

You can fall into a trap with is because sometimes you want to assign a falsy value. If one of the variables contained an integer `0`, an empty list `[]` or dict `{}` or a boolean `False` it will not assign. But as we know the target will either be an `Event` based object or `None` we are safe.

jacobtomlinson: Your mega one line if was so impenetrable I didn't realize this is what you were trying to do, otherwise I would've suggested it sooner 😆.

Cadair:
> This is a good start. I like the API for registering events in the connector.

Awesome, this is the main review I was after. Are you also onboard with the API of `Event.respond(EventType(content))`?

> We should think a little about the difference between `send`, `respond` and `reply`. I want this to be really clear to users.

I agree, I am not completely sold that what I have is the best solution, but let me explain what I think the purpose of the three methods are:

* `opsdroid.send` Should be for skills that want to emit events that are not a response to another event, i.e. on a cron etc. It's a replacement for the pattern of doing `Message().respond()` at the moment. `Connector.send` is not really supposed to be "public" API in the sense that it's called by other things that are more public i.e `Event.respond` and `opsdroid.send`. [1]

* `Event.respond` is a way of shortcutting `opsdroid.send` with the parameters of the previous event, [as done here](https://github.com/opsdroid/opsdroid/pull/826/files#diff-1119f556aa74c33d12583d03ced0b89cR56), so basically it's send to the same connector / target as the message you are responding to.

* `Event.reply` is a linked respond, so it's send with the same connector, but with the target as the previous message. This is the one that seems the least clear to me.



[1] Also `opsdroid.send` should support specifying connector by string to enable skills to do things with specific connectors without having to get the actual connector instance themselves.

opsdroid/events.py
```diff
@@ -69,13 +108,11 @@ class Message(Event):
 
     """
 
-    def __init__(self, user, room, connector,
-                 text, raw_event=None):  # noqa: D107
+    def __init__(self, text, **kwargs):
         """Create object with minimum properties."""
-        super().__init__(user, room, connector)
+        super().__init__(**kwargs)
```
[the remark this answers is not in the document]
Cadair: yes probably.

Cadair:
> However we then open ourselves up to having random methods for the different response types.

Yes, I really want to avoid this.


`reply` is probably a bad noun for the `reply` method as implemented here, but consider the following examples to try and get across the functionality that I think is useful:

(Consider `amessage` an instance of `Message` received by a skill)
```python
amessage.respond(Reply("This is a matrix-style reply to a message,"
                       " could also be a slack thread"), target=amessage)
```
and
```python
amessage.respond(Reaction("👍"), target=amessage)
```

here, a `Reply` event and a `Reaction` event, both have their target set to the message they are linked too as opposed to the room. This would be interpreted by the connector as linking the new event `Reply` or `Reaction` to the previous message itself, i.e to create a thread or react to a message.

The objective of the `reply` method is just to avoid having to do an explicit `target=amessage` so I am very open to a better noun.

Cadair:
> Why is it bad?

The main objective of this whole exercise for me is to make the core opsdroid API be event type agnostic, so having a bunch of event-specific methods is the opposite of that to me.

> It means that the api of some events (like `Message`) could creep with the addition of new event subclasses.

I don't really follow on this. Addition of a subclass shouldn't affect the API of parent classes?

I do agree that it's important that we keep the bar low for skills that only use the basic few event types, but I hope that there will soon be many many different event types supported for different types of connector (i.e. github), so we also want to come up with an API which has a nice ram up to emitting and processing other events.

> Skills look nice and neat with simple methods like `respond` and `react`.
> [...]
> I'm very worried that these changes hugely increase the complexity for people who are not confident with Python.

I agree we should strive to keep skills that are only dealing with `Message` nice and simple. We should keep the `Message.respond("a string")` API to default to sending back a `Message` event.

Specifically with `react` I am loathed to special case it, because *"Special cases aren't special enough to break the rules."*, and as I elaborate on in a moment there are many other event types which are conceptually similar to reactions.

> Why not make the default target option `self` within the `respond` method?

I think that the default target for `Message.respond("string")` should stay as it is at the moment and be the "room" not the previous message, because conceptually you are sending the message into the "room" not directly related to the event (like a reply or a reaction).

In a busy room by the time your response goes in it could be 10-20 messages later, there might not be a direct link to the previous message any more. (There also might be, so that's the difference as I see it).

> The other option would be for the target to be part of the `Reply` and `Reaction` objects.

Given we already have `target` for all events (currently implemented in opsdroid as `room`) i.e. to send to a given room or a given PR / Issue on the GH connector etc. It seems redundant to have it defined in the Event object as well as the `respond` / `send` method?

The main thrust of me implementing a `reply` method to replace `react` is that there is a category of event which conceptually links to the previous event not just reactions. If you start thinking about using `target` in this way it could have a lot of uses for GitHub like connectors (a commit status would have a target of a commit etc), so I would like to work through an API which supports as many use cases as we can think of.

Cadair:
Ok, so after talking this through with @SolarDrew I realised that my dual use of `target` was probably a bad idea. So I have refactored the code to follow a different API, which hopefully gets rid of any magic and allows a clean API for users.

The first point is that I have removed the `target=` (previously `room=`) kwarg from all the `send`/`respond` methods, this is because it's already specified as part of the event.[1]

This lead to the following API:
```python
amessage.respond(Reply("A reply"))
```
which is equivalent (if not responding to an event) to:
```python
opsdroid.send(Reply("A reply", connector="matrix",
                    linked_event=amessage, target="#general", user="some_user_id))
```
*Note: `linked_event` could also be a string ID rather than a previous `Event` subclass here.*

The idea here is that an event that's been through `respond` should always know which event it has been responded to, this is now done through a separate kwarg / property called `linked_event`. `target` is still the room.

What happens underneath is that inside `Event.respond` the [target is inherited](https://github.com/opsdroid/opsdroid/pull/826/files#diff-1119f556aa74c33d12583d03ced0b89cR62) from the event you are responding to and the `linked_event` is [set to the event to which you are responding](https://github.com/opsdroid/opsdroid/pull/826/files#diff-1119f556aa74c33d12583d03ced0b89cR64).

This means that how links between events are handled in the connector is based solely on the type of the event, `Reply` vs `Message` rather than the `target` argument, which seems easier to understand for people unfamiliar with the situation to me.

*[1] To be honest, I don't really understand why `Connector.respond` has a `room=` kwarg at the moment. It's probably only useful as a shortcut to prevent having to pass `Message()` rather than a `str` into `Connector.respond` which post this PR is a pattern we shouldn't be encouraging anyway.

jacobtomlinson:
I much prefer this. It is conceptually simpler but still has the full power and flexibility that you are trying to add.

I agree with your comments about special cases. It does mean that `message.react("😄")` will now be `message.respond(Reaction("😄"))`. I'm still a little conflicted about this, but I think I just need to accept that you are right.

Cadair:
@SolarDrew @jacobtomlinson I think this is complete and ready for review :smile:

docs/extending/connectors.md
```diff
@@ -16,15 +16,28 @@ Connectors are a class which extends the base opsdroid Connector. The class has
...
-#### react
-*react* will take a Message object and an emoji name and add this emoji as a Message reaction (if the chat service supports this feature).
-
 #### disconnect
 *disconnect* there is also an optional disconnect method that will be called upon shutdown of opsdroid. This can be used to perform any disconnect operations for the connector.
 
+
+### Handling Events
+
+Opsdroid supports different types of events, which can both be sent and received via connectors, for more information on the different types of events see the [events documentation](./events.md).
```
FabioRosado: I'm not sure if the link will work in readthedocs - because linking is a bit weird. I believe if you use `/events.md` it will work fine

docs/extending/events.md
```diff
@@ -0,0 +1,80 @@
...
+
+The base `Event` class has the following attributes and methods:
+
+
+#### Attributes
+* `user`: A string containing the username of the user who wrote the message.
+
+* `target`: A string normally containing the name of the room or chat channel the message was sent in.
+
+* `connector`: A pointer to the opsdroid _connector object_ which receieved the message.
```
FabioRosado: Typo fix - `which received the message`

docs/extending/skills.md
```diff
@@ -28,32 +28,41 @@ If the message matches the regular expression then the decorated function is cal
...
+* File
+* Image
 
-### `text`
+_Note: Not all connectors support all event types. To find out which events a connector will emit you can access the `.events` attribute of a connector._
 
-A _string_ containing the message from the user.
 
-### `user`
+The base `Event` class has the following attributes. Also depending on the matcher it may have parser specific properties too. See the [matchers documentation](../tutorials/introduction.md#matchers-available) for more details.
```
FabioRosado: same thing with linking here, I believe we need to remove the `..`

docs/extending/skills.md
```diff
@@ -28,32 +28,41 @@ If the message matches the regular expression then the decorated function is cal
...
-A _string_ containing the username of the user who wrote the message.
 
-### `room`
+* `user`: A _string_ containing the username of the user who wrote the message.
 
-A _string_ containing the name of the room or chat channel the message was sent in.
+* `target`: A _string_ normally containing the name of the room or chat channel the message was sent in.
 
-### `connector`
+* `connector`: A pointer to the opsdroid _connector object_ which receieved the message.
```
FabioRosado: typo fix - `which received the message`

opsdroid/connector/__init__.py
```diff
@@ -75,7 +130,25 @@ def configuration(self):
             bool: True for message successfully sent. False otherwise.
 
         """
-        raise NotImplementedError
+        warnings.warn(
```
FabioRosado: Should this be `_LOGGER.warning` instead? I'm mostly just asking here haha

opsdroid/connector/__init__.py
```diff
@@ -92,24 +165,49 @@ def configuration(self):
...
+            bool: True for event successfully sent. False otherwise.
 
-        Triggers the "user is typing" event if the chat service that
-        opsdroid is connected to accepts it.
         """
+        if not isinstance(event, Event):
+            raise TypeError(
+                "The event argument to send must be an opsdroid Event object")
 
-    async def disconnect(self):
```
FabioRosado: Should we get rid of the disconnect method? (even though we don't have many connections implementing them)

opsdroid/events.py
```diff
@@ -2,45 +2,84 @@
...
+        opsdroid = get_opsdroid()
+
+        # Inherit the user, target and event from the event we are responding
+        # to if they are not explicitly provided by this Event
+        event.user = event.user or self.user
+        event.target = event.target or self.target
+        event.connector = event.connector or self.connector
+        event.linked_event = event.linked_event or self
+
+        opsdroid = get_opsdroid()
```
FabioRosado: is this a duplicate `opsdroid = get_opsdroid()`?

opsdroid/message.py
```diff
@@ -11,10 +11,52 @@
...
+    def __init__(self, text, user, room, connector,
+                 raw_message=None):  # noqa: D401
+        """Deprecated opsdroid.message.Message object."""
+        super().__init__(text, user, room, connector,
+                         raw_event=raw_message)
+
+    @property
+    def room(self):  # noqa: D401
+        """The room the message was sent into."""
+        warn(
```
FabioRosado: Aboved I asked about warnings vs _Logger, if we choose warnings, perhaps we should choose to either use `warninings.warn` or just `warn` throughout the codebase, what do you think?

opsdroid/connector/__init__.py
Cadair: Hmmm, maybe, I don't know. Also is `DeprecationWarning` the best thing to use?

opsdroid/connector/__init__.py
Cadair: I haven't deleted it, just moved it up.

FabioRosado: I missed that sorry haha

opsdroid/connector/__init__.py
FabioRosado: To be fair the warnings.warn act similar to logging the warning, might be good to just keep to one way to log/warn users, perhaps Jacob has an opinion about this. I dont see a problem with `DeprecationWarning` 😄

Cadair: Well `DepreactionWarning` isn't printed by default, so people will only see it if it's logged by opsdroid somehow.

FabioRosado: @SolarDrew  added `warning.war` to message.py and the message is being logged, it seems that you can ignore these warnings so perhaps that is the reason why it was added?

opsdroid/events.py
```diff
@@ -102,69 +139,98 @@ def __init__(self, user, room, connector,
...
+        super().__init__(*args, **kwargs)
+
+        self._file_bytes = file_bytes
         self.url = url
 
+    @property
+    def file_bytes(self):  # noqa: D401
+        """The `bytes` representation of this file."""
+        if not self._file_bytes and self.url:
+            with urllib.request.urlopen(self.url) as response:
```
Cadair: I haven't made this async, which is ugly. However, it seems jarring to have async properties?

How would people recommend doing this?

opsdroid/message.py
jacobtomlinson: We should probably be using proper warnings (`warninings.warn`).

opsdroid/events.py
jacobtomlinson: Not sure if this will work but you could create a private coroutine which gets awaited in the property.

```python
    @property
    def file_bytes(self):
        """The `bytes` representation of this file."""
        return await self._file_bytes()

    async def _file_bytes(self):  # noqa: D401
        # Change to async stuff...
        return self._file_bytes
```

jacobtomlinson:
This looks great. If you can address the open questions, fix the conflicts I think I'm ready to merge.

opsdroid/events.py
Cadair: You can't await not in a `async def`. It's probably just best to implement this as a async property?

jacobtomlinson: Possibly. This is a new one to me. Best to google about. I might raise a Stack Overflow question.

jacobtomlinson: https://stackoverflow.com/questions/54984337/how-should-you-create-properties-when-using-asyncio

jacobtomlinson: So it looks like PEP8 says don't do IO stuff in properties. Therefore negating the question.

Cadair: I will change it to a method and to async.

opsdroid/connector/__init__.py
SolarDrew: Yes, let's assume that was my reasoning.... Definitely not because of my limited understanding of the different behaviours of the two options...

opsdroid/connector/__init__.py
```diff
@@ -92,24 +165,49 @@ def configuration(self):
             bool: True for message successfully sent. False otherwise.
 
         """
-        _LOGGER.debug(_("%s connector can't react to messages"), self.name)
-        return False
+        warnings.warn(
+            "Connector.react is deprecated. Use "
+            "Connector.respond(events.Reaction(emoji)) instead.",
```
SolarDrew: Is `Connector.respond()` not also deprecated, as per the [warning above](https://github.com/opsdroid/opsdroid/pull/826/files#diff-cf8aeca0533e0e524ca35bb33398ad42R134)?

opsdroid/connector/facebook/__init__.py
```diff
@@ -34,9 +34,7 @@ def __init__(self, config, opsdroid=None):
         """Connector Setup."""
         super().__init__(config, opsdroid=opsdroid)
         _LOGGER.debug("Starting facebook connector")
-        self.config = config
```
SolarDrew: I assume this is now being dealt with elsewhere or isn't used any more?

opsdroid/core.py
```diff
@@ -431,3 +450,23 @@ def start_databases(self, databases):
...
+        `OpsDroid.default_connector`. If ``event.connector`` is a string, it
+        will be resolved to the name of the connectors configured in this
+        instance.
+
+        Args:
+            event (opsdroid.events.Event): The event to send.
+
+        """
+        if isinstance(event.connector, str):
+            event.connector = self._connector_names[event.connector]
```
SolarDrew: Overwriting `event.connector` ?

opsdroid/connector/facebook/__init__.py
Cadair: it's set in the base class.

opsdroid/core.py
Cadair: yes, it converts it to an actual object rather than the string representation thereof.

Cadair:
I think I have responded to all the review comments here.

Cadair:
Whooottt!! Thanks.

jacobtomlinson:
No thank you! This has been a large undertaking!

Issue #875, opened 2019-03-13
regex parser enhancement : Adding fullmatch and match capabilities


iobreaker:
This have been suggested as an enhancement in #872 .
The goal here is to allow users to choose what kind of regex matching they want to play against received message. for the moment the regex only allow search method.

Proposition as explained here in [python re module](https://docs.python.org/3/library/re.html) docs  :

- **search** : Scan through string looking for the first location where the regular expression pattern produces a match, and return a corresponding match object

- **match** : If zero or more characters at the beginning of string match the regular expression pattern, return a corresponding match object

- **fullmatch** : If the whole string matches the regular expression pattern, return a corresponding match object

This will add an other layer of security and give the user the ability to execute a skill only if the message match exactly the patten.

@jacobtomlinson proposed to use the existing kwarg to pass this new parameter.

Regards

iobreaker:
Hi,

Please find bellow the proposed solution (do not pay attention to the many _LOGGER.debug it's just for testing before the releasing)

> matchers.py

```python
def match_regex(regex, case_sensitive=True, matching_condition="search", score_factor=None):
    """Return regex match decorator."""
    def matcher(func):
        """Add decorated function to skills list for regex matching."""
        func = add_skill_attributes(func)
        func.matchers.append(
            {"regex": {
                "expression": regex,
                "case_sensitive": case_sensitive,
                "matching_condition": matching_condition,
                "score_factor": score_factor or REGEX_SCORE_FACTOR,
            }}
        )
        return func
    return matcher
```

> regex.py

```python
async def match_regex(text, opts):

    def is_case_sensitive():
        if opts["case_sensitive"] == True:
            return False
        else:    
            return re.IGNORECASE

    _LOGGER.debug("matching_condition     = " + str(opts["matching_condition"].lower()))
    _LOGGER.debug("case_sensitive         = " + str(is_case_sensitive()))
    _LOGGER.debug("score_factor           = " + str(opts["score_factor"]))
    _LOGGER.debug("text                   = " + text)
    _LOGGER.debug("expression             = " + str(opts["expression"]))
    _LOGGER.debug("")
    if opts["matching_condition"].lower() == "match":
        return re.match(opts["expression"], text, is_case_sensitive())
    elif opts["matching_condition"].lower() == "fullmatch":
        return re.fullmatch(opts["expression"], text, is_case_sensitive())
    else:
        return re.search(opts["expression"], text, is_case_sensitive())



[...]

Issue #872, opened 2019-03-09
Regex parser : No parsed group detected when using two skills with regex_matcher


iobreaker:
Hi

# Description

When activating more than one skill using regex  (example : hello skill and welcome skill or loudnoise skill) if the second skill use group name in regex parser, no group are detected and passed into **message.regex.group**

## Steps to Reproduce

1- clone the last version of opsdroid
2- activate regex parser in config file
3- activate hello skill in config file
4- activate welcome skill in config file (parsing sentence like : welcome iobreaker to paris)
5- Activate telegram connector (I tested using telegram)
6- send : welcome iobreaker to paris
7- error will appear in opsdroid logs

## Expected Functionality
- Groups should be present in **message.regex.groups()**

## Experienced Functionality

You can use this skill (welcome skill) i created to facilitate the debug

```python
from opsdroid.matchers import match_regex

import logging

_LOGGER = logging.getLogger(__name__)

@match_regex(r'(welcome)(\s*)(?P<name>\w+)(\s*)(to|at)(\s*)(?P<location>\w+)', case_sensitive=False)
async def welcome(opsdroid, config, message):
	_LOGGER.debug("=========> triggering welcome skill")
	_LOGGER.debug("=========> group('name')     = {}".format(message.regex.group("name")))
	_LOGGER.debug("=========> group('location') = {}".format(message.regex.group("location")))
```

When sending the message **welcome iobreaker to paris** without activating hello skill all works fine

```
INFO opsdroid.web: Started web server on http://127.0.0.1:8080
DEBUG opsdroid.connector.telegram: {'update_id': 246644582, 'message': {'message_id': 286, 'from': {'id': 111222555, 'is_bot': False, 'first_name': 'IOBreaker', 'language_code': 'en'}, 'chat': {'id': 111222555, 'first_name': 'IOBreaker', 'type': 'private'}, 'date': 1552162058, 'text': 'welcome hicham to paris'}}
DEBUG opsdroid.core: Parsing input: welcome hicham to paris
DEBUG opsdroid.core: Processing parsers...
DEBUG opsdroid-modules.skill.welcome: =========> triggering welcome skill
DEBUG opsdroid-modules.skill.welcome: =========> group('name')     = hicham
DEBUG opsdroid-modules.skill.welcome: =========> group('location') = paris
```

When activating the hello skill and the welcome skill the error happened : **IndexError: no such group**

```
INFO opsdroid.web: Started web server on http://127.0.0.1:8080
[...]

PR #877, opened 2019-03-15
Adding fullmatch and match capabilities to regex parser #875

iobreaker:
This have been suggested as an enhancement in #872 .
The goal here is to allow users to choose what kind of regex matching they want to play against received message. for the moment the regex only allow search method.

This will add an other layer of security and give the user the ability to execute a skill only if the message match exactly the patten.

@jacobtomlinson proposed to use the existing kwarg to pass this new parameter.

Enhancement #872

## Status
**READY** | **~~UNDER DEVELOPMENT~~** | **~~ON HOLD~~**

## Type of change

- Adding param matching_condition to regex matcher
- Adapting regex parser to manage matching conditions
- Adding Debug logs
- This change requires a documentation update -> **Need to be done on [opsdroid documentation](https://opsdroid.readthedocs.io/en/stable/matchers/regex/)**

# How Has This Been Tested?

I implemented the changes and tested against an updated hello skills

```python
import logging
import random


@match_regex(r'hi|hello|hey|hallo', case_sensitive=False, matching_condition="fullmatch")
async def hello(opsdroid, config, message):
    text = random.choice(["Hi {}", "Hello {}", "Hey {}"]).format(message.user)
    await message.respond(text)

@match_regex(r'bye( bye)?|see y(a|ou)|au revoir|gtg|I(\')?m off', score_factor=0.8, matching_condition="match")
async def goodbye(opsdroid, config, message):
    text = random.choice(["Bye {}", "See you {}", "Au revoir {}"]).format(message.user)
    await message.respond(text)
```

# Documentation update proposition

Example 3:

You can specify to the regex matcher wich kind of matching you want to apply, this can be done through  **matching_condition** kwarg. 

**matching_condition** is optional and defaults to search.

You can set  **matching_condition** to :
 
- **search** :  Scan through string looking for the first location where the regular expression pattern produces a match.
- **match** : Scan through string If zero or more characters at the beginning of string match the regular expression pattern
- **fullmatch** : Scan and check If the whole string matches the regular expression pattern

let's take this skill and play different matching conditions against it 

```python
import logging
import random

@match_regex(r'hi|hello|hey|hallo', case_sensitive=False, matching_condition="fullmatch")
async def hello(opsdroid, config, message):
    text = random.choice(["Hi {}", "Hello {}", "Hey {}"]).format(message.user)
    await message.respond(text)
```

- hello skill with matching_condition='search'

```
[6:13:11 PM] HichamTerkiba:
Hi opsdroid

[6:13:12 PM] opsdroid:
Hi HichamTerkiba
```

- hello skill with matching_condition='match'

```
[6:13:11 PM] HichamTerkiba:
Hi opsdroid

[6:13:12 PM] opsdroid:
Hi HichamTerkiba

[6:13:11 PM] HichamTerkiba:
i said Hi opsdroid

[6:13:12 PM] opsdroid:
-No answer from opsroid-
```

- hello skill with matching_condition='fullmatch'

```
[6:13:11 PM] HichamTerkiba:
Hi

[6:13:12 PM] opsdroid:
Hi HichamTerkiba

[6:13:11 PM] HichamTerkiba:
Hi!

[6:13:12 PM] opsdroid:
-No answer from opsroid-
```

> Please feel free to modify/add/update the doc part if neded

# Checklist:

- [ ] I have made corresponding changes to the documentation (Need to be Done)

### DIFF
--- a/docs/matchers/regex.md
+++ b/docs/matchers/regex.md
@@ -16,77 +16,100 @@ parsers:
 
 This is the simplest matcher available in opsdroid. It matches the message from the user against a regular expression. If the regex matches then the function is called.
 
-_note: The use of position anchors(`^` or `$`) are encouraged when using regex to match a function. This should prevent opsdroid to be triggered with every use of the matched regular expression_
+You can specify to the regex matcher which kind of matching you want to apply through the `matching_condition` kwarg (*match* matching is the default). This kwarg should give you more control on how to use the regex matcher.
 
-## Example 1
+Matching conditions:
 
-```python
-from opsdroid.skill import Skill
-from opsdroid.matchers import match_regex
+- **search** - Scans through the string looking for the first location where the regular expression pattern produces a match.
+- **match** - Scans through the string looking at the beginning of the string to match the regular expression pattern.
+- **fullmatch** - Scans and checks if the whole string matches the regular expression pattern.
 
-class MySkill(Skill):
-    @match_regex('hi', case_sensitive=False)
-    async def hello(self, message):
-        await message.respond('Hey')
-```
+### Example 1 - Search condition
 
-The above skill would be called on any message which matches the regex `'hi'`, `'Hi'`, `'hI'` or `'HI'`. The `case_sensitive` kwarg is optional and defaults to `True`.
-
-## Example 2
+Let's use the following hello skill with the kwarg `matching_condition="search"`.
 
 ```python
-from opsdroid.skill import Skill
 from opsdroid.matchers import match_regex
+import random
 
-class MySkill(Skill):
-@match_regex('cold')
-async def is_cold(self, message):
-    await message.respond('it is')
+@match_regex(r'hi|hello|hey|hallo', case_sensitive=False, matching_condition="search")
+async def hello(opsdroid, config, message):
+    text = random.choice(["Hi {}", "Hello {}", "Hey {}"]).format(message.user)
+    await message.respond(text)
 ```
 
-The above skill would be called on any message that matches the regex `'cold'` .
+When we run opsdroid and send the message *"Hi opsdroid"* opsdroid will match the hello to the skill and reply immediately.
 
-> user: is it cold?
->
-> opsdroid: it is
+```
+[6:13:11 PM] HichamTerkiba:
+Hi opsdroid
 
-Undesired effect:
+[6:13:12 PM] opsdroid:
+Hi HichamTerkiba
+```
 
-> user:  Wow, yesterday was so cold at practice!
->
-> opsdroid: it is.
+### Example 2 - match condition
 
-Since this matcher searches a message for the regular expression used on a skill, opsdroid will trigger any mention of the `'cold'`. To prevent this position anchors should be used.
-
-#### Fixed example
+Let's use the following hello skill with the kwarg `matching_condition="match"`.
 
 ```python
-from opsdroid.skill import Skill
 from opsdroid.matchers import match_regex
+import random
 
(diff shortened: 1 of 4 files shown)


iobreaker:
Modification done to pass appveyor and travis-ci checks
Can you please advise for codecov
Thx

iobreaker:
I think codecov is complaining because no testing code in test_parser_regex.py was provided to cover the new regex futures.

I am not an expert in mocking so based on what was done by @jacobtomlinson, perhaps this code will do the job for testing part

```python
    async def test_parse_regex_matching(self):
        with OpsDroid() as opsdroid:
            regex = r"(.*)"

            mock_skill_matching_search = await self.getMockSkill()
            opsdroid.skills.append(match_regex(regex, matching_condition="search")(mock_skill_matching_search))

            mock_skill_matching_match = await self.getMockSkill()
            opsdroid.skills.append(match_regex(regex, matching_condition="match")(mock_skill_matching_match))

            mock_skill_matching_fullmatch = await self.getMockSkill()
            opsdroid.skills.append(match_regex(regex, matching_condition="fullmatch")(mock_skill_matching_fullmatch))

            mock_connector = amock.CoroutineMock()
            message = Message("Hello world", "user", "default", mock_connector)

            skills = await opsdroid.get_ranked_skills(opsdroid.skills, message)
            self.assertEqual(mock_skill_matching_search, skills[0]["skill"])
```

FabioRosado:
Thank you so much for working on this @iobreaker, I think it would be better to divide that test into the three conditions, each test should test only one condition just in case we add something that brakes one and not the other one.

The three tests will be quite similar to one another. But I am going to assume that if you use the test as is only the last one `fullmatch` will be used in the end, but I might be wrong.

As for the documentation update, I am happy with that proposal so feel free to update the regex documentation, if you would rather not doing it I can do that for you 😄

iobreaker:
Hi @FabioRosado,

Thanks for yor reply. I divided unistest into 3 functions and did the same thing for unitest function that test the score_factor to consistent.
Can you please test the code  and feel free to modify/add/update.

if you don't mind, can you make the doc changes please ?

FabioRosado:
I'll try work on the documentation tomorrow. Did you choose the default regex mode to be `search` for some reason instead of `match`? I'm asking this because the default mode was `match` not sure if this will cause some confusion to people that are used to have regex using the match mode 🤔

iobreaker:
Oh no, i thought it was search. change done
Thnaks @FabioRosado

FabioRosado:
I was doing some testing and using a silly function to check  if wildcards will still work if no condition was set it was basically something like this:

```python
@match_regex(r'test (.*)')
async def test(opsdroid, config, message):
	wildcard = message.regex.group(1)
	await message.respond("the wildcard was {}".format(wildcard))
```

When I sent the message *test this thing* opsdroid would only trigger the `hello` message since *hi* is present on the words *this* and *thing*. So using the `match` as a default could avoid breaking things for users that haven't used the kwarg

iobreaker:
I am correcting the coverage, when all will be pointed green, you can go ahead

iobreaker:
Hi @jacobtomlinson 
Can you please restart the coverage tests, i got 100% coverage in my computer but i do not know why here tests are not ok !
Thanks

iobreaker:
Hi @FabioRosado 
you rocks, thanks for the updated regex.md doc file, your idea works great.
now cov is 100% :-)

FabioRosado:
No worries @iobreaker glad I could help 😄 👍 
Thanks again for working on this issue!

PR #883, opened 2019-03-26
Clean exit for Matrix connector

jerrykan:
# Description
When a user was pressing ctrl+c the Matrix connector was catching the `CancelledError` as part of the catchall `except` and just logging it to the console. The next (and subsequent) loops through would raise a `RuntimeError('Session is closed')` resulting in an infinite loop logging the same error to the console over and over.

By correctly catching the `CancelledError` we can ensure that the Matrix connector exits cleanly.

## Status
**READY** ?? not sure if this is being handled at the correct abstraction level or not. Also not sure how to change the test suite to test for this.

## Type of change

# How Has This Been Tested?
Tested manually by connecting to a Matrix homeserver then issuing a ctrl+c
Python 3.7.2

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -2,6 +2,7 @@
 
 import re
 import logging
+from concurrent.futures import CancelledError
 
 import aiohttp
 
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Thanks for this. The very broad except worries me, in the long term I would be keen to remove it, but this helps for now.

PR #887, opened 2019-03-30
Suppress Matrix timeout errors in console (fixes #835)

### LINKED ISSUE #835: Matrix Connector should not log HTML page contents on a 504
If the matrix server responds with a 504 error on `/sync` (which is pretty common) the connector should not log the whole HTML page, as it clutters the log.

jerrykan:
# Description
The Matrix connector will often receive a timeout error when doing a sync request. The catch-all except would just dump the entire HTML of the timeout response to the console. This output can be quite voluminous and not that useful. By checking for know timeout error codes we can suppress the useless HTML output and just log the occurrence at the info level for those that are interested to know when timeouts occur.

In addition to the standard `504 Gateway Timeout` error we also check for `522 Connection timed out` and `524 A Timeout Error` which are timeout errors specific to Cloudflare. See:

  https://support.cloudflare.com/hc/en-us/articles/115003011431/

We check for these Cloudflare errors because the largest Matrix host (matrix.org) is fronted by Cloudflare.

Fixes #835 

## Status
**READY**

## Type of change

# How Has This Been Tested?
As far as I can tell there are no existing tests covering this part of the codebase, but the changes have been tested against the matrix.org homeserver using Python 3.7.2+ (Debian testing).

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -142,6 +142,14 @@ class ConnectorMatrix(Connector):
                 message = await self._parse_sync_response(response)
                 await self.opsdroid.parse(message)
 
+            except MatrixRequestError as mre:
+                # We can safely ignore timeout errors. The non-standard error
+                # codes are returned by Cloudflare.
+                if mre.code in [504, 522, 524]:
+                    _LOGGER.info('Matrix Sync Timeout (code: %d)', mre.code)
+                    continue
+
+                _LOGGER.exception('Matrix Sync Error')
             except CancelledError:
                 raise
             except Exception:  # pylint: disable=W0703


jerrykan:
Now that we are handling `MatrixRequestError` exceptions, it is tempting to remove the catch-all `except Exception`, but I'm not familiar enough with the codebase to make this call.

opsdroid/connector/matrix/connector.py
```diff
@@ -142,6 +142,14 @@ def filter_json(self):
...
                 await self.opsdroid.parse(message)
 
+            except MatrixRequestError as mre:
+                # We can safely ignore timeout errors. The non-standard error
+                # codes are returned by Cloudflare.
+                if mre.code in [504, 522, 524]:
+                    _LOGGER.info('Matrix Sync Timeout (code: %d)', mre.code)
+                    continue
+
+                _LOGGER.exception('Matrix Sync Error')
```
FabioRosado: I'd remove this `_LOGGER` and change the above one from `.info` to `.exception` since it will be logged even if the level is set to info. This second log is a bit redundant as well since we already got the information that there was a Matris Sync Timeout

FabioRosado:
The broad exception has been used since the beginning of opsdroid, I believe it was implemented in order to catch any exception that doesn't fit the main one. This is the reason why it was added, I'm not entirely sure if it's something that we need.

Although this is a great catch and it does make sense to add a different catch for the timeout errors, also, like you said it's good to avoid filling up the terminal with the whole html file 👍

jacobtomlinson:
There needs to be a broad exception catch somewhere to stop broken skills from killing the whole opsdroid process. However I'm happy to discuss whether it would be more appropriate to have it somewhere else.

opsdroid/connector/matrix/connector.py
jerrykan: I've set the logging of timeouts to the `INFO` level because they are pretty common and don't really cause any problems (ie. not really an error). They are also not really exceptions because they have been handled correctly. If I set my logging level at `ERROR` ([which includes `.exception()`](https://docs.python.org/3.7/library/logging.html#logging.Logger.exception)) then I really only want to know about errors that are potentially going to cause problems. There could be an argument made for logging them at the `WARNING` level, but `INFO` is just what I picked at the time.

The second `_LOGGER.exception('Matrix Sync Error')` is to report any `MatrixRequestError` exceptions that haven't been correctly handled (ie. errors that aren't timeouts). If we remove it then more serious errors won't be reported and will go unnoticed.

FabioRosado: That makes a lot of sense, thanks for taking the take to explain. I'm happy with the PR so I am going to update the branch and then merge this PR when travis is happy 😄 👍

PR #902, opened 2019-04-18
Fix _constrain_skills method #901

### LINKED ISSUE #901: Constraints remove skills "forever"
# Description
The new constraints feature (which by the way, are an awesome feature) remove a skill "forever" when are activated.

I write "forever" in quotes because the skills will be unavailable until Opsdroid is restarted.

## Steps to Reproduce
- Add any constraint to a skill (for example, `@constrain_users(['alice'])`)
- Send a message from the user `alice`. Opsdroid will answer as expected.
- Send a message from the user `bob`. Opsdroid will not answer, as expected.
- Send again a message from the user `alice`. Unexpectedly, opsdroid will not answer.

## Expected Functionality
Opsdroid should answer when `alice` sends the message after `bob`.

## Experienced Functionality
Opsdroid doesn't answer when `alice` sends the message after `bob`.

## Versions
- **Opsdroid version: 0.14.1**
- **Python version: 3.6.7**
- **OS/Docker version: Ubuntu 18.04**

## Additional Details
The problem seems to be in the methods `parse` and `_constrain_skills` of the `OpsDroid` class.

`parse` calls `_constrain_skills` with `self.skills` as a parameter. `self.skills` is a list, so is passed by reference. Inside `_constrain_skills`, that list is mutated in `skills.remove(skill)`. So we're mutating `self.skills` directly.

anxodio:
# Description

Changes _constraint_skills to create a new skill list avoiding the mutation of the original list.

Fixes #901

## Status
**READY**

## Type of change

# How Has This Been Tested?

- tests/test_constraints.py -> test_constraint_can_be_called_after_skip

# Checklist:

### DIFF
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -422,11 +422,12 @@ class OpsDroid():
             list: A list of the skills which were not constrained.
 
         """
-        for skill in skills:
-            for constraint in skill.constraints:
-                if not constraint(message):
-                    skills.remove(skill)
-        return skills
+        return [
+            skill for skill in skills if all(
+                constraint(message)
+                for constraint in skill.constraints
+            )
+        ]
 
     async def parse(self, message):
         """Parse a string against all skills."""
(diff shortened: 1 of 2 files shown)


anxodio:
AppVeyour failed in the test `test_install_default_remote_module` with a timeout... :thinking:

jacobtomlinson:
Awesome thanks for this!

PR #903, opened 2019-04-19
First attempt at token auth for matrix connector

colonelkrud:
# Description

This will allow the user to define an authentication token in the config for the matrix connector instead of the username and password. This should simplify some situations where the matrix homeserver considers each login as a new device and provide users with a method of revoking authentication to a user without changing their credentials.

Fixes n/a

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Removed password from config file and tested login with token for two separate users. (need to run actual tests)

# Checklist:

### DIFF
--- a/docs/connectors/matrix.md
+++ b/docs/connectors/matrix.md
@@ -16,6 +16,7 @@ connectors:
     # Required
     mxid: "@username:matrix.org"
     password: "mypassword"
+    token: "user-access-token" #If token is desired, remove password else password will be used.
     # Name of a single room to connect to
     room: "#matrix:matrix.org"
     # Alternatively, a dictionary of multiple rooms
--- a/docs/connectors/matrix.md
+++ b/docs/connectors/matrix.md
@@ -16,7 +16,7 @@ connectors:
     # Required
     mxid: "@username:matrix.org"
     password: "mypassword"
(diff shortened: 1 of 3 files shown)


Cadair:
Looks good to me. You could add a unit test that checks login isn't called if a token is provided.

The only other thing that might be worth doing is checking what happens if the token is invalid.

oleg-fiksel:
Works for me with the following config:

```
connectors:
  - name: matrix
    mxid: "@testuser:mymatrixdomain.com"
    token: "**********************************************"
    homeserver: "https://my-matrix.server.com"

## Configure the web server
web:
  host: '0.0.0.0'
  port: 8080

parsers:
  - name: dialogflow
    access-token: "*******************"

skills:
  - name: dialogflow-test
    path: /skills/dialogflow-test
```

colonelkrud:
Yep. It’s been working on my bridge as well. Just need to update the matrix connector test file to check for a few things but I’ve been away for a bit. If someone wants to take a shot at the tests, I’ll gladly merge them :p 

Also need to make sure the docs are all up to date. I think I got them all but there might be a default config that needs to be updated.

FabioRosado:
Hey guys is it possible to get my PR merged so we update this one and we can merge is issue soon? Thanks 😄

PR #922, opened 2019-05-05
Add parse_format matcher

anxodio:
# Description

Add a new simple parser based on https://github.com/r1chardj0n3s/parse .

It is like the regex parser but simpler, useful for newcomers that don't need all the possibilities (and complexity) of the regular expressions.

Look the new doc page or the tests to see some examples of use :blush: 

## Status
**READY**

## Type of change

# How Has This Been Tested?

- A new test class in `tests/test_parser_format.py`

# Checklist:

### DIFF
--- a/docs/matchers/parse_format.md
+++ b/docs/matchers/parse_format.md
@@ -1,0 +1,163 @@
+# Parse Format Matcher
+
+## Configuring opsdroid
+
+As in [regex matcher](/matchers/regex.md), in order to enable parse format skills, you must set the `enabled` parameter to true in the parsers section of the opsdroid configuration file.
+
+```yaml
+parsers:
+  - name: parse_format
+    enabled: true
+```
+
+## About Parse Format Matcher
+
+This is the simplest matcher available in opsdroid. It matches the message from the user against a string with python [format syntax](https://docs.python.org/3/library/string.html#format-string-syntax). If the string matches then the function is called.
+
+The third party library [parse](https://github.com/r1chardj0n3s/parse) is used to parse the strings.
+
+You can specify which kind of matching you want to apply through the `matching_condition` kwarg (*match* matching is the default). Be careful, the matching condition names are the same than in regex but there are some differences between them.
+
+Matching conditions:
+
+- **match** - Scans through the string looking at the beginning of the string to match the pattern, **but the end must match too**. So if you have `hi` it will match `hi`, but no `say hi` or `hi!`.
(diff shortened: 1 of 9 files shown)


docs/matchers/parse_format.md
```diff
@@ -0,0 +1,163 @@
...
+import random
+
+class MySkill(Skill):
+    @match_parse('hi', case_sensitive=False, matching_condition='search')
+    async def hello(self, message):
+        text = random.choice(['Hi {}', 'Hello {}', 'Hey {}']).format(message.user)
+        await message.respond(text)
+```
+
+With the matching condition kwarg set to search and case sensitive set to `False`, opsdroid will be more fleixble.
```
FabioRosado: Typo on flexible

docs/matchers/parse_format.md
```diff
@@ -0,0 +1,163 @@
...
+import random
+
+class MySkill(Skill):
+    @match_parse('hi')
+    async def hello(self, message):
+        text = random.choice(['Hi {}', 'Hello {}', 'Hey {}']).format(message.user)
+        await message.respond(text)
+```
+
+When we run opsdroid and send the message *"hi"* opsdroid will match the hello to the skill and reply immediately. Opsdroid will only trigger the skill if *hi* is only word in the message.
```
FabioRosado: Maybe add `the` before "only word" in the message. So the text will be:  `if *hi* is the only word in the message`

docs/matchers/parse_format.md
```diff
@@ -0,0 +1,163 @@
...
+    @match_parse('remember {}')
+    async def remember(self, message):
+        remember = message.parse_result[0]
+        await self.opsdroid.memory.put("remember", remember)
+        await message.respond("OK I'll remember that")
+```
+
+### Named Fields
+
+Instead of using empty brackets and then access by position you can assign them a name and access by name.
```
FabioRosado: Add the article between by and position - `Instead of using empty brackets and then access by the position you can assign them a name and access by name.`

anxodio:
Thank you for the English fixes @FabioRosado :smile: 

> I was wondering if you had a reason to change the Regex const variable?

My reason is that this constant is used now for regex and parse_format scoring, so IMHO it's better to have it reflected in the name.
If we keep `REGEX_SCORE_FACTOR`, we can make changes thinking that they will only impact regex skills, but we will be impacting parse skills collaterally.

Another idea: have two different constants, one for regex and one for parse :thinking:

FabioRosado:
I've missed that bit but yeah it makes sense! I think it would be okay to just have one const variable since they are related to one another - since this is still somewhat related to the regex parser just not using regex haha

PR #931, opened 2019-05-11
WIP Add Slack Image and File Events

jacobtomlinson:
# Description

Add Image and File event support to Slack connector.

## Status
**ON HOLD**

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:

### DIFF
--- a/opsdroid/connector/slack/__init__.py
+++ b/opsdroid/connector/slack/__init__.py
@@ -11,7 +11,7 @@ from aioslacker import Slacker
 from emoji import demojize
 
 from opsdroid.connector import Connector, register_event
-from opsdroid.events import Message, Reaction
+from opsdroid.events import Message, Reaction, File, Image
 
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
In order to download images from slack you need to provide the access token in a header. Therefore I've extended the `File` event to take some extra headers. 

However I'm now stuck because the token I was developing with doesn't have file permissions, which results in an HTML login page being downloaded instead of the file. This could be because the request is following a redirect?

opsdroid/connector/slack/__init__.py
```diff
@@ -193,3 +225,12 @@ def __init__(self, config, opsdroid=None):
             message = message.replace("<@{userid}>".format(userid=userid),
                                       user_info["name"])
         return message
+
+    async def check_for_files_access(self):
+        """Check whether bot token has access to Slack files."""
+        try:
+            await self.slacker.files.list()
+        except slacker.Error as error:
```
jacobtomlinson: In Python 3.5.7 this exception is not being caught. This is due to a bug that is not present in 3.6.8. I'm planning on putting this PR on hold until #953 has been discussed thoroughly.

PR #938, opened 2019-05-15
Attempt at fixing appveyor fails on 3.5

FabioRosado:
I have no idea if this will even work or affect something, but since pip is really outdated on python 3.5 I wonder if this could help fix the issue 🤔


FabioRosado:
Didn't work so closing

PR #940, opened 2019-05-15
Fix regex and parse_format matchers

### LINKED ISSUE #934: Regex.group broke after events got merged into core
# Description
After the events PR was merged to Core an `AttributeError` exception is raised when we try to get a regex group.

The issue stars in `opsdroid.core` on [line 467](https://github.com/opsdroid/opsdroid/blob/d1d25927107892008f62e6bf83d4f82fb994736c/opsdroid/core.py#L467)

And possibly something in `opsdroid.events` is not parsing the regex groups.

## Steps to Reproduce
Have a skill that tries to get a regex group from a sentence.

## Expected Functionality
Opsdroid should be able to use that group in the reply.

## Experienced Functionality
An `AttributeError` exception is raised

_Note: This is a very broad issue. I'll be more than happy to dig deeper into it if needed._

Cadair:
# Description

This fixes a regression in 0.15.

I decided to apply both the fixes, I can revert one or the other if people want.

Fixes #934

## Status
**READY**

## Type of change

# How Has This Been Tested?

unit tests

# Checklist:

### DIFF
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -464,7 +464,7 @@ class OpsDroid():
                     self.eventloop.create_task(
                         self.run_skill(ranked_skills[0]["skill"],
                                        ranked_skills[0]["config"],
-                                       event)))
+                                       ranked_skills[0]["message"])))
 
         return tasks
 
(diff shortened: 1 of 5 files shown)


FabioRosado:
I have tested these changes with my skill that generates images and uses regex group to get the details passed to opsdroid and everything worked fine 👍

PR #943, opened 2019-05-15
Make web server open by default

### LINKED ISSUE #942: Make web server default to all networks
Currently the opsdroid web server is only served on `127.0.0.1` by default. In order to make opsdroid accessible via other networks you much set the `web.host` config option to something else (usually `0.0.0.0`).

This can be misleading for new users and is also causing problems when running on more complex infrastructure like Kubernetes.

I propose that the default is changed to `0.0.0.0` which resolves up front issues but still allows users to lock things down to specific networks if they choose.

jacobtomlinson:
# Description

Change default web server host to `0.0.0.0` to allow all traffic.

Fixes #942

## Status
**READY** 

# Checklist:

### DIFF
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -282,11 +282,11 @@ lang: <ISO 639-1 code -  example: 'en'>
 
 Configure the REST API in opsdroid.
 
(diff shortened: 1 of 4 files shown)


FabioRosado:
This looks great I'm happy to merge it 😄 👍

PR #946, opened 2019-05-15
Add long description content type in setup.py

### LINKED ISSUE #944: PyPI deployments are failing
Looks like PyPI deployments are failing. `v0.15.1` and `v0.15.2` haven't gone out.

```
HTTPError: 400 Client Error: The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information. for url: https://upload.pypi.org/legacy/
```

anxodio:
# Description

I'm not 100% sure, but this could fix #944 . See https://pypi.org/help/#description-content-type :
> PyPI will reject uploads if the description fails to render.

Fixes #944

## Status
**READY**

## Type of change

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,7 @@ setup(
     author_email='jacob@tom.linson.uk',
     description='An open source ChatOps bot framework.',
     long_description=README,
+    long_description_content_type='text/markdown',
     packages=PACKAGES,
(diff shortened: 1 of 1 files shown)


anxodio:
I tested it with https://github.com/pypa/readme_renderer and seems that this fixes the problem :tada:

PR #951, opened 2019-05-18
[WIP] Events, phase II

SolarDrew:
# Description

This PR introduces some new and different Event types, in an effort to start branching out from standard message-type events that can be sent _in_ a chat room (`Message`, `Image`, etc) to more meta events that can change properties of rooms (things like setting a room avatar, topic, or permissions).

The general motivation is to be able to use opsdroid for community-management type tasks, creating and editing rooms and inviting users, and so on, as well as simply interacting with users in a small, pre-defined set of rooms.

Closes https://github.com/SolarDrew/skill-picard/issues/26.

## Status
**UNDER DEVELOPMENT**

## Type of change

# How Has This Been Tested?

No tests as yet.

# Checklist:


opsdroid/events.py
```diff
@@ -286,3 +288,35 @@ class Image(File):
...
+
+class RoomCreation(Event):
+    """Event class to represent the creation of a new room."""
+
+    def __init__(self, public=True, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.is_public = public
+
+
+class SetRoomAlias(Event):
```
Cadair: maybe use "address" or something a little more non-matrixy here?

opsdroid/events.py
```diff
@@ -286,3 +288,35 @@ class Image(File):
...
+class SetRoomTopic(Event):
+    def __init__(self, room_topic, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.room_topic = room_topic
+
+
+class SetRoomImage(Event):
+    def __init__(self, room_image, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.room_image = room_image
```
Cadair: should be checked as an `events.Image` object?

opsdroid/events.py
```diff
@@ -286,3 +288,35 @@ class Image(File):
         """Return the image dimensions `(w,h)`."""
         fbytes = await self.get_file_bytes()
         return get_image_size_from_bytesio(io.BytesIO(fbytes), len(fbytes))
+
+
+class RoomCreation(Event):
+    """Event class to represent the creation of a new room."""
+
+    def __init__(self, public=True, *args, **kwargs):
```
Cadair: How matrix specific is the concept of public rooms? Could we have some kind of generic room properties key or something?!

SolarDrew: Yeah, this is a problem I just generally wasn't sure how to approach, I suspect there are going to be a lot of properties some platforms' rooms have and others' don't. Room properties makes sense, then it could just be a dict with whatever in it. 

Either way I think the event should be lenient on taking properties and it should be up to the connector to know how to deal with them.

opsdroid/events.py
SolarDrew: Yeah, wasn't sure what the more generic term would be. But I also wanted to avoid ambiguity in what this event actually does, if you're in matrix-only land. Possibly this could be a use-case for connector-specific events in core?

SolarDrew:
Nooooo, not the stale bot!

jacobtomlinson:
How are we doing here? Any chance this can be picked up again?

Cadair:
Hopefully we can spend some time tidying it next week. Are you OK with leaving it as one big PR?

jacobtomlinson:
It's up to you. If you break it up I'm more likely to review things quicker.

SolarDrew:
This became a great big ugly PR and went very very stale so I've replaced it with #1069, #1070, and #1071. Closing.

PR #966, opened 2019-05-27
Add filters to logging

### LINKED ISSUE #954: Add filters to logging
Currently, opsdroid has a lot of things logging into debug mode, which makes it hard to try and find things while doing the actual debugging.

It would be good to implement filters that could be added to the logging through the `config.yaml` file. 

After sharing this in the channel Jacob suggested to implement filters by creating a class that goes through all those logs and then use the config to add a list of things to filter and that class would check every log against that list and either show that log or drop it.

FabioRosado:
# Description

I have decided to create a new `logging.py` file because the logging related things were getting pretty long - I'm not sure if this is something you would like to do, if not I can just revert it back to the `__main__` file.

In the `logging.py` file I have added the filtering class and the logic to handle whitelisting and blacklisting names, but I have also added an `extended` argument that adds the function name of where the log was called, the reason why I decided to add this was to allow users to have a finer control of what to show in the logs and what to skip.

This PR is still in progress but would like to know what do you think about it so far @jacobtomlinson

--EDIT--
I have marked this PR as ready for review, I refactored the logging code and added tests that bump the coverage of `logging.py` to 100%, although for some reason lines 325-326 from `loader.py` are missing coverage for some reason.

I tried to create a test for this Log but everything I tried kept failing (I tried to pass a side_effect to the` load_config_file` and then to the `move_to_appdir` helper but didn't do anything), I'm not sure why would this happen just because I moved some logging from `__main__.py`, if you have any idea let me know please.

--EDIT 2 --

Well, it seems the coverage issue might have been from my end since it doesn't show here so I'm happy haha

Fixes #954

## Status
**READY** | ~~**UNDER DEVELOPMENT**~~ |~~**ON HOLD**~~

## Type of change

# How Has This Been Tested?

- Hand test - whitelist/blacklist items - worked as expected
- Hand test - whitelist/blacklist with extended set - worked as expected
- Tox - tests green - drop in loader coverage o.O

# Checklist:


tests/test_main.py
```diff
@@ -54,70 +53,70 @@ def test_configure_lang(self):
...
-    def test_configure_default_logging(self):
-        config = {}
-        opsdroid.configure_logging(config)
-        rootlogger = logging.getLogger()
-        self.assertEqual(len(rootlogger.handlers), 2)
-        self.assertEqual(logging.StreamHandler, type(rootlogger.handlers[0]))
-        self.assertEqual(rootlogger.handlers[0].level, logging.INFO)
-        self.assertEqual(logging.FileHandler, type(rootlogger.handlers[1]))
-        self.assertEqual(rootlogger.handlers[1].level, logging.INFO)
+    # def test_set_logging_level(self):
```
jacobtomlinson: Can you delete this rather than comment

opsdroid/__main__.py
```diff
@@ -25,6 +25,10 @@
 gettext.install("opsdroid")
 _LOGGER = logging.getLogger("opsdroid")
 
+# class NoParsingFilter(logging.Filter):
```
jacobtomlinson: Probably needs tidying up

opsdroid/logging.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class ParsingFilter(logging.Filter):
+    """Class that filters logs."""
+
+    def __init__(self, config, *parse_list):
+        """Create object to implement filtering."""
+        super(ParsingFilter, self).__init__()
+        self.config = config["logging"]
+        self.parse_list = parse_list[0].get("whitelist") or parse_list[0].get(
```
jacobtomlinson: If the user configures both the whitelist and blacklist the blacklist will be ignored here. I could imagine this happening if you want to only see messages from one place, but it has a noisy subclass.

Either the logic needs updating to handle this. Or a warning should be raised and the docs should reflect that only one at a time can be used.

Happy either way.

FabioRosado: Good catch! Initially, I was thinking to allow both filters on the config, but I quickly realised that this wouldn't make much sense because when you whitelist something like - `opsdroid.core` nothing else other than the logs in `opsdroid.core` will be shown, so it wouldn't make sense to use blacklist with it.

Unless you want to mix the extended to blacklist a specific thing, but yeah not sure if this will work (i don't think you can whitelist opsdroid.core and then blacklist some function name with extended mode)

I'll update the docs and raise a warning.

FabioRosado: I have just tested opsdroid with both whitelist and blacklist and it breaks everything badly haha

PR #973, opened 2019-06-02
Replace Asyncio_redis with aioredis

### LINKED ISSUE #792: Redis database keeps auto-reconnecting
# Description
I have found an issue with the Redis database, if you set Redis on your `config.yaml` but the server is not running, the client will try to reconnect to the server ad infinitum.

This is possibly an issue with `asyncio_redis` and I have raised an issue within the project ([#126](https://github.com/jonathanslenders/asyncio-redis/issues/126)).

I have tried to set `auto_reconnect` to False directly on the `connect` method, but had no change.

The`asyncio_redis` source code uses an infinite loop when trying to reconnect and this is triggered whenever the client fails to connect to the server, it should be turned off if `auto_reconnect` is set to `False` but it doesn't seem to be happening.

Finally, when we try to terminal opsdroid with SIGINT the client is unable to the terminal due to the following exception:

```python
INFO opsdroid.core: Stopping database ...
ERROR opsdroid.core: Caught exception
ERROR opsdroid.core: {'message': 'Task exception was never retrieved', 
'exception': AttributeError("'NoneType' object has no attribute 'close'",), 
'future': <Task finished coro=<OpsDroid.handle_signal() done, defined at /Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/core.py:120> exception=AttributeError("'NoneType' object has no attribute 'close'",)>}
```

And the asyncio_redis will keep that infinite loop until you close the terminal window 😬 

Since this is an issue with `asyncio_redis` there isn't much we can do at the moment, the only thing we could try to do would be fix the `AttributeError` on `opsdroid.core.unload` - perhaps pass if this exception is raised?

## Steps to Reproduce

* Add Redis database to your `config.yaml`
* run opsdroid without starting the Redis database

## Expected Functionality
asyncio_redis shouldn't keep trying to reconnect to the client unless `reconnect` was set to true. This flag doesn't seem to be working as expected.

## Experienced Functionality
Explain what happened instead(Please include the debug log).

## Versions
- **Opsdroid version:** dev
- **Python version:** 3.7
- **OS/Docker version:** macOS 10.14.2

## Configuration File
Please include your version of the configuration file bellow.

```yaml
databases:
   - name: redis

```

FabioRosado:
# Description

I've replaced asyncio_redis with aioredis. In this PR I have:
- updated docker to be able to install aioredis
- move JSONEncoder/Decoder from sqlite.py to helper.py since I am using it in redis database as well
- Fix all the tests
- Update documentation - since reconnect is not in use anymore

I've also tested opsdroid with the redis database set in the config.yaml but without the redis database running. I handled this case with a try/except and only do - get/put/close - if `self.client` is true. Opsdroid will run fine if the database is set but not active,  I've noticed that even though opsdroid works fine, there are issues when `seen` is called since `opsdroid.memory.get` is unable to get anything from the database.

I've also tested opsdroid with a connection to redis, and then killed the redis client - opsdroid works fine and just has issues when trying to put/get things from memory - perhaps we need to handle this?

Let me know what you think of these changes.

Fixes #792 

## Status

## Type of change

- New feature (non-breaking change which adds functionality) (redis actually works)

# How Has This Been Tested?

- Tox - all green
- Run opsdroid with redis database running - working as expected
- Run opsdroid with redis set in config but not running - opsdroid still runs

# Checklist:


jacobtomlinson:
Awesome stuff thanks!

PR #994, opened 2019-07-01
Google docstring fix core

killerontherun1:
# Description

Fixing docstring issues for core.py. 

Fixes #532

## Status

## Type of change

# How Has This Been Tested?

Tested: 

- using the google docstring link  - https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

# Checklist:

- [1] I have performed a self-review of my own code
- [1] I have made corresponding changes to the documentation (if applicable)
- [0] I have added tests that prove my fix is effective or that my feature works
- [0] New and existing unit tests pass locally with my changes

### DIFF
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -35,6 +35,7 @@ _LOGGER = logging.getLogger(__name__)
 
 class OpsDroid:
     """Root object for opsdroid.
+
     """
 
     # pylint: disable=too-many-instance-attributes
@@ -43,7 +44,9 @@ class OpsDroid:
     instances = []
 
     def __init__(self, config=None):
-        """Start opsdroid."""
+        """Start opsdroid.
+
+        """
         self.bot_name = "opsdroid"
         self._running = False
         self.sys_status = 0
@@ -75,7 +78,9 @@ class OpsDroid:
         self.stored_path = []
 
     def __enter__(self):
-        """Add self to existing instances."""
+        """Add self to existing instances.
+        
+        """
         self.stored_path = copy.copy(sys.path)
         if not self.__class__.instances:
             self.__class__.instances.append(weakref.proxy(self))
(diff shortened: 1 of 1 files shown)


opsdroid/core.py
```diff
@@ -1,4 +1,4 @@
-"""Core components of OpsDroid."""
+="""Core components of OpsDroid."""
```
FabioRosado: I think you added a `=` by mistake and this is breaking the tests, could you please delete this?

opsdroid/core.py
```diff
@@ -34,15 +34,19 @@
...
+    """
 
     # pylint: disable=too-many-instance-attributes
     # All are reasonable in this case.
 
     instances = []
 
     def __init__(self, config=None):
-        """Start opsdroid."""
+        """Start opsdroid.
```
FabioRosado: Here you should add the Arg config to the docstring, you could do something like:

```
Args:
    config (dict or None):  Dictionary containing opsdroid configuration.

File: opsdroid/core.py [hunk 3]
```diff
@@ -74,7 +78,9 @@ def __init__(self, config=None):
         self.stored_path = []
 
     def __enter__(self):
-        """Add self to existing instances."""
+        """Add self to existing instances.
+
```
FabioRosado: If you want you could explain what this method does, if you don't wish to do such you can just revert this change back to the previous format.

opsdroid/core.py
```diff
@@ -101,19 +111,25 @@ def default_connector(self):
...
+        """Exit application.
+        
+        """
         _LOGGER.info(_("Exiting application with return code %s"), str(self.sys_status))
         sys.exit(self.sys_status)
 
     def critical(self, error, code):
-        """Exit due to unrecoverable error."""
+        """Exit due to unrecoverable error.
+
```
FabioRosado: Same thing here, critical has `error` and `code` arguments, you should add these to the docstring and explain what this method does - basically it will exit opsdroid due to an unrecoverable error, logs the error message to logging and then exits.

opsdroid/core.py
```diff
@@ -349,7 +385,9 @@ def start_databases(self, databases):
                     self.eventloop.run_until_complete(database.connect())
 
     async def run_skill(self, skill, config, message):
-        """Execute a skill."""
+        """Execute a skill.
+
```
FabioRosado: On this method you should add these Args and explain what they are/do. 
Since there is two lines with comment they should be inside the docstring as well

PR #1002, opened 2019-07-07
Added configuration validation

### LINKED ISSUE #770: Add config validation
Currently we are not doing a very good job of validating config before loading opsdroid.

An example of this is that we have [very old and now invalid configs](https://github.com/opsdroid/opsdroid/blob/bc95a2bd4d7cf921974f4ee01ec1b51003249bc3/tests/configs/minimal.yaml) in our tests and they are still passing. In that config `skills` is a dictionary, which was changed to a list a long time ago.

We should check our configs are valid before starting to load.

kollyuchka:
# Description

Added a validation scheme using the yamale library to check configuration files

Fixes #770

## Status
**READY** 

## Type of change

# How Has This Been Tested?

Two tests have been added to the TestLoader class to validate correct and broken configuration files. Tests that deals with old type of configs were skipped because they fail validation.

* test_load_config_valid (validates the completed configuration file)
* test_load_config_broken (gives an exception when checking an incorrectly filled configuration file)

# Checklist:


FabioRosado:
Thanks for working on this, did you want to close this PR or was it a mistake?

PR #1056, opened 2019-09-01
Upgrade Dialogflow to V2

### LINKED ISSUE #566: Upgrade Dialogflow to V2 API
There is now a V2 API for Dialogflow.

We should look at the parser and skill and upgrade the API if applicable.

https://dialogflow.com/docs/reference/v2-agent-setup

FabioRosado:
# Description

This is my initial work on updating the dialogflow parser to v2, there are a few changes that we needed to address. The recommended way to use dialogflow api is to use the official dialogflow sdk.

The api tokens no longer work so you need to generate a json file used in authentication - the path needs to be set as an envvar.

I've also decided to remove mentions of api.ai since it's been a while since the change of name happened and because this v2 changed a lot of things, let me know if you agree with this change.

Fixes #566

## Status

## Type of change

# How Has This Been Tested?

- Tox - all green
- Run opsdroid with dialogflow set and a test matcher for intent and actions - all working ok
- Run opsdroid with Portuguese set as language and English on parser - parses English
- Run opsdroid with no language set but parser with Portuguese - parses Portuguese

# Checklist:


FabioRosado:
I've found a way to get dot notation from a dict on stackoverflow so it kind of works on the tests. For some reason coverage is complaining that the debug and the return statement is score is less than the one set in config, is not being tested.

Had to touch on the core tests as well to add the fake google authentication envvar. But this test is still failing and I'm not too sure why to be honest.

FabioRosado:
Well I tried to add alpine-sdk because i read on stackoverflow that this might help on docker, comparing to my previous failure it did help since opsdroid ran but it seems we will need to have gcc in order to have dialogflow working.

```python
ImportError: Error loading shared library libstdc++.so.6: No such file or directory (needed by /usr/local/lib/python3.7/site-packages/grpc/_cython/cygrpc.cpython-37m-x86_64-linux-gnu.so)
```

I'm not entirely sure what is the best way to make dialogflow working with v2 and keeping the docker image small. So I will leave this PR open for now and await for feedback, like I said on Matrix we have some time anyway.

jacobtomlinson:
You could try adding `libstdc++` to the docker image and see how much bigger things get?

FabioRosado:
I don't have docker installed on my machine atm but will update the docker file, im going to assume that we will be able to see the docker size then right? 🤔

jacobtomlinson:
Just tested it locally. It adds about 120MB to the image, which is a bit of a pain.

image

FabioRosado:
I was also thinking that since we are installing alpine-sdk just for the gcc we can just remove gcc and musl-dv right? 🤔 

Should we try to do the update some other way, instead of using the oficial dialog dependency?

jacobtomlinson:
My current thoughts on this would be that we should look at creating optional dependency packs. https://www.python.org/dev/peps/pep-0508/#extras

So in this instance, we would add `dialogflow` as an optional dependency in a pack called `dialogflow` which would be installed as `pip install opsdroid[dialogflow]`. Then in the code we would catch the import error that would be raised if the dependency is missing, warn the user that they need to install it to use dialogflow and then disable the parser.

This probably means we would need to think about having multiple docker images too which match the dependency packs.

**Alternatively** we could just accept that opsdroid has a big container image and move on.

What do you think?

FabioRosado:
I think in the future we might have to accept that the docker image will be larger than it is. The reason for this is that this is the second dependency that requires gcc - I don't remember which one needed it before but we went around it by installing and then uninstalling it from the docker image.

With that in mind, I think the extra package is a great solution to this problem. Not everyone will choose to have dialogflow active and this could give users a better way to manage opsdroid since they can install different modules. One project that comes to mind is typography and the different themes available - all you need to do is install them with npm. 

Now I haven't come across  dependency packs yet so I will need to do a bit of research on how to do this 😄 👍

jacobtomlinson:
Ok great! So just to clarify in order to get this merged lets move the dialogflow dependency out to be optional, but still include it in the Docker image and accept that it will not be bigger.

FabioRosado:
With the last few PRs merged into core and with me touching this PR after a few days passed I've noticed one thing. I tried to run opsdroid with a shell that didn't have the google credentials envvar, I expected a Warning exception to be raised, but this is what happened:

```
INFO opsdroid.logging: Started opsdroid v0.14.1+200.gd63d023.dirty
INFO opsdroid: You can customise your opsdroid by modifying your configuration.yaml
WARNING opsdroid.loader: No databases in configuration.This will cause skills which store things in memory to lose data when opsdroid is restarted.
INFO opsdroid.loader: Cloning hello from remote repository
WARNING opsdroid.core: Function based skills are deprecated and will be removed in a future release. Please use class-based skills instead.
WARNING opsdroid.core: Function based skills are deprecated and will be removed in a future release. Please use class-based skills instead.
ERROR: Unhandled exception in opsdroid, exiting...
Caught exception
{'message': 'Task exception was never retrieved', 'exception': Warning("Authentication file not found or 'project-id' not in configuration, dialogflow parser will not be available",), 'future': <Task finished coro=<OpsDroid.parse() done, defined at /Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/core.py:438> exception=Warning("Authentication file not found or 'project-id' not in configuration, dialogflow parser will not be available",)>}
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/__main__.py", line 190, in <module>
    init()
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/__main__.py", line 187, in init
    main()
  File "/Users/fabiorosado/.local/share/virtualenvs/opsdroid-13bLHlYD/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/fabiorosado/.local/share/virtualenvs/opsdroid-13bLHlYD/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/fabiorosado/.local/share/virtualenvs/opsdroid-13bLHlYD/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/fabiorosado/.local/share/virtualenvs/opsdroid-13bLHlYD/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/__main__.py", line 181, in main
    opsdroid.run()
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/core.py", line 144, in run
    self.eventloop.run_until_complete(asyncio.gather(*pending))
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/asyncio/base_events.py", line 466, in run_until_complete
    return future.result()
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/core.py", line 120, in handle_async_exception
    context["future"].result()
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/core.py", line 448, in parse
    ranked_skills = await self.get_ranked_skills(unconstrained_skills, event)
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/core.py", line 387, in get_ranked_skills
    self, skills, message, dialogflow[0]
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/parsers/dialogflow.py", line 89, in parse_dialogflow
    result = await call_dialogflow(message, config)
  File "/Users/fabiorosado/Documents/GitHub/opsdroid/opsdroid/parsers/dialogflow.py", line 53, in call_dialogflow
    "Authentication file not found or 'project-id' not in configuration, dialogflow parser will not be available"
Warning: Authentication file not found or 'project-id' not in configuration, dialogflow parser will not be available
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/asyncio/base_events.py", line 511, in __del__
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/asyncio/unix_events.py", line 65, in close
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/asyncio/unix_events.py", line 146, in remove_signal_handler
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/signal.py", line 47, in signal
ERROR: Unhandled exception in opsdroid, exiting...
ERROR: Unhandled exception in opsdroid, exiting...
ERROR: Unhandled exception in opsdroid, exiting...
ERROR: Unhandled exception in opsdroid, exiting...
ERROR: Unhandled exception in opsdroid, exiting...
ERROR: Unhandled exception in opsdroid, exiting...
ERROR: Unhandled exception in opsdroid, exiting...
ERROR: Unhandled exception in opsdroid, exiting...
```

What is the best way to raise an exception in this context? 🤔  
Or do you think I should avoid raising a warning and use `logging.warning` instead?

FabioRosado:
I've been trying to make this test pass but haven't have much success. I think the issue is that the ImportError is not being raised with "self.assertRaises", but I tried to mock the call_dialogflow function but that didn't help as well.

I tested the parser on my end and the error message will be logged and the config will be an empty dictionary. 

Do you have any suggestions for me to try and get this import error test working?

Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM python:3.7-alpine
 LABEL maintainer="Jacob Tomlinson <jacob@tom.linson.uk>"
 
-RUN apk add --no-cache gcc musl-dev
+RUN apk add --no-cache gcc musl-dev alpine-sdk
```
jacobtomlinson: Do we need this now that we aren't adding the dependency?

FabioRosado: Actually we don't, i thought this was a dependency that we added with the gcc, my bad! 
I'll update the docker file 😄

jacobtomlinson: Thinking about it we may need it for the tests. Let's see how the CI goes.

PR #1069, opened 2019-09-23
Events IIa - Core additions

### LINKED ISSUE #999: Add a way of running a skill on startup
This could be an event that's emitted on startup or shutdown or a matcher.

It should be triggered after all the connectors and skills have completed their startup etc.

SolarDrew:
# Description

This PR introduces some new and different Event types, in an effort to start branching out from standard message-type events that can be sent _in_ a chat room (`Message`, `Image`, etc) to more meta events that can change properties of rooms (things like setting a room avatar, topic, or permissions).

The general motivation is to be able to use opsdroid for community-management type tasks, creating and editing rooms and inviting users, and so on, as well as simply interacting with users in a small, pre-defined set of rooms.

This PR follows on from (and replaces) #951, which became nebulous and upleasant. Included here are tweaks to the core events functionality and additions of some generic event types. Connector-specific events and changes will follow in separate PRs.

~~Closes- https://github.com/SolarDrew/skill-picard/issues/26.~~
Edit: No it doesn't.

closes #999

## Status
**UNDER DEVELOPMENT**

## Type of change

# How Has This Been Tested?

Coverage extended to the new functionality, nothing breaks.

# Checklist:

### DIFF
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -308,7 +308,7 @@ class OpsDroid:
 
         if connectors:
             for connector in self.connectors:
-                await asyncio.wait(self.eventloop.create_task(connector.connect()))
+                await self.eventloop.create_task(connector.connect())
 
             for connector in self.connectors:
                 task = self.eventloop.create_task(connector.listen())
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -347,7 +347,7 @@ class OpsDroid:
                     _LOGGER.debug(_("Adding database: %s"), name)
                     database = cls(database_module["config"])
                     self.memory.databases.append(database)
-                    self.eventloop.run_until_complete(database.connect(self))
+                    self.eventloop.run_until_complete(database.connect())
 
     async def run_skill(self, skill, config, message):
         """Execute a skill."""
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -114,15 +114,14 @@ class OpsDroid:
     @staticmethod
     def handle_async_exception(loop, context):
         """Handle exceptions from async coroutines."""
+        print("ERROR: Unhandled exception in opsdroid, exiting...")
         if "future" in context:
             try:  # pragma: nocover
                 context["future"].result()
             # pylint: disable=broad-except
             except Exception:  # pragma: nocover
-                _LOGGER.exception(_("Caught exception"))
-        else:
-            _LOGGER.error(_("Caught exception"))
-        _LOGGER.error(context)
+                print("Caught exception")
+                print(context)
 
     def is_running(self):
         """Check whether opsdroid is running."""
@@ -164,6 +163,8 @@ class OpsDroid:
         self.start_connectors(self.modules["connectors"])
         self.cron_task = self.eventloop.create_task(parse_crontab(self))
         self.eventloop.create_task(self.web_server.start())
+
+        self.eventloop.create_task(self.parse(events.OpsdroidStarted()))
 
     async def unload(self, future=None):
         """Stop the event loop."""
@@ -303,7 +304,7 @@ class OpsDroid:
         if connectors:
             for connector in self.connectors:
                 if self.eventloop.is_running():
-                    self.eventloop.create_task(connector.connect())
+                    asyncio.wait(self.eventloop.create_task(connector.connect()))
                 else:
                     self.eventloop.run_until_complete(connector.connect())
             for connector in self.connectors:
@@ -346,7 +347,7 @@ class OpsDroid:
                     _LOGGER.debug(_("Adding database: %s"), name)
                     database = cls(database_module["config"])
                     self.memory.databases.append(database)
-                    self.eventloop.run_until_complete(database.connect())
+                    self.eventloop.run_until_complete(database.connect(self))
 
     async def run_skill(self, skill, config, message):
         """Execute a skill."""
@@ -360,15 +361,14 @@ class OpsDroid:
             else:
                 await skill(message)
         except Exception:
+            _LOGGER.exception(
+                _("Exception when running skill '%s' "), str(config["name"])
+            )
             if message:
                 await message.respond(
                     events.Message(_("Whoops there has been an error"))
                 )
                 await message.respond(events.Message(_("Check the log for details")))
-
-            _LOGGER.exception(
-                _("Exception when running skill '%s' "), str(config["name"])
-            )
 
     async def get_ranked_skills(self, skills, message):
         """Take a message and return a ranked list of matching skills."""
--- a/opsdroid/events.py
+++ b/opsdroid/events.py
@@ -19,15 +19,16 @@ _LOGGER = logging.getLogger(__name__)
 
 class EventCreator:
     """Create opsdroid events from events detected by a connector."""
+
     def __init__(self, connector, dispatch_key="type"):
-        """Initialise the event creator"""
+        """Initialise the event creator."""
         self.connector = connector
         self.dispatch_key = dispatch_key
 
         self.event_types = defaultdict(lambda: self.skip)
 
     async def create_event(self, event, target):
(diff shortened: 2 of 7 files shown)


jacobtomlinson:
Thanks for raising this!

Interestingly I think GitHub isn't clever enough to know you've put a strikethrough the "closes" section and I expect this will still close it on merge. 
image

tests/test_connector.py
```diff
@@ -62,6 +62,17 @@ def send_my_event(self, event):
...
+    def test_event_subclasses(self):
+        class MyEvent(Message):
+            pass
+
+        class MyConnector(Connector):
+            @register_event(Message, include_subclasses=True)
+            def send_my_event(self, event):
+                pass
+
+        MyConnector({})
```
jacobtomlinson: Can we add some kind of assertion in here? Perhaps try sending an instance of `MyEvent` and assert that it doesn't raise.

opsdroid/events.py
```diff
@@ -269,12 +311,14 @@ def __init__(
...
         self._file_bytes = file_bytes
         self.url = url
+        self._url_headers = url_headers
 
     async def get_file_bytes(self):
         """Return the bytes representation of this file."""
         if not self._file_bytes and self.url:
             async with aiohttp.ClientSession() as session:
-                async with session.get(self.url) as resp:
+                _LOGGER.debug(self._url_headers)
```
jacobtomlinson: I assume we are debug logging the headers here for a reason. Perhaps it would be good to include the url too for some context?

opsdroid/events.py
```diff
@@ -168,6 +199,10 @@ def __init__(self, text, *args, **kwargs):
         self.text = text
         self.raw_match = None
 
+    def __repr__(self):
+        """Override Message's representation so you can see the text when you print it."""
+        return f"<opsdroid.events.Message(text={self.text})>"
```
jacobtomlinson: I think this is the first f-string that has made it in since we dropped 3.5 🎉.

opsdroid/core.py
```diff
@@ -114,15 +114,14 @@ def critical(self, error, code):
     @staticmethod
     def handle_async_exception(loop, context):
         """Handle exceptions from async coroutines."""
+        print("ERROR: Unhandled exception in opsdroid, exiting...")
```
Cadair: Note to reviewer: using logging here doesn't actually work, I found a cpython bug about it, but now it's lost to the sands of time.

jacobtomlinson: Why are we printing instead of logging?

opsdroid/core.py
```diff
@@ -114,15 +114,14 @@ def critical(self, error, code):
...
         if "future" in context:
             try:  # pragma: nocover
                 context["future"].result()
             # pylint: disable=broad-except
             except Exception:  # pragma: nocover
-                _LOGGER.exception(_("Caught exception"))
-        else:
-            _LOGGER.error(_("Caught exception"))
-        _LOGGER.error(context)
+                print("Caught exception")
```
jacobtomlinson: Same as above

tests/test_connector.py
Cadair: Without trying it, I think you can test the behaviour here with this:
```suggestion
        c = MyConnector({})
        assert MyEvent in MyConnector.events
```

opsdroid/core.py
```diff
@@ -303,7 +304,7 @@ def start_connectors(self, connectors):
         if connectors:
             for connector in self.connectors:
                 if self.eventloop.is_running():
-                    self.eventloop.create_task(connector.connect())
+                    asyncio.wait(self.eventloop.create_task(connector.connect()))
```
jacobtomlinson: AFAIK `asyncio.wait` is a coroutine and will return a future which never gets awaited here. What was the reasoning behind changing this?

https://docs.python.org/3/library/asyncio-task.html#asyncio.wait

opsdroid/core.py
jacobtomlinson: Ah fun! You beat me to it!

Cadair: See my comment here: https://github.com/opsdroid/opsdroid/pull/1069#discussion_r327158672

The logger is in the process of getting cleaned up when the python process is exiting, so this was triggering that really long traceback from inside this handler.

tests/test_connector.py
jacobtomlinson: Even neater!

opsdroid/core.py
Cadair: It returns an awaitable, but isn't a coroutine I don't think. The docstring says:

> Run awaitable objects in the aws set concurrently and block until the condition specified by return_when.

So by default this will block until the task finishes, which is what we desire here, so we can correctly emit the startup event knowing the connectors have all connected.

SolarDrew: I don't recall, but the tests pass :man_shrugging: I'll test it without as well and see if I can reverse-engineer why we did this. In the meantime @Cadair might remember better than I do.

jacobtomlinson: image

opsdroid/core.py
SolarDrew: Probably just a mistake, I'll change it.

opsdroid/core.py
SolarDrew: Same as above.

opsdroid/events.py
SolarDrew: All hail f-strings!

opsdroid/core.py
jacobtomlinson: image

tests/test_connector.py
SolarDrew: Addressed by @Cadair's suggested change, I think.

opsdroid/core.py
jacobtomlinson: I think you do need to await it to get it to block.

jacobtomlinson: @SolarDrew I expect everything works but there may be a warning about not awaiting it. The previous implementation also did not block at this point, it just submitted the task to the loop and moved on. 

I wondered if the idea here was to try and avoid some race condition.

opsdroid/core.py
Cadair: I found it, it was this: https://bugs.python.org/issue26789

opsdroid/core.py
Cadair: Note this is being called from inside a function which isn't `async def` so we can't await it.

The objective was to ensure that all the connectors have started before returning to the calling scope where we emit the `OpsDroidStarted` event.

Cadair: There must be a way to block in non-async code and wait for a `Task` or a `Future` to finish though?! If this isn't it, what is?

Cadair: https://docs.python.org/3/library/asyncio-task.html#asyncio.as_completed ?!

Cadair: I will have to see if I can setup a test that fails if this doesn't block...

jacobtomlinson: Check out #1072, might help.

jacobtomlinson: Ok #1072 is now merged, you'll need to deal with the conflict but you should be unblocked here now.

jacobtomlinson:
This closes #999.

opsdroid/events.py
Cadair: I am not sure we are including this for any real reason here...

SolarDrew:
Do you have a preference for where you want that documented? I think at some point it's worth having a section in the docs that briefly describes each event and how to use it, but I don't think we have that anywhere yet, bar what's in the docstrings.

jacobtomlinson:
This is great many thanks! This looks good to go 🚀. 

As @SolarDrew raised this PR I'll leave @Cadair to do the merging once CI has finished.

Cadair:
:tada:  Nice!

PR #1070, opened 2019-09-23
Events IIb - Matrix

SolarDrew:
# Description

Follows from PRs #951 and #1069. Adds Matrix-specific event types and appropriate handling by the connector.

Closes https://github.com/SolarDrew/skill-picard/issues/26.

Requires #1069.

## Status
**UNDER DEVELOPMENT**

## Type of change

# How Has This Been Tested?

It hasn't yet.

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -288,6 +288,11 @@ class ConnectorMatrix(Connector):
 
         if isinstance(message.linked_event, str):
             edited_event_id = message.linked_event
+        elif isinstance(message.linked_event, events.EditedMessage):
+            # If we are attempting to edit an edit, move up the tree and then
+            # try again.
+            message.linked_event = message.linked_event.linked_event
+            return self._send_edit(message)
         else:
             edited_event_id = message.linked_event.event_id
 
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -173,7 +173,11 @@ class ConnectorMatrix(Connector):
 
             await self.opsdroid.parse(
                 events.UserInvite(
-                    target=roomid, user=sender, connector=self, raw_event=invite_event
+                    target=roomid,
+                    user_id=invite_event["sender"],
+                    user=sender,
+                    connector=self,
+                    raw_event=invite_event,
                 )
             )
 
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -56,13 +56,16 @@ class ConnectorMatrix(Connector):
 
         self._event_creator = MatrixEventCreator(self)
 
-    def message_type(room):
+    def message_type(self, room):
         if self.send_m_notice:
-            return True
+            return "m.notice"
         reverse_room_ids = {v: k for k, v in self.room_ids.items()}
         room = reverse_room_ids.get(room, room)
         if room in self.rooms:
-            return self.rooms[room].get("send_m_notice", False)
+            if self.rooms[room].get("send_m_notice", False):
+                return "m.notice"
+
+        return "m.text"
 
 
     def _process_rooms_dict(self, rooms):
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -1,6 +1,7 @@
 """Connector for Matrix (https://matrix.org)."""
 
 import re
+import json
 import logging
 from concurrent.futures import CancelledError
 from urllib.parse import urlparse
@@ -12,10 +13,11 @@ from matrix_client.errors import MatrixRequestError
 from voluptuous import Required
 
 from opsdroid.connector import Connector, register_event
-from opsdroid.events import Message, Image, File
+from opsdroid import events
 
 from .html_cleaner import clean
 from .create_events import MatrixEventCreator
+from . import events as matrixevents
 
 
 _LOGGER = logging.getLogger(__name__)
@@ -38,7 +40,7 @@ class ConnectorMatrix(Connector):
         """Init the config for the connector."""
         super().__init__(config, opsdroid=opsdroid)
 
-        self.name = "ConnectorMatrix"  # The name of your connector
+        self.name = "matrix"  # The name of your connector
         self.rooms = config["rooms"]
         self.room_ids = {}
         self.default_target = self.rooms["main"]
@@ -61,22 +63,16 @@ class ConnectorMatrix(Connector):
             "account_data": {"limit": 0, "types": []},
             "presence": {"limit": 0, "types": []},
             "room": {
-                "rooms": [],
                 "account_data": {"types": []},
-                "timeline": {"types": ["m.room.message"]},
                 "ephemeral": {"types": []},
                 "state": {"types": []},
             },
         }
 
-    async def make_filter(self, api, room_ids):
+    async def make_filter(self, api):
         """Make a filter on the server for future syncs."""
         fjson = self.filter_json
-        for room_id in room_ids:
-            fjson["room"]["rooms"].append(room_id)
-
         resp = await api.create_filter(user_id=self.mxid, filter_params=fjson)
-
         return resp["filter_id"]
 
     async def connect(self):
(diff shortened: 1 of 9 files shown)


Cadair:
This has broken some of the matrix connector tests that already exist, and needs tests for the new behaviour.

opsdroid/connector/matrix/connector.py
```diff
@@ -266,3 +286,81 @@ def get_roomname(self, room):
...
+
+    @register_event(events.RoomAddress)
+    async def _send_room_address(self, address_event):
+        try:
+            return await self.connection.set_room_alias(address_event.target,
+                                                        address_event.address)
+        except MatrixRequestError as err:
+            if err.code == 409:
+                pass
+            # err["error"] = "Room alias #picard_general:localhost already exists."
```
```suggestion
```

opsdroid/connector/matrix/connector.py
```diff
@@ -266,3 +286,81 @@ def get_roomname(self, room):
...
+    @register_event(events.UserInvite)
+    async def _send_user_invitation(self, invite_event):
+        try:
+            return await self.connection.invite_user(invite_event.target,
+                                                     invite_event.user)
+        except MatrixRequestError as err:
+            content = json.loads(err.content)
+            if (err.code == 403 and
+                "is already in the room" in content["error"]):
+                pass
```
```suggestion
                return
```

opsdroid/events.py
```diff
@@ -357,3 +357,60 @@ class Image(File):
...
+class NewRoom(Event):
+    """Event class to represent the creation of a new room."""
+
+    def __init__(self, name=None, params=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.name = name
+        self.room_params = params or {}
+
+
+class RoomName(Event):
```
Cadair: all these events need docstrings.

jacobtomlinson: These new events are interesting because it's not clear whether they are inbound, outbound or both. Does this event happen when the room name changes? Or do you send it in order to change the room name?

Should events for both of those things exist, or should it just be one event? Is this that one event?

Ad you say docstrings would help a lot here.

Cadair: All these events are designed to be bi-directional, they can be emitted or consumed by connectors. This fact has caused some confusion (@SolarDrew and I got caught out multiple times).

opsdroid/events.py
```diff
@@ -271,6 +270,29 @@ def __repr__(self):
         return await super().respond(response)
 
 
+class EditedMessage(Message):
```
Cadair: @jacobtomlinson This is at least part of #1230 and not just scoped to the matrix connector if you want to give some feedback on it.

opsdroid/core.py
```diff
@@ -369,7 +369,7 @@ def _connector_names(self):  # noqa: D401
                     and cls is not Database
                 ):
                     _LOGGER.debug(_("Adding database: %s."), name)
-                    database = cls(database_module["config"])
+                    database = cls(database_module["config"], opsdroid=self)
```
Cadair: @jacobtomlinson I assume this is correct, it seems a bunch of the databases expected this?

jacobtomlinson: Yeah probably.

opsdroid/events.py
jacobtomlinson: Thanks for adding this. I'm still unsure what to do in this situation. 

- When a message edit comes through should it still be put through the NLU pipline and responded to like a message? 
- What if the bot has already responded? 
- Should it edit its own response? 
- Or add another one?

SolarDrew: I think it should probably still go through the pipeline, because I'm not sure we can have any way to be sure the edit hasn't changed the meaning of the message without just checking it again (but then I'm not that familiar with NLU so maybe that happens less often than I think). And I think editing responses makes more sense if the message has been edited, otherwise you have two reponses per visible user message, which I think doesn't help clarity for anyone.

jacobtomlinson: Ok so we should be fine [here](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/core.py#L492) as it is a subclass of `opsdroid.events.Message` but we probably need to implement some way of editing the previous message in the connectors.

opsdroid/connector/matrix/connector.py
```diff
@@ -112,11 +154,31 @@ def filter_json(self):
...
-            if room and "timeline" in room:
+
+        # Emit Invite events for every room in the invite list.
+        for roomid, room in response["rooms"]["invite"].items():
+            # Process the invite list to extract the person who invited us.
+            invite_event = [
+                e
+                for e in room["invite_state"]["events"]
+                if "invite" == e.get("content", {}).get("membership")
+            ][0]
```
SolarDrew: This is quite the list comprehension. Is there not a way to do this more concisely?

opsdroid/connector/matrix/connector.py
```diff
@@ -31,28 +34,73 @@
 __all__ = ["ConnectorMatrix"]
 
 
+def ensure_room_id_and_send(func):
+    """
+    Ensure that the target for the event is a matrix room id.
+
+    Also retry the function call if the server disconnects.
+    """
```
SolarDrew: Good call on this. Is there any situation we're not going to want to use this though? Would be good if there were some way to just do this by default on all the functions, without having to add an extra decorator to them.

opsdroid/events.py
Cadair: If you respond to an edit with an edit it should edit the parent message. I would need to verify this though.

opsdroid/connector/matrix/connector.py
Cadair: I guess we could override `Connector.send` to do it there, but I think this is probably easier to track.

SolarDrew: You mean for debugging purposes? Yeah I guess so. But think of all the lines of code you could delete :P

Cadair: We would need the decorator anyway for the conversion to room id, so I really don't think it would save us much.

opsdroid/events.py
Cadair: I think that will be a thing for the connectors to implement. (On some platforms maybe you want to edit the edit event, on some maybe you need to edit the original one again).

opsdroid/connector/matrix/connector.py
Cadair: I can't think of one, it's parsing quite a complex response structure in the /sync feed.

Cadair:
I gave the reactions and edits a bit of an update last night, to better make use of `linked_event`, it puts more burden on the connectors when receiving these events, but makes it nicer for users and also more symmetrical with the responding workflow.

opsdroid/events.py
```diff
@@ -271,6 +270,26 @@ def __repr__(self):
...
+    The ``linked_event`` property should hold either an `opsdroid.events.Event`
+    class or an id for an event to which the edit applies.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """Create object with minimum properties."""
+        super().__init__(*args, **kwargs)
+
+
+class Reply(Message):
```
jacobtomlinson: Should calling `Message.respond(string)` generate a `Reply` event instead of a `Message` event then?

Cadair: I would say no, because that would be a rather large change in behavior. I am seeing Reply here as also applying to slack threads, where you would reply to a message to put it in a thread.

jacobtomlinson: Ah that makes sense!

opsdroid/events.py
```diff
@@ -271,6 +270,26 @@ def __repr__(self):
...
+class EditedMessage(Message):
+    """A  `opsdroid.events.Message` which has been edited.
+
+    The ``linked_event`` property should hold either an `opsdroid.events.Event`
+    class or an id for an event to which the edit applies.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """Create object with minimum properties."""
+        super().__init__(*args, **kwargs)
```
SolarDrew: `linked_event` is already an attribute that you can set on the base `Message` class then?

opsdroid/connector/matrix/connector.py
```diff
@@ -195,33 +271,82 @@ def _get_formatted_message_body(message, body=None, msgtype="m.text"):
...
 
-        try:
+        new_content = self._get_formatted_message_body(
+            message.text, msgtype=self.message_type(message.target)
+        )
+
+        content = {
+            "msgtype": self.message_type(message.target),
+            "m.new_content": new_content,
+            "body": f"* {new_content['body']}",
```
SolarDrew: All hail f-strings!

jacobtomlinson:
🎉🚀

opsdroid/events.py
Cadair: it's on `Event`.

Cadair:
Wooop!

PR #1072, opened 2019-09-23
Make load and start_connectors async

jacobtomlinson:
# Description

Make `opsdroid.OpsDroid.load` and `opsdroid.OpsDroid.start_connectors` async. This makes it easier to block on connector startup from within `start_connectors`.

## Status
**READY**

### DIFF
--- a/opsdroid/__main__.py
+++ b/opsdroid/__main__.py
@@ -68,7 +68,7 @@ def print_example_config(ctx, param, value):
 def edit_files(ctx, param, value):
     """Open config/log file with favourite editor."""
     if value == "config":
-        file = DEFAULT_CONFIG_PATH
+        file = f'"{DEFAULT_CONFIG_PATH}"'
     elif value == "log":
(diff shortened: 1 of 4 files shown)


opsdroid/__main__.py
```diff
@@ -68,7 +68,7 @@ def print_example_config(ctx, param, value):
 def edit_files(ctx, param, value):
     """Open config/log file with favourite editor."""
     if value == "config":
-        file = DEFAULT_CONFIG_PATH
+        file = f'"{DEFAULT_CONFIG_PATH}"'
```
SolarDrew: There goes my "first f-strings merged" achievement :(

Cadair:
This looks like it would make things easier.

opsdroid/__main__.py
jacobtomlinson: This actually was shrapnel from something else that didn't work, need to remove before merging so your crown is still intact!

PR #1075, opened 2019-09-24
Refactor CLI commands

jacobtomlinson:
In order to make the opsdroid CLI API more extensible, I thought it might be good to refactor things a little here. I'm considering adding more things to the command line and using flags to do so is not so clear. E.g how do you have flags which are only valid when other flags are set, etc.

I've deprecated the existing flags in favour of a subcommand style syntax. This allows functionality to be partitioned off and then those subcommands can have flags. For example `opsdroid logs -f` could be to tail the logs, but the `-f` could be force in other scenarios.

- `opsdroid` becomes `opsdroid start`
- `opsdroid --gen-config` becomes `opsdroid config gen`
- `opsdroid --edit-config` becomes `opsdroid config edit`
- `opsdroid --view-log` becomes `opsdroid logs`
- `opsdroid --version` becomes `opsdroid version`

The current syntax still exists and works, but now with deprecation warnings. I've not added any tests or docs yet as I'd like feedback on this change.


opsdroid/__main__.py
```diff
@@ -160,14 +171,25 @@ def welcome_message(config):
...
-def main():
+def cli(ctx):
     """Opsdroid is a chat bot framework written in Python.
 
     It is designed to be extendable, scalable and simple.
     See https://opsdroid.github.io/ for more information.
     """
+    if ctx.invoked_subcommand is None:
+        warn_deprecated_cli_option(
+            "Running `opsdroid` without a subcommand is now deprecated. "
```
FabioRosado: Do you think it makes sense adding information about which commands you can run when the user types `opsdroid`? I was thinking that the deprecation message should be logged and then show all the other subcommands that can be run, that way the user could get more information on how to run other commands in case they need.

jacobtomlinson: What about something like this to let people know there is more. Rather than blurting the whole help text here?

```suggestion
            "Running `opsdroid` without a subcommand is now deprecated. "
            "Please run `opsdroid start` instead."
            "You can also run `opsdroid --help` to learn about the other subcommands"
```

FabioRosado: Yeah I think that looks good 👍

opsdroid/__main__.py
```diff
@@ -160,14 +171,27 @@ def welcome_message(config):
...
+            "Please run `opsdroid start` instead. "
+            "You can also run `opsdroid --help` to learn about the other subcommands."
+            "Please run `opsdroid start` instead."
```
jacobtomlinson:
```suggestion
```

jacobtomlinson:
Things have grown a little here. I've factored the CLI code out into its own submodule `opsdroid.cli`. This should make adding new commands easy and neat.

SolarDrew:
LGTM :+1:

opsdroid/cli/utils.py
```diff
@@ -0,0 +1,115 @@
...
+                _(
+                    "Install Opsdroid Desktop at: \n"
+                    "https://github.com/opsdroid/opsdroid-desktop/"
+                    "releases"
+                )
+            )
+            _LOGGER.info("=" * 40)
+    except KeyError:
+        _LOGGER.warning(
+            _("'welcome-message: true/false' is missing in " "configuration.yaml")
```
FabioRosado: I've noticed that this watning is split in two strings, I'm sure this was due to linting but now that we are using black, I think black moved two lines into one. Should we change this to a single string?

PR #1116, opened 2019-10-01
Adding user id Parmeter in Event class

### LINKED ISSUE #1082: All events should take both "user" and "user_id"
Currently the base [`Event` class](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/events.py#L103) takes a `user=` parameter to it's constructor. This is generally implemented as the human readable name of the user. For a number of events we also need the machine readable user id for the event. 

I propose that we add `user_id=` to the event base class and all subclasses, and in all places connectors create events. This will allow us to provide a standard way of using one or the other for the event types that need one or both.

himanshupathak21061998:
# Description

_Adding user_id keyword in Event class. This is generally implemented as the human readable name of the user. For a number of events, we also need the machine readable user id for the event._

Fixes #1082

## Status
 **Ready**

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/create_events.py
+++ b/opsdroid/connector/matrix/create_events.py
@@ -48,6 +48,7 @@ class MatrixEventCreator:
         """Send a Message event."""
         return events.Message(
             text=event["content"]["body"],
+            user_id=event["sender"],
             user=await self.connector.get_nick(roomid, event["sender"]),
             target=roomid,
             connector=self.connector,
@@ -60,6 +61,7 @@ class MatrixEventCreator:
         return dict(
             url=url,
             name=event["content"]["body"],
+            user_id=event["sender"],
             user=await self.connector.get_nick(roomid, event["sender"]),
             target=roomid,
             connector=self.connector,
--- a/opsdroid/connector/matrix/create_events.py
+++ b/opsdroid/connector/matrix/create_events.py
@@ -47,10 +47,10 @@ class MatrixEventCreator:
     async def create_message(self, event, roomid):
         """Send a Message event."""
         return events.Message(
-            event["content"]["body"],
-            await self.connector.get_nick(roomid, event["sender"]),
-            roomid,
-            self.connector,
+            text=event["content"]["body"],
+            user=await self.connector.get_nick(roomid, event["sender"]),
+            target=roomid,
+            connector=self.connector,
             event_id=event["event_id"],
             raw_event=event,
         )
--- a/opsdroid/connector/rocketchat/__init__.py
+++ b/opsdroid/connector/rocketchat/__init__.py
@@ -94,10 +94,10 @@ class RocketChat(Connector):
         """
         if response["messages"]:
             message = Message(
-                response["messages"][0]["msg"],
-                response["messages"][0]["u"]["username"],
-                response["messages"][0]["rid"],
-                self,
+                text=response["messages"][0]["msg"],
+                user=response["messages"][0]["u"]["username"],
+                target=response["messages"][0]["rid"],
+                connector=self,
             )
             _LOGGER.debug(
                 "Received message from Rocket.Chat %s", response["messages"][0]["msg"]
--- a/opsdroid/connector/rocketchat/__init__.py
+++ b/opsdroid/connector/rocketchat/__init__.py
@@ -97,6 +97,7 @@ class RocketChat(Connector):
         if response["messages"]:
             message = Message(
                 text=response["messages"][0]["msg"],
+                user_id=response["messages"][0]["u"]["_id"],
                 user=response["messages"][0]["u"]["username"],
                 target=response["messages"][0]["rid"],
                 connector=self,
--- a/opsdroid/connector/telegram/__init__.py
+++ b/opsdroid/connector/telegram/__init__.py
@@ -59,13 +59,16 @@ class ConnectorTelegram(Connector):
 
         """
         user = None
+        user_id = None
+
         if "username" in response["message"]["from"]:
             user = response["message"]["from"]["username"]
 
         elif "first_name" in response["message"]["from"]:
             user = response["message"]["from"]["first_name"]
+        user_id = response["message"]["from"]["id"]
 
-        return user
+        return user, user_id
 
     def handle_user_permission(self, response, user):
         """Handle user permissions.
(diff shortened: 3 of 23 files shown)


opsdroid/events.py
```diff
@@ -100,13 +101,15 @@ class Event(metaclass=EventMetaClass):
...
         self,
+        user_id=None,
         user=None,
         target=None,
         connector=None,
         raw_event=None,
         event_id=None,
         linked_event=None,
     ):  # noqa: D107
+        self.user_id = user_id
```
himanshupathak21061998: I have one question over here should I do it like this
```
        user=None,
        target=None,
        connector=None,
        raw_event=None,
        event_id=None,
        linked_event=None,
        user_id=None,
```
Or I need to do changes in all test

Cadair: This change needs to happen here in the base class as well as in all the tests and places where events are created.

The big thing you need to do is go through the connectors and make sure they specify this keyword argument when they construct events having relieved them. i.e. [here](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/matrix/create_events.py) for matrix and [here](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/gitter/__init__.py#L81) for gitter. There should likely be a change in every connector.

himanshupathak21061998:
@Cadair  Everything changed as suggested thanks for suggestion

Cadair:
@himanshupathak21061998 Sorry I haven't looked at this in a while, can you resolve the merge conflicts and I will review it and get it merged.

himanshupathak21061998:
@Cadair  all conflicts have been resolved :grinning:

opsdroid/connector/matrix/create_events.py
```diff
@@ -47,10 +47,10 @@ def __init__(self, connector):
...
         """Send a Message event."""
         return events.Message(
-            event["content"]["body"],
-            await self.connector.get_nick(roomid, event["sender"]),
-            roomid,
-            self.connector,
+            text=event["content"]["body"],
+            user=await self.connector.get_nick(roomid, event["sender"]),
+            target=roomid,
+            connector=self.connector,
```
Cadair: we are going to want a `user_id=event['sender']` here, and in all the other events in this file.

opsdroid/connector/rocketchat/__init__.py
```diff
@@ -96,10 +96,10 @@ def build_url(self, method):
         """
         if response["messages"]:
             message = Message(
-                response["messages"][0]["msg"],
-                response["messages"][0]["u"]["username"],
-                response["messages"][0]["rid"],
-                self,
+                text=response["messages"][0]["msg"],
+                user=response["messages"][0]["u"]["username"],
```
Cadair: this structure implies that rocketchat might have the concept of ids. Would need someone who knows though.

opsdroid/connector/telegram/__init__.py
```diff
@@ -164,7 +164,10 @@ def build_url(self, method):
             elif "message" in result and "text" in result["message"]:
                 user = self.get_user(result)
                 message = Message(
-                    result["message"]["text"], user, result["message"]["chat"], self
+                    text=result["message"]["text"],
+                    user=user,
```
Cadair: I am 90% sure telegram has user IDs, so we would want to add them to this as well.

opsdroid/message.py
```diff
@@ -22,9 +22,11 @@ class Message(NewMessage):
     # Skip registering this class with Event
     _no_register = True
 
-    def __init__(self, text, user, room, connector, raw_message=None):  # noqa: D401
+    def __init__(
```
Cadair: we don't want to change this file as it's only provided for backwards compatibility.

opsdroid/connector/rocketchat/__init__.py
FabioRosado: This is correct, you can check the docs for the [channel history](https://rocket.chat/docs/developer-guides/rest-api/channels/history/).

The user id is found on: `response["messages"][0]["u"]["_id"]`

opsdroid/connector/telegram/__init__.py
FabioRosado: You might have to update the `self.get_user` helper method but yeah telegram does return user ids as well.

This is what Telegram returns:
```python
{'ok': True, 'result': [{'update_id': 664568166, 'message': {'message_id': 1198, 'from': {'id': 639889348, 'is_bot': False, 'first_name': 'Fabio', 'last_name': 'Rosado', 'username': 'FabioRosado', 'language_code': 'en'}, 'chat': {'id': 639889348, 'first_name': 'Fabio', 'last_name': 'Rosado', 'username': 'FabioRosado', 'type': 'private'}, 'date': 1571692254, 'text': 'hi'}}]}
```

So you can find the user id at `response["message"]["from"]["id"]`

himanshupathak21061998: I changes `self.get_user`  now it returns `user` as well as `user_id`

himanshupathak21061998:
Thanks @FabioRosado for reviewing. It's my pleasure to work on this. :smile:

Cadair:
This looks really good. I just have one question (which doesn't need to be addressed to merge this).

Should we make it so that all the connectors put `user_id=` in all their events? i.e. even if it's a duplication of the `user` field should we make it so that it's populated. This would mean that I could always access it irrespective of the connector.

One way we could do this which would be minimally disruptive, is to have the default in the `Event` class be `self.user_id = user_id or user` so that it defaults to `user`?

jacobtomlinson:
@himanshupathak21061998 are you planning on picking this up again?

himanshupathak21061998:
@jacobtomlinson yes I will.

himanshupathak21061998:
Please suggest if you want any changes in this.

opsdroid/events.py
```diff
@@ -133,6 +136,7 @@ def __init__(
 
         # Inherit the user, target and event from the event we are responding
         # to if they are not explicitly provided by this Event
+        event.user_id = event.user_id or self.user_id
```
Cadair: This suggestion, (with moving  the event.user line) will mean that we always populate `Event.user_id`. I think this is a good plan as it means it can always be relied on and can be overridden where appropriate.

```suggestion
        event.user = event.user or self.user
        event.user_id = event.user_id or self.user_id or event.user
```

opsdroid/events.py
```diff
@@ -134,6 +137,8 @@ def __init__(
         # Inherit the user, target and event from the event we are responding
         # to if they are not explicitly provided by this Event
         event.user = event.user or self.user
+        event.user_id = event.user_id or self.user_id or event.user
+        event.user = event.user or self.user
```
Cadair: You have this line twice now.
```suggestion
```

himanshupathak21061998: Sorry, removed it. Thanks for suggestion

Cadair:
Thanks @himanshupathak21061998 :tada:

PR #1122, opened 2019-10-02
added better quick link for documentation resource

Swati4star:
# Description

A.dded better quick link for documentation resource
I had to go to each documentation folder to check the resource available inside. I have created a better index to directly go to required resource

## Status
**READY** 

## Type of change

# How Has This Been Tested?

Just the documentation chaneg

# Checklist:

### DIFF
--- a/docs/Readme.md
+++ b/docs/Readme.md
@@ -1,0 +1,1 @@
+### Documentation
--- a/docs/Readme.md
+++ b/docs/Readme.md
@@ -1,1 +1,46 @@
-### Documentation
+# Documentation
+
+## Connectors
+- [Facebook](./connectors/facebook.md) 
+- [Github](./connectors/github.md)
+- [Gitter](./connectors/gitter.md)
+- [Matrix](./connectors/matrix.md)
+- [Slack](./connectors/slack.md)
+- [Telegram](./connectors/telegram.md)
+- [Webexteams](./connectors/webexteams.md)
+- [Websocket](./connectors/websocket.md)
+
+## Databases
+- [Mongo](./databases/mongo.md) 
+- [Redis](./databases/redis.md) 
+- [Sqlite](./databases/sqlite.md) 
+
+## Extending
+- [connectors](./extending/connectors.md) 
+- [databases](./extending/databases.md) 
+- [events](./extending/events.md) 
+- [intro](./extending/intro.md) 
+- [packaging](./extending/packaging.md) 
+- [skills](./extending/skills.md) 
+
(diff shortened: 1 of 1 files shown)


FabioRosado:
Hello @Swati4star  Thank you for raising this issue, you have mentioned something interesting.

@jacobtomlinson do you think we should replace the `docs/index.md` with a quick link menu like this instead of having the project `README.md` duplicated on index? 🤔

Swati4star:
@FabioRosado Sure. but I'll suggest to remove index.md file.
The benefit of using "Readme.md" is, it is automatically rendered when link is opened
What I mean is, check this link https://github.com/Swati4star/opsdroid/tree/doc-fixes/docs. Readme is already shown at bottom.

jacobtomlinson:
I'm not sure that this benefits the project. You have replaced a page filled with information for a list of links which are already available in the menu on the left hand side.

image

jacobtomlinson:
Oh wait I see you haven intended this just for use within the GitHub repo! Sorry about that.

In that case I agree that we might want a `README.md` file within that directory to help people find things. But my preference would probably be to simply link people to the documentation website.

Swati4star:
@jacobtomlinson  Are you suggesting some change in my PR here?

Swati4star:
If you suggest, I can add both to the Readme.

jacobtomlinson:
My concern is about keeping that list up to date.

jacobtomlinson:
Could I get an opinion from another @opsdroid/maintainers bod? I'm worried this list is another thing to maintain and I'm not sure how much value it adds.

Looking at the insights/analytics over the last 14 days we've had 37 people hit this page and 99 hit the docs site. My preference would be to divert the GitHub traffic to readthedocs, rather than providing them links deeper into the GitHub docs.

SolarDrew:
Hi @Swati4star, thanks for the PR. I'm inclined to agree with @jacobtomlinson, I think linking more directly to the docs is probably going to be more intuitive than asking people to look in the docs folder on GitHub. But I can also see the utility in having a list of quick links. 

Perhaps a reasonable middle-ground would be to add the links to the main project README? If we had a landing page for each of the headings (Connectors, Databases, etc.) then we could just link to those, which would reduce the maintainance burden and would mean the links wouldn't take up much space in the README, but it would make it easier for users to go straight to whichever part of the docs they need from the main GitHub page.

Swati4star:
So, @SolarDrew You want to have all these links (and sublinks) in the main readme? or just the doc headings.

PR #1124, opened 2019-10-02
Additional page fixes.

butuzov:
1. readme.md - full path to the graph/contributors, as file also used in website generation.
2. added tutorials/introduction.md instead tutorials as website doesn't have index pages
3. remove starting slash from events page, its better without it.
4. url for skills page changed, removed doc/ and added .md extention. lets generator deside how its going to be linked.

Status - Ready. Documentation changes.


README.md
```diff
@@ -206,8 +206,8 @@ _\* databases are optional, however bot memory will not persist between differen
 
 ### Contributors
 
-This project exists thanks to all the people who contribute. [[Contribute](/docs/contributing.md)].
-<a href="graphs/contributors"><img src="https://opencollective.com/opsdroid/contributors.svg?width=890" /></a>
+This project exists thanks to all the people who contribute. [[Contribute](contributing.md)].
```
jacobtomlinson: This one is tricky because it will now be broken from the GitHub README.

Maybe this needs to become an absolute URL?

butuzov: Interesting. Which one is priority? github or website? (i mean this is what we can get if we use the index as a link to the readme). On a website, you getting 404 too.

jacobtomlinson: Both should work. Perhaps adding a symlink from `contributing.md` to `docs/contributing.md` would be a suitable fix?

butuzov: Honestly? It looks like a bad design. You will get one more 404 (this time in GitHub, for tutorials/introduction.md that not exists).

jacobtomlinson: > Honestly? It looks like a bad design. 
Why?

butuzov: It introduces one more link that gives 404. Give me a few minutes, need to think about the issue.

jacobtomlinson: Why would it? Just leave the file in docs/contributing.md and update the link as you suggested so the website works fine. Then add a symlink in the repo from the root so that the repo README works.

butuzov: Well, lets try it. ping me, I will check site online once it will be regenerated.

butuzov: download

You know what? You have solve this issue already. Checkout the **To get started see the...*** at readme.md or index page of the https://docs.opsdroid.dev/en/stable/ or of the https://opsdroid.readthedocs.io/en/latest/

jacobtomlinson: Sure let's to that!

butuzov:
Now I am not so sure, making one more symlink will work.  Once I added symlink, `mkdocs` failing to generate docs. Guess I will just add link to https://opsdroid.readthedocs.io/en/latest/

butuzov:
I do believe we are (almost) ready to go. 

image

No errors on readthedocs page, and no on the local versions (root and path). The only thing that trouble me mkdocs message (see - bottom right window)

FabioRosado:
This is great thanks for updating and fixing these issues.

On mkdoc.yml we are using the pages, could you update that to nav? It should fix that deprecation warning 😄

butuzov:
and... done.

jacobtomlinson:
Awesome work!

butuzov:
thank you!

PR #1129, opened 2019-10-05
Make remaining logging messages translatable

### LINKED ISSUE #1128: Make all logging translatable
We have started using gettext to translate our logging messages, although I've come across a lot of messages that won't be translated.

In order to make a message translatable it has to be wrapped with `_(<string>)` so we should update all the logging messages to use this convention so we can translate all the messages available.

As an example on opsdroid.logging.ParsingFilter we are logging the following warning:

```python
                _LOGGER.warning(
                    "Both whitelist and blacklist "
                    "filters found in configuration. "
                    "Only one can be used at a time - "
                    "only the whitelist filter will be used."
                )

```

this logging needs to e changed like this:

```python

_LOGGER.warning(
    _(
        "Both whitelist and blacklist "
         "filters found in configuration. "
         "Only one can be used at a time - "
         "only the whitelist filter will be used."
     )
)
```

Pralkarz:
## Description

I searched for `_LOGGER` in VS Code and it yielded 322 results, out of which about 100 were relevant and not yet translatable. I wrapped all of them in `_(<string>)` and tested after that. Some of the logging calls with more than two arguments caused some tests to fail after that, so I reverted the changes, ultimately changing about ~50 calls. I also ran `black` manually as per @jos3p's [suggestion](https://github.com/opsdroid/opsdroid/issues/1128#issuecomment-538653664) since I had some issues with `pre-commit`.

Fixes #1128

## Status
**READY**

## Type of change

# How Has This Been Tested?

I followed the [contributing guide]https://opsdroid.readthedocs.io/en/latest/contributing/l), i.e.:

`pip install -U pre-commit`
`pre-commit install`
`pip install -U tox`
`tox`

As explained in the description, some of the tests failed after changing ALL the logging messages to be wrapped inside `_()`, so I reverted the changes on those to achieve the same testing results as before the changes.

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -132,7 +132,7 @@ class ConnectorMatrix(Connector):
                 # We can safely ignore timeout errors. The non-standard error
                 # codes are returned by Cloudflare.
                 if mre.code in [504, 522, 524]:
-                    _LOGGER.info(_("Matrix Sync Timeout (code: %d)", mre.code))
+                    _LOGGER.info(_("Matrix Sync Timeout (code: %d)"), mre.code)
                     continue
 
                 _LOGGER.exception(_("Matrix Sync Error"))
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -135,11 +135,11 @@ class ConnectorMatrix(Connector):
                     _LOGGER.info(_("Matrix Sync Timeout (code: %d)", mre.code))
                     continue
 
-                _LOGGER.exception("Matrix Sync Error")
+                _LOGGER.exception(_("Matrix Sync Error"))
             except CancelledError:
                 raise
             except Exception:  # pylint: disable=W0703
-                _LOGGER.exception("Matrix Sync Error")
+                _LOGGER.exception(_("Matrix Sync Error"))
 
     async def get_nick(self, roomid, mxid):
         """
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -123,7 +123,7 @@ class ConnectorMatrix(Connector):
                     timeout_ms=int(60 * 1e3),  # 1m in ms
                     filter=self.filter_id,
                 )
-                _LOGGER.debug("matrix sync request returned")
+                _LOGGER.debug(_("matrix sync request returned"))
                 message = await self._parse_sync_response(response)
                 if message:
                     await self.opsdroid.parse(message)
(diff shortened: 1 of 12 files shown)


Pralkarz:
Of course, I'll get to it right away!

Pralkarz:
It should be all done right now. Apologies if I skipped any, although I'm certain that at least 99% of them are translatable now. I also intentionally skipped calls like `_LOGGER.debug(fbmsg)` since I assume that wouldn't change anything.

opsdroid/connector/matrix/connector.py
```diff
@@ -132,14 +132,14 @@ def filter_json(self):
                 # We can safely ignore timeout errors. The non-standard error
                 # codes are returned by Cloudflare.
                 if mre.code in [504, 522, 524]:
-                    _LOGGER.info("Matrix Sync Timeout (code: %d)", mre.code)
+                    _LOGGER.info(_("Matrix Sync Timeout (code: %d)", mre.code))
```
FabioRosado: Here the  `mre.code` needs to be outside the _().

opsdroid/connector/websocket/__init__.py
```diff
@@ -124,8 +124,8 @@ def __init__(self, config, opsdroid=None):
             if message.target is None:
                 message.target = next(iter(self.active_connections))
             _LOGGER.debug(
-                "Responding with: '" + message.text + "' in target " + message.target
+                _("Responding with: '" + message.text + "' in target " + message.target)
```
FabioRosado: Would you mean replacing this strip with the other convention? Instead of using <string> + var + <string> we should keep it like the rest and use the %s convention.

If not I can change it after I merge the PR 😄

Pralkarz:
All done! When you do merge this PR, please let me know if there are any other issues that would be fitting for a beginner/intermediate to work on!

Pralkarz:
@FabioRosado I know this PR is closed, but I have one last question. As the commits got squished, how should I keep my fork up-to-date? I needed to re-fork, because it said that my fork is 5 commits ahead of the remote.

FabioRosado:
The way I do it is to open a new branch when I want to work on an issue and then delete the branch when its done. This will still make your fork outdated, you will need to add the opsdroid repository to your remote with the command:

`git remote add upstream https://github.com/opsdroid/opsdroid.git`

then you can do `git pull upstream master` and finally `git push` this will keep your opsdroid fork up to date.

Note that you can call the main opsdroid remote whatever you like, I just tend to call origin(my repo) and upstream (the main repo).

I hope this makes sense, please let me know if you need any help with anything else 😄

Also I was thinking that since you are from Poland you could work on translating opsdroid into Polish maybe? we don't have the issue open but it will count for your hacktoberfest anyway 😄 👍

Pralkarz:
@FabioRosado Thanks, that makes sense!

And yes, I could certainly work on translating opsdroid into Polish. I can work both on this and on the docstrings to mix things up a little bit if that's okay. :P

FabioRosado:
Haha of course that would be great 😄 
Again let me know if you need any help 👍

Issue #933, opened 2019-05-13
Overview of event flow


jacobtomlinson:
Thanks to the awesome work in #826 and #851 we now have an events system. This has sparked me to think more about the overall design of opsdroid and how events should flow through it. I wanted to raise this issue to lay out my thoughts and gather feedback.

Currently events are created by connectors and passed to `opsdroid.parse()`. Opsdroid then runs the event through some parsers and then calls one or more skills depending on the results. Skill can then do whatever people want with the event including creating response events and sending them back to the connector.

Events also end up getting modified as part of the parsing stage and can end up with parser specific attributes like `Event.regex`. If skill developers want to get more information about what the parser inferred they must interrogate these in a number of different ways. I raised #446 with the aim to resolving this, but this discussion should help frame that issue within a bigger plan.

## Event flow

I would like to be more explicit about the flow of events through the opsdroid core and specifically clean up what happens in `opsdroid.parse()`. Some of this is true already, some of this is a big of a change.

![image](https://user-images.githubusercontent.com/1610850/57621414-488ee280-7583-11e9-9eaa-f739300331f1.png)

1. Connectors should construct events and pass them to `opsdroid.parse()`.
1. Parsers should use add information to the event either by running some local processing or by sending the event off to a third party API. More on this is the **Parsers** section.
1. Matchers should determine which skills should be called for the event.
1. Skills should do whatever the developer wants, but will likely return an event back to the connector to be sent on to the chat client.

## Parsers

Parsers should use different techniques to add more information to an event. The most common thing would be to take a `Message` event and determine the intent and extract any entities from the sentence. A good example of this is any of the third-party intent matching APIs such as [wit.ai](https://opsdroid.readthedocs.io/en/stable/matchers/wit.ai/). 

```json
// Example wit.ai response
{
   "msg_id": "0zrCQ5LEkWd0MoHYM",
   "_text": "What's the weather like in London?",
   "entities": {
      "location": [
        {
          "suggested": true,
          "confidence": 0.74044071131585,
          "value": "London",
          "type": "value"
        }
      ],
      "intent": [
        {
          "confidence": 0.99979499373014,
          "value": "get_weather"
        }
      ]
   }
}
```

We can see that in this response wit.ai has concluded that for the sentence `What's the weather like in London?` the intent is `get_weather` and that there is one entity `location` with the value of `London`. In order to access this information the skill must explore this dictionary which can be found at `Event.witai`. If a different parser is used then the attribute will be different.

[...]

PR #1156, opened 2019-10-11
Add IBM Watson parser 

FabioRosado:
# Description

Decided to work on something new and added IBM Watson assistant to opsdroid. Tested the parser and works alright. Watson allows you to have a conversation with the assistant and it follows a logic.

I'm not sure if Watson uses the session id to keep track of the conversation. Initially I was using the session id but when I tried to talk to opsdroid it would crash because Watson api would give an error due to the session id (this is the reason why I'm getting a new session with every parse).

I still need to add the tests to this PR as well.

Fixes #<issue>

## Status

## Type of change

# How Has This Been Tested?

- Ran opsdroid with Watson and dummy hello skill - all working ok
- Ran opsdroid with Watson and default customer entity - all working ok
- Tox - all green

# Checklist:

### DIFF
--- a/docs/matchers/watson.md
+++ b/docs/matchers/watson.md
@@ -1,9 +1,11 @@
 # IBM Watson Matcher
 
+[IBM Watson](https://www.ibm.com/watson) is an NLP API for matching strings to [intents](https://cloud.ibm.com/docs/services/assistant?topic=assistant-intents). Intents are created on your assistant intent page.
+
 ## Configuring opsdroid
 
-In order to enable wit.ai skills, you must specify an `access-token` for your bot in the parsers section of the opsdroid configuration file.
-You can find this `access-token` in the settings of your App under the name: `'Server Access Token: '`.
+In order to enable Watson skills, you must specify an `access-token`, an `assistant-id` and a `gateway` for your bot in the parsers section of the opsdroid configuration file.
+You can find these informations inside your Assistant Settings under the `API Details`. Note that depending where your bot is located the gateway will be different, just use this. For example: `gateway-fra`.
 
 You can also set a `min-score` option to tell opsdroid to ignore any matches which score less than a given number between 0 and 1. The default for this is 0 which will match all messages.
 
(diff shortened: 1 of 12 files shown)


docs/matchers/watson.md
```diff
@@ -0,0 +1,161 @@
+# IBM Watson Matcher
+
+[IBM Watson](https://www.ibm.com/watson) is an NLP API for matching strings to [intents](https://cloud.ibm.com/docs/services/assistant?topic=assistant-intents). Intents are created on your assistant intent page.
+
+## Configuring opsdroid
+
+In order to enable Watson skills, you must specify an `access-token`, an `assistant-id` and a `gateway` for your bot in the parsers section of the opsdroid configuration file.
+You can find these informations inside your Assistant Settings under the `API Details`. Note that depending where your bot is located the gateway will be different, just use this. For example: `gateway-fra`.
```
jacobtomlinson: You've got `gateway-fra` here and `gateway-fr` further down. Are these both valid?

opsdroid/const.py
```diff
@@ -38,3 +38,6 @@
 WITAI_API_ENDPOINT = "https://api.wit.ai/message?"
 
 SAPCAI_API_ENDPOINT = "https://api.cai.tools.sap/v2/request"
+
+WATSON_API_ENDPOINT = ".watsonplatform.net/assistant/api"
```
jacobtomlinson: This looks strange. Is this part of a longer URL stored elsewhere?

opsdroid/parsers/watson.py
```diff
@@ -0,0 +1,95 @@
+"""A helper function for parsing and executing IBM watson skills."""
+import logging
+import json
+
+from ibm_watson import AssistantV2
+from ibm_cloud_sdk_core.authenticators import IAMAuthenticator
+from ibm_watson import ApiException
+
+from opsdroid.const import WATSON_API_ENDPOINT
+from opsdroid.const import WATSON_API_VERSION
```
jacobtomlinson: You have this import twice, you could combine them into 

```suggestion
from opsdroid.const import WATSON_API_ENDPOINT, WATSON_API_VERSION
```

opsdroid/parsers/watson.py
```diff
@@ -0,0 +1,95 @@
...
+
+_LOGGER = logging.getLogger(__name__)
+
+
+async def get_session_id(service, config):
+    """Authenticate and get session id from watson."""
+    service.set_service_url(
+        "https://{}{}".format(config["gateway"], WATSON_API_ENDPOINT)
```
jacobtomlinson: I would be tempted to change the `WATSON_API_ENDPOINT` in the const file to be `"https://{gateway}.watsonplatform.net/assistant/api"` and then change this line to be

```suggestion
        WATSON_API_ENDPOINT.format(gateway=config["gateway"])
```

docs/matchers/watson.md
```diff
@@ -0,0 +1,161 @@
...
+import logging
+
+_LOGGER = logging.getLogger(__name__)
+
+class MySkill(Skill):
+    @match_watson('Customer_Care_Appointments')
+    async def book_slot(self, message):
+        """Book an appointment"""
+        reply = message.watson['output']['generic'][0]['text']
+        booking_date = message.watson['output']['entities'][0]['value']
```
jacobtomlinson: Can we use the new entities syntax here in the example?

opsdroid/parsers/watson.py
```diff
@@ -0,0 +1,95 @@
...
+        for skill in skills:
+            for matcher in skill.matchers:
+
+                if "watson_intent" in matcher:
+                    _LOGGER.info(matcher)
+                    if (
+                        matcher["watson_intent"]
+                        in result["output"]["intents"][0]["intent"]
+                    ):
+                        message.watson = result
```
jacobtomlinson: As we have new development here can we try going with the proposal in #933?

```suggestion
                        message.raw_parses['watson'] = result
```

This may require some changes in the events classes to add the `raw_parses` attribute and updated to the docs you've written for this.

docs/matchers/watson.md
```diff
@@ -0,0 +1,161 @@
...
+from opsdroid.matchers import match_watson
+import logging
+
+_LOGGER = logging.getLogger(__name__)
+
+class MySkill(Skill):
+    @match_watson('Customer_Care_Appointments')
+    async def book_slot(self, message):
+        """Book an appointment"""
+        reply = message.watson['output']['generic'][0]['text']
```
jacobtomlinson: What is this `reply` object? I see you log it below but don't do anything else with it.

docs/matchers/watson.md
jacobtomlinson:
```suggestion
You can find this information inside your Assistant Settings under the `API Details`. Note that depending where your bot is located the gateway will be different, just use this. For example: `gateway-fra`.
```

docs/matchers/watson.md
FabioRosado: I was planning to do something with the reply but then didn't, will add another example that uses the reply it will make more sense I think

FabioRosado:
Thank you for the review! I've fixed some of the issues you have pointed, still need to work on the tests and stuff so I'll update the PR when I finally have some tests running 😄

docs/matchers/watson.md
jacobtomlinson: Ok 👍

docs/matchers/watson.md
FabioRosado: Do you mean to change it to  `message.raw_parses['watson']` once the `raw_parses` is implemented?

FabioRosado:
I've finally went around and finish the initial PR for the parser, I've changed everything that you suggested and wrote tests for all of the parser.

The only thing left to do is implement the `raw_parse`

docs/matchers/watson.md
jacobtomlinson: `message.watson['output']['entities'][0]['value']` could be changed to `message.entities['<name>']['value']`

FabioRosado:
I've been trying to set the dict as a property but everything I tried so far failed. This was mostly because the raw_parsers would be NoneType and wouldn't allow me to turn it into a dictionary.

After being stuck and playing around with it setting `raw_parsers=None or {}` seems to do the trick but then again this seems just a different way to do what I did before.

So I'm guessing I didn't fully understand what you suggested in the matrix chat and I'm not sure how to proceed here

opsdroid/events.py
```diff
@@ -116,10 +119,20 @@ def __init__(
...
+    @property
+    def raw_parses(self):
+        """Helper gets raw_parser."""
+        return self.raw_parser
+
+    @raw_parses.setter
+    def raw_parses(self, val):
+        """Helper sets raw_parser."""
+        self.raw_parser = val
+
```
jacobtomlinson: As you're not doing much here I wouldn't worry about these at all. We can just directly modify the `Event.raw_parser` attribute.

FabioRosado: On the last commit I’ve reverted these changes and moved the name back to raw_parses 😄

jacobtomlinson:
Awesome stuff thanks a lot for doing this!

PR #1164, opened 2019-10-15
refactor: Reduce Dockerfile steps

malexanderboyd:
# Reduced Dockerfile steps by removing redundant steps, and chaining commands

Removed some redundant lines in the Dockerfile. `WORKDIR` will create a directory if it doesn't exist, meaning the `RUN` line above is unneeded.

Also chained together some of the `RUN` commands together to reduce the total number of layers in the Dockerfile. Docker will create a new layer per command line, which increases build time. By chaining these commands together you can reduce build time + cache some dependencies so that don't have to rebuild the dependencies on a code change.

Happy Hacktoberfest 🎃 

## Status
**READY** 

## Type of change

# How Has This Been Tested?

- `docker build -t <name> .`
- `docker run -t <name>`

# Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,12 +5,12 @@ WORKDIR /usr/src/app
 
 # Copy source
 COPY opsdroid opsdroid
-COPY setup.cfg setup.cfg
-COPY README.md README.md
-COPY MANIFEST.in MANIFEST.in
-COPY requirements.txt requirements.txt
 COPY setup.py setup.py
 COPY versioneer.py versioneer.py
+COPY setup.cfg setup.cfg
+COPY requirements.txt requirements.txt
+COPY README.md README.md
+COPY MANIFEST.in MANIFEST.in
 
 RUN apk update \
 && apk add --no-cache gcc musl-dev git openssh-client \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,23 +1,22 @@
...
+COPY setup.py setup.py
+COPY versioneer.py versioneer.py
 
-RUN apk update && apk add --no-cache git openssh-client
-RUN pip3 install --upgrade pip
-RUN pip3 install --no-cache-dir --no-use-pep517 .
-RUN apk del gcc musl-dev
+RUN pip3 install --upgrade pip \
+&& pip3 install --no-cache-dir --no-use-pep517 . \
+&& apk del gcc musl-dev
```
jacobtomlinson: I would be tempted to move the `apk update` and `apk add` commands from line 4 to this block too. Some of these dependencies are only required to run the pip install, so it makes sense to remove them again in the same layer. This should really help with overall layer size.

malexanderboyd: Good idea, I moved the apk operations down into this run command. Another layer down!

PR #1166, opened 2019-10-15
Fixes schema.yaml and matrix connector to allow connections to selfho…

### LINKED ISSUE #1159: Error when starting with matrix-connector
When I try to start opsdroid I keep getting the following error. I double checked indentation and when I delete the matrix-connector part opsdroid starts fine. 
```
opsdroid_1     | Error validating data /root/.config/opsdroid/configuration.yaml with schema /usr/src/app/opsdroid/configuration/schema.yaml
opsdroid_1     | 	connectors.1.name: 'matrix' is not a regex match.
opsdroid_1     | 	connectors.1.name: 'matrix' is not a regex match.
opsdroid_1     | 	connectors.1.name: 'matrix' is not a regex match.
opsdroid_1     | 	connectors.1.webhook-url: Required field missing
opsdroid_1     | 	connectors.1.access-token: Required field missing
opsdroid_1     | 	connectors.1.name: 'matrix' is not a regex match.
opsdroid_1     | 	connectors.1.verify-token: Required field missing
opsdroid_1     | 	connectors.1.page-access-token: Required field missing
opsdroid_1     | 	connectors.1.name: 'matrix' is not a regex match.
opsdroid_1     | 	connectors.1.mxid: 'xxx' is not a regex match.
opsdroid_1     | 	connectors.1.room: Required field missing
opsdroid_1     | 	connectors.1.rooms.main: 'xxx' is not a regex match.
opsdroid_1     | 	connectors.1.rooms.other: 'xxx' is not a regex match.
opsdroid_1     | 	connectors.1.homeserver: 'xxx' is not a regex match.
opsdroid_1     | 	connectors.1.name: 'matrix' is not a regex match.
opsdroid_1     | 	connectors.1.api-token: Required field missing
opsdroid_1     | 	connectors.1.name: 'matrix' is not a regex match.
opsdroid_1     | 	connectors.1.token: Required field missing
opsdroid_1     | 	connectors.1.name: 'matrix' is not a regex match.
opsdroid_1     | 	connectors.1.consumer_key: Required field missing
opsdroid_1     | 	connectors.1.consumer_secret: Required field missing
opsdroid_1     | 	connectors.1.oauth_token: Required field missing
opsdroid_1     | 	connectors.1.oauth_token_secret: Required field missing
opsdroid_1     | 	connectors.1.name: 'matrix' is not a regex match.
opsdroid_1     | 	connectors.1.room-id: Required field missing
opsdroid_1     | 	connectors.1.access-token: Required field missing
```

## Configuration File

```yaml
##                      _           _     _
##   ___  _ __  ___  __| |_ __ ___ (_) __| |
##  / _ \| '_ \/ __|/ _` | '__/ _ \| |/ _` |
## | (_) | |_) \__ \ (_| | | | (_) | | (_| |
##  \___/| .__/|___/\__,_|_|  \___/|_|\__,_|
##       |_|
##                   __ _
##   ___ ___  _ __  / _(_) __ _
##  / __/ _ \| '_ \| |_| |/ _` |
## | (_| (_) | | | |  _| | (_| |
##  \___\___/|_| |_|_| |_|\__, |
##                        |___/
##
## A default config file to use with opsdroid
## see https://github.com/opsdroid/opsdroid/blob/master/opsdroid/configuration/example_configuration.yaml for more config options

## Set the logging level
logging:
  level: info
  path: opsdroid.log
  console: true

## Show welcome message
welcome-message: true

## Connector modules
connectors:

  - name: websocket
    # optional
    bot-name: "mybot" # default "opsdroid"
    max-connections: 10 # default is 10 users can be connected at once
    connection-timeout: 10 # default 10 seconds before requested socket times out

  ## Matrix (core)
  - name: matrix
    # Required
    mxid: "xxx"
    password: "xxx"
    # a dictionary of multiple rooms
    # One of these should be named 'main'
    rooms:
      'main': 'xxx'
      'other': 'xxx'
    # Optional
    homeserver: "xxx"
    nick: "xxx"  # The nick will be set on startup
    room_specific_nicks: False  # Look up room specific nicknames of senders (expensive in large rooms)

## Skill modules
skills:

  ## Dance (https://github.com/opsdroid/skill-dance)
  - name: dance

  ## Hello (https://github.com/opsdroid/skill-hello)
  - name: hello

  ## Loudnoises (https://github.com/opsdroid/skill-loudnoises)
  - name: loudnoises

  ## Seen (https://github.com/opsdroid/skill-seen)
  - name: seen
```

awesome-michael:
…sted matrix homeserver

# Description

Changes parsing of matrix connector config to allow self-hosted matrix homeservers

Fixes #1159

## Status
**READY**

## Type of change

- Bug fix

# How Has This Been Tested?

Tested with the following configuration that connected to my own matrix homeserver:

```
## Set the logging level
logging:
  level: info
  path: opsdroid.log
  console: true

## Show welcome message
welcome-message: true

## Connector modules
connectors:

  ## Matrix (core)
  - name: matrix
    # Required
    mxid: "@xxx:example.com"
    password: "xxx"
    # a dictionary of multiple rooms, at least one is needed
    rooms:
      - '#xxx:example.com'
      - '#xxx2:example.com'
    homeserver: "https://example.com"
    nick: "xxx"  # The nick will be set on startup
    room_specific_nicks: False  # Look up room specific nicknames of senders (expensive in large rooms)

## Skill modules
skills:
  ## Hello (https://github.com/opsdroid/skill-hello)
  - name: hello
```

# Checklist:

### DIFF
--- a/opsdroid/configuration/schema.yaml
+++ b/opsdroid/configuration/schema.yaml
@@ -68,18 +68,13 @@ github:
 ---
 matrix:
   name: regex('^(?i)matrix$')
-  mxid: regex('^\@[a-zA-Z]*:matrix.org$')
+  mxid: str()
   password: str()
-  room: str()
-  rooms: (include('room1'))
-  homeserver: regex('^https://matrix.org$')
+  rooms: list(str(), required=True)
+  homeserver: str()
   nick: str(required=False)
   room_specific_nicks: bool(required=False)
   send_m_notice: bool(required=False)
----
-room1:
-  'main': regex('^#matrix:matrix.org')
-  'other': regex('^#riot:matrix.org$')
 ---
 slack:
   name: regex('^(?i)slack$')
--- a/opsdroid/configuration/schema.yaml
+++ b/opsdroid/configuration/schema.yaml
@@ -70,7 +70,6 @@ matrix:
   name: regex('^(?i)matrix$')
   mxid: str()
   password: str()
-  room: str(required=False)
   rooms: (include('room1'))
   homeserver: str()
   nick: str(required=False)
--- a/opsdroid/configuration/schema.yaml
+++ b/opsdroid/configuration/schema.yaml
@@ -79,7 +79,6 @@ matrix:
 ---
 room1:
   'main': str()
-  'other': str()
 ---
 slack:
   name: regex('^(?i)slack$')
--- a/opsdroid/configuration/schema.yaml
+++ b/opsdroid/configuration/schema.yaml
@@ -70,11 +70,16 @@ matrix:
   name: regex('^(?i)matrix$')
   mxid: str()
   password: str()
-  rooms: list(str(), required=True)
+  room: str(required=False)
+  rooms: (include('room1'))
   homeserver: str()
   nick: str(required=False)
   room_specific_nicks: bool(required=False)
   send_m_notice: bool(required=False)
+---
+room1:
+  'main': str()
+  'other': str()
 ---
 slack:
   name: regex('^(?i)slack$')
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -31,8 +31,6 @@ class ConnectorMatrix(Connector):
 
         self.name = "ConnectorMatrix"  # The name of your connector
         self.rooms = config.get("rooms", None)
-        if not self.rooms:
-            self.rooms = {"main": config["room"]}
         self.room_ids = {}
         self.default_target = self.rooms["main"]
         self.mxid = config["mxid"]
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -30,7 +30,7 @@ class ConnectorMatrix(Connector):
         super().__init__(config, opsdroid=opsdroid)
 
         self.name = "ConnectorMatrix"  # The name of your connector
-        self.rooms = config.get("rooms", None)
+        self.rooms = config["rooms"]
         self.room_ids = {}
         self.default_target = self.rooms["main"]
         self.mxid = config["mxid"]
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -31,10 +31,8 @@ class ConnectorMatrix(Connector):
 
         self.name = "ConnectorMatrix"  # The name of your connector
         self.rooms = config.get("rooms", None)
-        if not self.rooms:
-            self.rooms = {"main": config["room"]}
         self.room_ids = {}
-        self.default_target = self.rooms["main"]
+        self.default_target = self.rooms[0]
         self.mxid = config["mxid"]
         self.nick = config.get("nick", None)
         self.homeserver = config.get("homeserver", "https://matrix.org")
@@ -84,9 +82,9 @@ class ConnectorMatrix(Connector):
         mapi.token = login_response["access_token"]
         mapi.sync_token = None
 
-        for roomname, room in self.rooms.items():
+        for room in self.rooms:
             response = await mapi.join_room(room)
-            self.room_ids[roomname] = response["room_id"]
+            self.room_ids[room] = response["room_id"]
         self.connection = mapi
 
         # Create a filter now, saves time on each later sync
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -31,8 +31,10 @@ class ConnectorMatrix(Connector):
 
         self.name = "ConnectorMatrix"  # The name of your connector
         self.rooms = config.get("rooms", None)
+        if not self.rooms:
+            self.rooms = {"main": config["room"]}
         self.room_ids = {}
-        self.default_target = self.rooms[0]
+        self.default_target = self.rooms["main"]
         self.mxid = config["mxid"]
         self.nick = config.get("nick", None)
         self.homeserver = config.get("homeserver", "https://matrix.org")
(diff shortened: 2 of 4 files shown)


Cadair:
What was the motivation for turning rooms into a list? I am using that to allow people to write skills which reference rooms for a purpose but with different rooms for each deployment. i.e. `main` and `backend` to allow commands in the backend rooms.

awesome-michael:
In the previous solution I was only capable of defining two rooms the bot should connect with. I wanted to give the possibility to specify as much rooms as needed. I thought about making a dict instead of a list but wanted to keep things simple.

Cadair:
There should be no limit to how long the dict of `name: room_alias` mappings should be, if there is it's only because the schema is too restrictive.

I am going to have to defer to @jacobtomlinson or @FabioRosado who know more about the yaml schemas here, the schema should just be a `str: str` mapping of arbitrary length.

jacobtomlinson:
cc @kollyuchka @goruha

FabioRosado:
I’m going to say that perhaps we should do what Cadair suggested and just check if rooms are of type dict - not sure if it’s worth to check if inside the dict the key/values are of type string (mostly because it seems obvious that a room should have a name in a string format - but i could be wrong)

Also I think the first room needs to be called `main` right? I’m just wondering as I looked at the code :thinking:

awesome-michael:
Hey,
I've been testing a bit and I was wrong. The rooms can be left as is and are not limited to the two ones specified in the schema.yaml (`main` and `other`).
Since this is my first PR I could need a little help on how to revert some of my changes.

jacobtomlinson:
@awesome-michael You have two options here. You can either go through yourself and change the values back, add a new commit and push it up. Or you can reset your branch back to an older commit [like this](https://stackoverflow.com/a/14873290) and force push that up.

Both approaches have pros and cons. My preference is to manually make the changes and add a new commit, that way you don't lose your changes if you decide in the future you were right the first time. We squash all your commits into a single one when we merge anyway.

opsdroid/configuration/schema.yaml
```diff
@@ -68,18 +68,18 @@ github:
...
+  homeserver: str()
   nick: str(required=False)
   room_specific_nicks: bool(required=False)
   send_m_notice: bool(required=False)
 ---
 room1:
-  'main': regex('^#matrix:matrix.org')
-  'other': regex('^#riot:matrix.org$')
+  'main': str()
+  'other': str()
```
Cadair: Does this forbid other keys? if no, we probably don't need this, and if yes it wont work properly anyway :grin:

awesome-michael: No it doesn't. I checked it. Maybe the 'other' could be removed. But I would leave the 'main' as this forces you to have at least one room that is named 'main' which is needed by the connector if I'm right.

FabioRosado: Yeah I think that's the best option, keep the main since it's needed and just remove the other since they can be called whatever you want I believe

Cadair: yeah, the main should stay, I was referring to just the other line :smile:

awesome-michael: I'm still a bit unhappy with the solution. We have a field `room` which I tagged as not required (which is not super correct, it's required when `rooms` is not used but it's optional otherwise) and the 'main' entry of `rooms` is a required field.
Would it be ok to delete the `room` field? Then rooms can have 1 to n lines but it needs a line with the name 'main'.

FabioRosado: Do you mean delete the room field from the validation? I'm assuming that if you mark main as required it will always fail even if room is not required so I'd say that it's acceptable to remove the rooms field from the validation schema.

Also, I think the matrix connector already throws an error(or logs it) if there is no main room available.

Cadair: I am happy to drop room and just go with rooms, assuming @SolarDrew has no objections.

opsdroid/connector/matrix/connector.py
```diff
@@ -31,8 +31,6 @@ def __init__(self, config, opsdroid=None):  # noqa: D107
 
         self.name = "ConnectorMatrix"  # The name of your connector
         self.rooms = config.get("rooms", None)
-        if not self.rooms:
```
FabioRosado: I was looking at the PR and are we sure we need to remove this from the code? 
since we need a room called main, perhaps we should leave this in place?  What do you think @Cadair ?

Cadair: I think we should just not make rooms optional. See suggestion on the line above.

opsdroid/connector/matrix/connector.py
```diff
@@ -31,8 +31,6 @@ def __init__(self, config, opsdroid=None):  # noqa: D107
 
         self.name = "ConnectorMatrix"  # The name of your connector
         self.rooms = config.get("rooms", None)
```
```suggestion
        self.rooms = config["rooms"]
```

opsdroid/connector/matrix/connector.py
awesome-michael: Since this is already checked by the schema the optional rooms are not needed. Thanks for the suggestion, I just added it.

opsdroid/configuration/schema.yaml
```diff
@@ -68,18 +68,17 @@ github:
 ---
 matrix:
   name: regex('^(?i)matrix$')
-  mxid: regex('^\@[a-zA-Z]*:matrix.org$')
+  mxid: str()
   password: str()
-  room: str()
+  room: str(required=False)
```
Cadair: We can drop this now correct?
```suggestion

```

awesome-michael: Sure.. Thought I already deleted it

Cadair:
Thanks @awesome-michael !! :tada:

FabioRosado:

Issue #1137, opened 2019-10-06
Inconsistency in the logging messages


ziebam:
# Description
A lot of the logging messages are inconsistent. Let's take the file `loader.py` for example. On the line 254 there's that message that doesn't end with a dot:
```
_LOGGER.debug(
    _(
        "Couldn't find the command 'pip', "
        "trying again with command 'pip3'"
    )
)
```
Yet just a few lines ahead, on line 267, there's a message that does end with a dot:
```
_LOGGER.debug(
    _(
        "Couldn't find the command 'pip3', "
        "install of %s will be skipped."
    ),
    str(requirements_path),
)
```

There are also some inconsistencies with the parsers and connectors naming. In some logging messages they are capitalized properly while in the others they are not like in telegram's `__init__.py` file on lines 24 and 41:
```
_LOGGER.debug(_("Loaded telegram connector"))
```
```
_LOGGER.error(
    _(
        "Unable to login: Access token is missing. "
        "Telegram connector will be unavailable."
    )
)
```

I don't know for certain, but this might require some tweaking of the tests so they recognize the capital or lowercase letters properly.

I will tackle this issue if there will be no volunteers. :D

FlipTheBit:
Attempted to fix the issue with the dots, can I try my hand at the capitalisation issue?

Abelarm:
I would like to work on this.
I had an idea to solve this problem in a nice way, but it requires pylint.
I would write a custom checker for pylint that will check all the logs and raise and warning based on a config file.

ammar0211:
I would like to work on this issue and look for all sorts of logging messages issues to make them all uniform

jacobtomlinson:
Thanks so much for raising this issue @ziebam and for picking it up folks.

Perhaps a good thing to do would be to add a section to the [contributing docs](https://github.com/opsdroid/opsdroid/blob/master/docs/contributing.md) with guidelines on how logs should be formatted.

FabioRosado:
Thats actually a good idea Jacob since we should explain better how gettext works and how we should pass the params to a logging that uses gettext 🤔

Abelarm:
Hi guys,
Are you looking more for:
- a test/checker for bad formatted logs
- a guideline for writing logs

jacobtomlinson:
> a guideline for writing logs

Let's start with this.

PR #1180, opened 2019-10-21
Update loader.py

Pranav-Jain:
# Description

Added full stops, where necessary to make the debugging statements consistent.

Fixes #<issue>
Fixes Issue #1137 

## Status
**READY**

## Type of change

### DIFF
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -48,7 +48,7 @@ class Loader:
         self.opsdroid = opsdroid
         self.modules_directory = None
         self.current_import_config = None
-        _LOGGER.debug(_("Loaded loader"))
+        _LOGGER.debug(_("Loaded loader."))
 
     @staticmethod
     def import_module_from_spec(module_spec):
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
@FabioRosado do we need to update the translations too when doing this?

FabioRosado:
Since we are just doing small fixes like adding punctuation and use uppercase letters on the beginning of the sentences, the translations could stay the same unless someone wants to fix them.

These fixes won't change the translations much 🤔

jacobtomlinson:
They wont but will the correct translations still be found because the string is changed within the `_(<string>)`?

jacobtomlinson:
Yeah I think it has to be a perfect match.

Pranav-Jain:
@FabioRosado By translations, do you mean the files at present at opsdroid/opsdroid/locale/ ?

FabioRosado:
Yes you will have to update the translations. I might be wrong but I think if you run the command `python setup.py update_catalog`  all the translations should be updated with these logs.

I'm going to merge #1181 soon so update your branch and then run the command I mentioned above and check if the messages got updated 😄 👍

PR #1184, opened 2019-10-22
Changes to _thinking_delay() and  _typing delay()

nishu88:
# Description
Made changes to _thinking_delay() and  _typing delay()
Fixes #532 

## Type of change


opsdroid/events.py
```diff
@@ -215,6 +215,12 @@ def __repr__(self):
         """Make opsdroid wait x-seconds before responding.
 
         Number of seconds defined in YAML config. file, accessed via connector.
+
+        Args:
```
FabioRosado: If there are no arguments being passed you can just remove this

opsdroid/events.py
```diff
@@ -228,6 +234,11 @@ def __repr__(self):
 
         Seconds to delay equals number of characters in response multiplied by
         number of seconds defined in YAML config. file, accessed via connector.
+
+        Args:
+            text (str): The text input to perform typing simulation on.
+        Returns:
+            N/A
```
FabioRosado: Same thing as above remove if doesnt return anything

jacobtomlinson:
Things are passing here now so let's get this in.

PR #1187, opened 2019-10-23
Fix configuration schema location

gergelypolonkai:
# Description

Instead of loading it from the module directory, it was loading it from the current directory.

## Status
**READY**

## Type of change

Bug fix (non-breaking change which fixes an issue)

# Checklist:


jacobtomlinson:
CI failed on linting. Please check out the [contributing docs](https://github.com/opsdroid/opsdroid/blob/master/docs/contributing.md#automatic-linting-with-black) in how to avoid this in future. For now I've run black and pushed another commit.

gergelypolonkai:
@jacobtomlinson thanks, although we already discussed that i switch from the `cls_module` hack(?) to using plain `__file__`.

On second thought, maybe querying the `opsdroid` module directly would be more future proof:

```
opsdroid_module = importlib.import('opsdroid')
schema_file = os.path.join(os.path.dirname(opsdroid_module.__file__), 'configuration/schema.yaml')
```

This way the calculated path isn’t relative to `loader.py` but to the module’s root.

What do you think?

jacobtomlinson:
I am not concerned too much about the path being relative to `loader.py`. I would prefer consistency (we use plain `__file__` ion other places) if that's ok.

opsdroid/const.py
```diff
@@ -25,6 +26,7 @@
     "configuration/example_configuration.yaml",
 )
 REGEX_PARSE_SCORE_FACTOR = 0.6
+MODULE_ROOT = os.path.dirname(os.path.abspath(opsdroid.__file__))
```
jacobtomlinson: You could shift this higher up (I suggest line 8) and then use this in the definition of `LOCALE_DIR`.

jacobtomlinson:

PR #1203, opened 2019-10-25
Added a documentation of a opsdroid skill

imrishabh18:
# Description

Added a documentation of a opsdroid skill.

Fixes #<1198>

## Status
UNDER DEVELOPMENT

## Type of change

- Documentation (fix or adds documentation

# How Has This Been Tested?

# Checklist:

- [ ✔] I have performed a self-review of my own code
- [ ✔ ] I have made corresponding changes to the documentation (if applicable)


docs/examples/random.md
```diff
@@ -0,0 +1,75 @@
+# opsdroid skill-random
```
jacobtomlinson: Let's have a more useful title here that describes what this page is about. Perhaps something like 'Creating a skill for making random chance decisions'.

docs/examples/random.md
```diff
@@ -0,0 +1,75 @@
...
+
+A skill for [opsdroid](https://github.com/opsdroid/opsdroid) to perform random events like flipping a coin or rolling a dice.
```
jacobtomlinson: We are talking to the reader here and telling them what we will cover in this page. Perhaps something like 'we will create a skill for performing random events...'

You can go into more detail about what the skill will do, why is it useful?

docs/examples/random.md
```diff
@@ -0,0 +1,75 @@
...
+## Creating skills
+
+Skills are designed to be simpler than other modules to ensure that it is easy to get started.
+
+To create a skill you need to at minimum create a single python file in your repository with the  `__init__.py`  name. For example the skill  `hello`  has a single file called  `__init__.py`.
```
jacobtomlinson: We don't care about generic information on 'creating skills', that already exists in the documentation, this is a tutorial on creating a specific skill. 

This section should be talking the user through how to create the right folder and files for the random chance skill they are creating.

docs/examples/random.md
```diff
@@ -0,0 +1,75 @@
...
+
+##  Configuration
+```yaml
+None.
+```
```
jacobtomlinson: This section should not say `None.`.

How does the user configure opsdroid to use their random chance skill? Tell them that here.

docs/examples/random.md
```diff
@@ -0,0 +1,75 @@
...
+@match_regex(r'flip a coin', case_sensitive=False)
+
+async def flip_a_coin(opsdroid, config, message):
+
+	result = random.choice(["heads", "tails"])
+
+	text = random.choice(COIN_FLIP_RESPONSES).format(result=result)
+
+	await message.respond(text)
+```
```
jacobtomlinson: Copying and pasting the skill code here is not good enough. There aren't even any comments in the code explaining what is going on.

You should slowly walk the user through each section, explaining why it is there and what it does. What does each function call do, how does it help?

You should definitely include the full skill code in a single block like this, just after you have explained it all in pieces.

docs/examples/random.md
```diff
@@ -0,0 +1,75 @@
...
+
+> user: flip a coin
+>opsdroid: It landed on heads
+
+    roll a dice
+   Rolls a dice.
+   
+
+> user: roll a dice
+>opsdroid: You rolled a 4
```
jacobtomlinson: The markdown formatting is off here. You could add more words to explain to the user what is happening here.

mkdocs.yml
```diff
@@ -13,6 +13,7 @@ pages:
       - "Example Skills":
               - "Introduction": "examples/introduction.md"
               - "Weather": "examples/weather.md"
+              - "random": "examples/random.md"
```
jacobtomlinson: This is the documentation index. Please use correct capitalisation and try to make it clear what the skill does in a short number of words. Perhaps like this.

```suggestion
              - "Random Chance": "examples/random.md"
```

imrishabh18:
@jacobtomlinson Thank you for your review. I am working on the changes.

PR #1212, opened 2019-10-28
Add configuration validation and refactor load_config_file method

FabioRosado:
# Description

Started working on refactoring the load_config_file and might have gone too crazy on creating helper functions 😬 I'm opening this PR just to get your feedback on the work done so far and what you would like me to change stuff.

All tests are failing at the moment but did some manual testing and works okay

Fixes #<issue>

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Ran opsdroid - worked as expected
- Tested schema validation and exceptions - working as expected
- Tested envvar constructor - works as expected
- Ran command `opsdroid config lint` - worked as expected - failed/passed when should
- Tox - All green

# Checklist:


docs/configuration-reference.md
```diff
@@ -536,7 +536,6 @@ You can split the config into smaller modules by using the value `!include file.
 
 ```yaml
 skills: !include skills.yaml
-
 ```
 
-_Note: The file.yaml that you wish to include in the config must be in the same directory as your configuration.yaml (e.g ~/.opsdroid)_
+Since we use [appdirs](https://pypi.org/project/appdirs/) to keep our opsdroid files in a logical location depending on what operative system you use, the included file should be located on your [user configuration directory](https://github.com/ActiveState/appdirs#some-example-output).
```
jacobtomlinson: I don't think the users need to know that we use `appdirs` for this. It's probably unnecessary technical noise. Maybe just saying that files are stored in appropriate user configuration directories depending on which operating system you use.

opsdroid/configuration/constructors.py
```diff
@@ -0,0 +1,26 @@
...
+    return os.environ[env_var]
+
+
+def include_constructor(loader, node):
+    """Add a yaml file to be loaded inside another."""
+    main_yaml_path = appdirs.user_config_dir(NAME, appauthor=False)
+    included_yaml = os.path.join(main_yaml_path, loader.construct_scalar(node))
+
+    with open(included_yaml, "r") as included:
+        return yaml.load(included, Loader=yaml.CSafeLoader)
```
jacobtomlinson: I think putting these in their own file is unnecessary. They would be just fine in the `__init__.py`.

opsdroid/configuration/validation.py
```diff
@@ -0,0 +1,33 @@
+"""Schema validation for configuration.yaml."""
+
+from voluptuous import Schema, ALLOW_EXTRA
+
+
+schema = {
+    "logging": {"level": bool, "console": bool},
+    "welcome-message": bool,
+    "connectors": [{"name": str, "token": str, "access-token": str}],
```
jacobtomlinson: Not all connectors have a `token` or `access-token` here, is that ok?

opsdroid/core.py
```diff
@@ -14,6 +14,7 @@
 from opsdroid.const import DEFAULT_CONFIG_PATH
 from opsdroid.memory import Memory
 from opsdroid.connector import Connector
+from opsdroid.configuration import load_config_file
```
jacobtomlinson: Moving this here makes a load of sense to me.

opsdroid/configuration/validation.py
FabioRosado: Yeah it's alright because the lib will check if the key exists, if it does it will check for type if not it will skip which is great when it comes to some connectors. 
This schema is just a test one I still need to test it further since I'm having issues with `parsers` and `databases` because when they are set it will be a list, when they aren't set it will be just None so I need to check how to deal with it

docs/configuration-reference.md
FabioRosado: Good point will fix this 👍

FabioRosado:
I'm still waiting for the reply from the issue that I created on the voluptuous repository, but I'm afraid the schema will have to specify all of the possible params and then check for type.

This last update finishes the schema. Although while testing I've noticed two things:
- min-score breaks the validation - it expects a dictionary even though I specified an int
- databases can't be None otherwise the validation fails

Should the databases be an empty list like what happens to the parsers if none is set up? The issue is that I can't select None as default because that will make default be none and not type (using `type(None)` doesn't work either).

This PR is getting closer to be completed. One thing I've noticed was that the error will be shown like this:
`Configuration contains an error - expected bool for dictionary value @ data['connectors'][0]['name']`

Should I try to pass the name of the configuration param that failed instead of the `data['connectors'][0]['name']`? If so we will need to do something like this:

`location = data[error.path[0]][error.path[1]]['name']` 

This might look a bit ugly so what do you think?

jacobtomlinson:
Can we update the schema in chunks? I.e each connector/database/parser has a voluptuous section that the top which adds it's own parameters to the list? It would help break things down even if we are just updating a larger list.

FabioRosado:
I've updated the file to have each config param split up - I am assuming this is what you meant? I'm still eager to have validation for each connector/parser/etc but we will need to wait for anyone to comment on the issue I raised with voluptuous.

jacobtomlinson:
Yeah ok. Thanks for working on this. I think we should put it on hold until they get back to your issue. Is that ok?

FabioRosado:
This PR is almost done. 
I liked the idea that @iguyking of adding a validation flag for the cli so added it to the PR as well. Perhaps it could be good if we could pass a path to the flag so we can validate it but I'm not sure how to allow additional commands.

I tried to split the modules by type like suggested and this was the best way I found to do such. I'm still not sure if we should ditch the `Optional()` bit or not 🤔 

We have tallked about replacing all the `api-token` and `access-token` with the same rule `token`, not sure if we still want to do that now that we can check the rules by module.

Finally, I noticed that all the available flags aren't documented, should we add it somewhere? If so and since the configuration reference is being quite big should we refactor it as well?

jacobtomlinson:
> Finally, I noticed that all the available flags aren't documented, should we add it somewhere? If so and since the configuration reference is being quite big should we refactor it as well?

I'm working on a documentation rewrite at the moment. So I'll incorporate this over there.

opsdroid/cli/config.py
```diff
@@ -49,3 +49,10 @@ def gen(ctx):
 def edit(ctx):
     """Print out the example config."""
     edit_files(ctx, None, "config")
+
+
+@config.command()
+@click.pass_context
+def validate(ctx):
```
jacobtomlinson: This is a really nice touch, I hadn't thought of this.

However I think I would prefer the word `lint` to `validate`. Thoughts?

FabioRosado: I don't really mind changing this to lint, the only reason why I choose validate was to combine with the config command. If you prefer the word lint I'm happy to change it 😄

FabioRosado:
Initially, when I was writing the schema it didn't seem too bad, now that we have all the modules in the `validation.py` (but skills) I agree that this is less ideal and would cause a burden to maintain since in a year or so we will probably forget to update the schemas.

From the three suggestions, I am more inclined to go with either the 1st or the 2nd one. Mostly due to i**guyking** suggestion to have a cli way to validate changes of the config without the need to start opsdroid.

After thinking a bit I think the first option might be the best course of action. We have a lot of modules in opsdroid core and a lot of modules in the repository, but opsdroid is mean to allow users to change whatever they want and even create whole modules for their needs. This is an assumption but I guess it could add value to allow users to specify schemas for their custom made modules and get them validated when opsdroid starts.

I think we could add a const variable called `CONFIG_SCHEMA` like you suggested and use that helper function to get the details of this variable (I'm assuming this is what you had in mind as well). The only downside is that we will probably need to import voluptuous.Required on a few modules but don't think this will be problematic.

The only thing I have an issue at the moment is to picture the helper function and what to do to implement that now so I need to think about this further 😄

opsdroid/cli/config.py
jacobtomlinson: Yeah would you mind changing it? I think `lint` is used in quite a few cli tools and would feel more natural.

jacobtomlinson:
I've also been thinking about it and am leaning more towards option three 😂. The other two get us into a bit of a chicken and egg situation.

- 🐓 We need to import modules to get their schema. 
- 🥚 We need to load the config and start loading opsdroid in order to know which modules to import.

The other issue with this approach is that we could end up with a load of unnecessary load time imports. We would need to import all of the connectors in order to build the full schema, even if the user is only using one connector.

Opsdroid startup currently looks roughly like this:

- CLI Start
  - Check runtime dependencies
  - Load config
    - **Validate config**
  - Configure i11n
  - Configure logging
  - Opsdroid run
    - Load modules from config (all module imports happen here)
    - Setup skills
    - Create web server
      - Setup skill webhooks
    - Train NLU models
    - Start database connections
    - Start connectors
    - Start cron service (this will be moved into a connector in future)
    - Start web server

And your CLI validation implementation here looks like this:

- CLI Validate
  - Load config
    - **Validate config**

If we want true dynamic validation where all connectors, databases and skills (core and third party) can register their own schemas we would need to wait for validation until at least after the modules have been loaded from the config. However, we will have already used the config a bunch of times for all of the other configuration and to decide which modules to load, so really we could do with some level of validation early on too.

This general validation would probably just need to check opsdroid level stuff like logging, languages, the web server config, welcome message, etc and that all modules are a dict.

- CLI Start
  - Check runtime dependencies
  - Load config
    - **Validate general config**
  - Configure i11n
  - Configure logging
  - Opsdroid run
    - Load modules from config (all module imports happen here)
      - **Validate each module config**
    - Setup skills
    - Create web server
      - Setup skill webhooks
    - Train NLU models
    - Start database connections
    - Start connectors
    - Start cron service (this will be moved into a connector in future)
    - Start web server

This would mean your validation implementation would need to change to:

- CLI Validate
  - Load config
    - **Validate general config**
  - Opsdroid **Validate** (new method in the core?)
    - Load modules from config (all module imports happen here)
      - **Validate each module config**

This isn't actually a huge amount of opsdroid loading, we would still be bailing out about half way through. However it will trigger all of the module generation from the `Loader` which will in turn make network requests that could take time. However if someone configures a skill stored in a git repo the only way we are going to be able to validate the config for the skill is to pull the repo down and generate the module.

I just cant see a way around something like this.

FabioRosado:
After thinking about this issue for a bit I think it could make sense doing it this way.

This is what I thought of: 
- Do the general validation like you suggested - just basic things.
- On `_load_modules` when we do the check if `if module is not None:` and append it to the loaded module - do another check on the config for the module.

This would work because we allow extra keys and we can check the config partially - so we can just skip all the things that we had added to the config on the `setup_module`. 

A way to do this would be change the validation function to allow a schema to be passed as param and if this is passed then we would validate the data with that schema - that way each module would be validated in turn.

The only thing I am having an issue with is how to pass the expected rules to the module. Using a const variable makes sense and helps maintainability but how should we get it over to the general config? 🤔

jacobtomlinson:
> A way to do this would be change the validation function to allow a schema to be passed as param and if this is passed then we would validate the data with that schema - that way each module would be validated in turn.

Yeah so we just check each module against it's own schema. Which would result in lots of little schema checks at this stage.

> The only thing I am having an issue with is how to pass the expected rules to the module. Using a const variable makes sense and helps maintainability but how should we get it over to the general config? 🤔

I think a constant makes most sense. I would imagine we could insert something [here](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/loader.py#L569-#L571) along the lines of

```python
# Import module
self.current_import_config = config
module = self.import_module(config)
try:
    validate(config, module.CONFIG_SCHEMA)
except Exception as e:
    self.opsdroid.critical("Config for %s failed validation!" % config['name'], 1)
```

Then in a skill or other module you would have something like

```python
from voluptuous import Optional

CONFIG_SCHEMA = {
        "some-option": str,
        Optional("some-other-option"): str
    }

class HelloSkill(Skill):
    @match_regex(r'hi')
    async def hello(self, message):
        await message.respond('Hey')
```

FabioRosado:
Got it, I'm going to play around with this idea and see how it goes. Hopefully I'll be able to finish this PR soon 😄

FabioRosado:
Happy to say that the PR is pretty much done 😄 

In order to make the command `opsdroid config lint` work I had to use the Loader to load the modules from the config and run the validation since this is done module by module. 
Since I am loading the config the message related to no databases being set up will show, but I don't think this would be a big issue 🤔

jacobtomlinson:
Great thanks! I'll try and get some time to review it, it's a big one 😆.

You've also reminded me I was considering adding the `sqlite` database as enabled by default so that there is always a database (and therefore no warnings). User's can still disable this in their config if they wish.

PR #1243, opened 2019-11-06
Change configuration layout - use dictionaries instead of list of dicts

FabioRosado:
# Description

As discussed on matrix I am raising this PR that should change the configuration layout to use dictionaries instead of list of dicts. This will help with validation on #1212 .

All tests will fail atm since I haven't touched them yet. Would like to get some feedback to see if I'm on a good path. 

I'm thinking to maybe add a version check when running the helper function. If version is > than 0.16.0 we won't need to do anything since the config should be the right format already. What do you think?

I've also changed a few `{}` to use `dict()` since in python 3.6 it seems the code runs a fraction faster 🤷‍♂️

The `loaded_modules` from `_load_modules` is still a list I starting touching it but then became unsure if I should change this bit.

## Status
**READY**| ~~**UNDER DEVELOPMENT**~~ | ~~**ON HOLD**~~

## Type of change

- Documentation (fix or adds documentation

# How Has This Been Tested?

- Tox - all green
- Ran opsdroid with parser set up - parsed as expected
- Ran opsdroid with connector set up - connected as expected
- Ran opsdroid - all worked as expected

# Checklist:

### DIFF
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -29,6 +29,7 @@
     - [Disable dependency install](#disable-dependency-install)
   - [Environment variables](#environment-variables)
   - [Include additional yaml files](#include-additional-yaml-files)
+  - [Migrating to new configuration layout - post v0.16.0](#migrate-to-new-configuration-layout)
 
 ## Config file
 
@@ -58,10 +59,10 @@ For example, a simple barebones configuration would look like:
 
 ```yaml
 connectors:
-  - name: shell
+  shell: {}
 
 skills:
-  - name: hello
+  hello: {}
 ```
 
 This tells opsdroid to use the [shell connector](https://github.com/opsdroid/connector-shell) and [hello skill](https://github.com/opsdroid/skill-hello) from the official module library.
@@ -74,19 +75,19 @@ A more advanced config would like similar to the following:
 
 ```yaml
 connectors:
-  - name: slack
+  slack:
     token: "mysecretslacktoken"
 
 databases:
-  - name: mongo
+  mongo:
     host: "mymongohost.mycompany.com"
     port: "27017"
     database: "opsdroid"
 
 skills:
-  - name: hello
-  - name: seen
-  - name: myawesomeskill
+  hello: {}
+  seen: {}
+  myawesomeskill: {}
     repo: "https://github.com/username/myawesomeskill.git"
 ```
 
@@ -120,11 +121,11 @@ _Config options of the connectors themselves differ between connectors, see the 
 ```yaml
 connectors:
 
-  - name: slack
+  slack:
     token: "mysecretslacktoken"
 
   # conceptual connector
-  - name: twitter
+  twitter:
     oauth_key: "myoauthkey"
     secret_key: "myoauthsecret"
 ```
@@ -138,7 +139,7 @@ Example:
 
 ```yaml
 connectors:
-  - name: slack
+  slack:
     token: "mysecretslacktoken"
     thinking-delay: <int, float or two element list>
     typing-delay: <int, float or two element list>
@@ -164,7 +165,7 @@ _Config options of the databases themselves differ between databases, see the da
 
 ```yaml
 databases:
-  - name: mongo
+  mongo:
     host: "mymongohost.mycompany.com"
     port: "27017"
     database: "opsdroid"
@@ -209,11 +210,11 @@ logging:
   console: true
 
 connectors:
-  - name: shell
+  shell: {}
 
 skills:
-  - name: hello
-  - name: seen
+  hello: {}
+  seen: {}
 ```
 
 #### Optional logging arguments
@@ -345,11 +346,11 @@ Set the path for opsdroid to use when installing skills. Defaults to the current
 module-path: "/etc/opsdroid/modules"
 
 connectors:
-  - name: shell
+  shell: {}
 
 skills:
-  - name: hello
-  - name: seen
+  hello: {}
+  seen: {}
 ```
 
 ### Parsers
@@ -360,11 +361,11 @@ _Config options of the parsers themselves differ between parsers, see the parser
 
 ```yaml
 parsers:
-  - name: regex
+  regex:
     enabled: true
 
 # NLU parser
-  - name: rasanlu
+  rasanlu:
     url: http://localhost:5000
     project: opsdroid
     token: 85769fjoso084jd
@@ -383,8 +384,8 @@ _Config options of the skills themselves differ between skills, see the skill do
 
 ```yaml
 skills:
-  - name: hello
-  - name: seen
+  hello: {}
+  seen: {}
 ```
 
 See [module options](#module-options) for installing custom skills.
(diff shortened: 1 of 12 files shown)


opsdroid/core.py
```diff
@@ -286,12 +286,12 @@ def setup_skills(self, skills):
 
         """
         if "parsers" in self.config:
-            parsers = self.config["parsers"] or []
-            rasanlu = [p for p in parsers if p["name"] == "rasanlu"]
-            if len(rasanlu) == 1 and (
-                "enabled" not in rasanlu[0] or rasanlu[0]["enabled"] is not False
+            parsers = self.config["parsers"] or {}
```
jacobtomlinson:
```suggestion
            parsers = self.config.get("parsers", {})
```

opsdroid/core.py
```diff
@@ -286,12 +286,12 @@ def setup_skills(self, skills):
...
+            rasanlu = parsers.get("rasanlu")
+            if rasanlu and (
+                "enabled" not in rasanlu or rasanlu["enabled"] is not False
```
jacobtomlinson: We could make `enabled` a default key in the loader because this pattern is repeated a lot and is a bit gross.

Would be nicer if we could just have

```
            if rasanlu and rasanlu["enabled"]:
```

opsdroid/core.py
```diff
@@ -344,6 +344,7 @@ def _connector_names(self):  # noqa: D401
                 n_key = len(list(filter(lambda x: x.startswith(name), names)))
                 name += "_{}".format(n_key)
             names[name] = connector
+            print(connector)
```
jacobtomlinson: Stray debugging print.
```suggestion
```

opsdroid/helper.py
```diff
@@ -44,6 +44,67 @@ def del_rw(action, name, exc):
...
+
+    """
+    updated_config = {}
+    for config_type, modules in config.items():
+        if (
+            config_type == "parsers"
+            or config_type == "connectors"
+            or config_type == "skills"
+            or config_type == "databases"
+        ):
```
jacobtomlinson: You could simplify this with 

```python
if config_type in ("parsers", "connectors", "skills", "databases"):
```

opsdroid/helper.py
```diff
@@ -44,6 +44,67 @@ def del_rw(action, name, exc):
...
+            del module_copy["name"]
+
+            config[module["name"]] = module_copy
+
+        return config
+    except TypeError:
+        return {}
+
+
+def update_config(config):
```
jacobtomlinson: I would be tempted to make this more descriptive. We making a breaking change to our config and this is here for backward compatibility. Maybe something like `update_pre_0_17_config_format`.

opsdroid/helper.py
```diff
@@ -44,6 +44,67 @@ def del_rw(action, name, exc):
...
+        modules (list): List of dictionaries that contain the module configuration
+
+    Return:
+        List: New modified list following the new format.
+
+    """
+    config = dict()
+
+    try:
+        for module in modules:
```
jacobtomlinson: If the user is using the new format and modules is already a dict then this will have some strange behaviour.

We should check if modules is already a dict and just return it if it is.

We could also add a deprecation warning or some other warning notice if it is a list and point people to the documentation on how to migrate.

opsdroid/core.py
FabioRosado: So interesting thing here. If we do this change then the next line throw out an `AttributeError: 'NoneType' object has no attribute 'get'` when we do `self.config.get("parsers", {})` this expression is evaluating as none (i'm assuming because of the empty dict) whilst the or makes a difference (for some reason).

Leaving the line as it is works just fine 🤷‍♂️

jacobtomlinson: Weird! No worries.

FabioRosado:
I've updated this PR again with latest refactoring and changes that you have requested.

I still need to deal with the coverage issue, write the docs for the new layout and update the yamale files. 

I decided to split _load_modules and create a new method that deals with all the config updating. I've also used `dict.update` because it seems to be a bit faster than direct assign.

```shell
python -m timeit -s 'd = {"key": "value"}' 'd["key"] = "value"'
10000000 loops, best of 3: 0.0621 usec per loop

python -m timeit -s 'd = {"key": "value"}' 'd.update({"key": "value"})'
1000000 loops, best of 3: 0.261 usec per loop
```

I'm not sure if you like this new format but doesn't seem to be losing readability. 

I'm wondering if we should turn the _load_modules into a dict of itself instead of a list containing all the dicts. If so should we update this to have a dict containing the config for each module? At the moment the loaded_modules looks like this:

```python
[
    {
        'module': <module 'opsdroid.connector.telegram' from '<path>'>,
        'config': {
            'name': 'telegram',
            'module': '',
            'type': 'connector',
            'enabled': True,
            'entrypoint': None,
            'is_builtin': ModuleSpec(name='opsdroid.connector.telegram', loader=<_frozen_importlib_external.SourceFileLoader object at 0x105b1a0f0>, origin='<path>', submodule_search_locations=['<path>']),
            'module_path': 'opsdroid.connector.telegram',
            'install_path': '<path>',
            'branch': 'master',
            'token': '<API TOKEN>'
            },
        'intents': None
    },
    {
        'module': <module 'opsdroid.connector.websocket' from '<path>'>,
        'config': {
            'name': 'websocket',
            'module': '',
            'type': 'connector',
            'enabled': True,
            'entrypoint': None,
            'is_builtin': ModuleSpec(name='opsdroid.connector.websocket', loader=<_frozen_importlib_external.SourceFileLoader object at 0x105b1ab70>, origin='<path>', submodule_search_locations=['<path>']),
            'module_path': 'opsdroid.connector.websocket',
            'install_path': '<path>',
            'branch': 'master',
            'bot-name': 'mybot',
            'max-connections': 10,
            'connection-timeout': 10
            },
        'intents': None
    }
]
```

Also apologies if this doesn't make much sense I'm a bit tired at the moment haha

FabioRosado:
I've commented out the validation bit since it was causing problems with the new version and was a bit of headache to make it work. I'm assuming this is not too bad since once this PR is merged I'll work on updating the validation of the configuration from #1212 

I've tested using opsdroid and everything seems to be working okay.

Still wondering about returning `loaded_modules` as a list. 

Finally, I got no idea why appveyor stopped working 🤔

docs/configuration-reference.md
```diff
@@ -74,19 +75,19 @@ A more advanced config would like similar to the following:
...
     port: "27017"
     database: "opsdroid"
 
 skills:
-  - name: hello
-  - name: seen
-  - name: myawesomeskill
+  hello: {}
+  seen: {}
+  myawesomeskill: {}
```
jacobtomlinson: As this has a key `repo` we should leave out the `{}`.
```suggestion
  myawesomeskill: 
```

FabioRosado:
Yeah I completely missed that zero between the dot 🤦‍♂️
I've reverted the changes made to use config.update() and went back to dict assign. 

Do you think we need to do anything else with the PR now or should be ok to be merged? 🤔

jacobtomlinson:
Yeah I think this is basically there now. I would maybe like to see some more tests that actually load in example config files with both the old and new formats.

Also could you have a loook at #1252 as it touches some of the same config. I've been working on it quietly for a while. I'd like to get it in but it may cause conflicts here.

PR #1252, opened 2019-11-14
Add RSS/Atom feed connector/matcher

jacobtomlinson:
# Description

This PR adds a new `match_rss` matcher which allows skills to be triggered when a new feed item is created. I've implemented this as a connector under the hood which is enabled by default, this aligns with the plans in #933 to make connectors the place to generate events. 

```python
from opsdroid.matchers import match_rss

class RSSSkill(Skill):
    """An example RSS skill class."""

    @match_rss('https://examples.com/feed.rss', interval=30)
    async def myrssskill(self, event):
        """Send a message to our default room/connector when we get a new feed item."""
        await self.opsdroid.send(
            Message(f"Just got a new feed item called '{event.item.title}'.")
        )
```

In order to enable the RSS connector by default I've also made some changes to the way that the config is loaded. Instead of just loading the user's config we are now loading a default config file first (`opsdroid/configuration/default.yaml`) and then updating that with the user's config. That way we can have the RSS connector enabled by default but the user can override it and disable it.

## Status
**READY** 

## Type of change

# How Has This Been Tested?

- Added new unit tests
- Run maually with https://lorem-rss.herokuapp.com (which is awesome btw)

# Checklist:


Cadair:
@jacobtomlinson I could really use this for a hack I am working on, any chance you are going to have time to revive it? Is the design still sensible?

jacobtomlinson:
It's on my list of things to do. It's probably going to need a bit fo a rewrite now though.

jacobtomlinson:
Closing as super stale

PR #1256, opened 2019-11-15
delete operation in opsdroid memory and database connectors

### LINKED ISSUE #1253: Ability to delete a key from Opsdroid memory
# Description
The Opsdroid's memory class currently has two methods: `get(key)` and `put(key, object)` to fetch and persist data in memory. It would be good to have an additional method `delete(key)` to delete the data from memory.

daniccan:
# Description

Added `delete` operation to the opsdroid `memory` class and the different `database` connectors.

Fixes #1253 

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Tested with all existing test cases of opsdroid memory and database connectors.
- Tested with additional test cases for delete operation in opsdroid memory and database connectors.

# Checklist:


opsdroid/database/mongo/__init__.py
```diff
@@ -62,3 +62,13 @@ def __init__(self, config, opsdroid=None):
...
         )
+
+    async def delete(self, key):
+        """Delete a document from the database (key).
+
+        Args:
+            key (str): the key is the database name.
+
+        """
+        logging.debug("Deleting %s from mongo", key)
```
jacobtomlinson: Strange that this file isn't using the `_LOGGER` pattern. In most files we have `_LOGGER = logging.getLogger(__name__)` somewhere at the top and then would use `_LOGGER.debug` here.

I appreciate this isn't your problem but would you mind changing this file to follow the pattern? See the redis database for reference.

opsdroid/database/sqlite/__init__.py
```diff
@@ -116,3 +116,18 @@ def __init__(self, config, opsdroid=None):
...
+            key (string): The key to delete in the database.
+
+        """
+        _LOGGER.debug(_("Deleting %s from sqlite"), key)
+
+        async with aiosqlite.connect(self.db_file, **self.conn_args) as _db:
+            cur = await _db.cursor()
+            await cur.execute("DELETE FROM {} WHERE key=?".format(self.table), (key,))
+
+        self.client = _db
```
jacobtomlinson: This assignment seems strange. But it seems to be a common pattern throughout this database module.

I expect we should be creating `self.client` in the connect method, and probably not in a context manager either. We should then be closing the client in a disconnect method (which doesn't appear to be implemented here). Then the get, put, delete should all use the client instead of making a new connection.

This feels out of scope for this PR. If you want to fix it now then please go ahead, but I'm also happy to just take it as is. Up to you.

tests/test_database_redis.py
```diff
@@ -100,6 +100,15 @@ class TestRedisDatabaseAsync(asynctest.TestCase):
 
         result = await db.put("string", dict(key="value"))
 
+    async def test_delete(self):
+        db = RedisDatabase({})
+        db.client = MockRedisClient()
+        db.client.execute = amock.CoroutineMock(return_value=None)
```
jacobtomlinson: I was having some CI issues with this pattern yesterday. Could you change it to the following, just for peace of mind.

```suggestion
        db.client.execute = amock.CoroutineMock()
        db.client.execute.return_value = None
```

opsdroid/database/mongo/__init__.py
daniccan: Made the mentioned changes.

opsdroid/database/sqlite/__init__.py
daniccan: Made the changes and added a `disconnect` method too.

tests/test_database_redis.py
daniccan: Made the changes as mentioned.

daniccan:
@jacobtomlinson The build timed-out in travis-ci. Is there a way to retry the build?

opsdroid/database/sqlite/__init__.py
```diff
@@ -126,8 +123,11 @@ def __init__(self, config, opsdroid=None):
         """
         _LOGGER.debug(_("Deleting %s from sqlite"), key)
 
-        async with aiosqlite.connect(self.db_file, **self.conn_args) as _db:
-            cur = await _db.cursor()
-            await cur.execute("DELETE FROM {} WHERE key=?".format(self.table), (key,))
+        cur = await self.client.cursor()
+        await cur.execute("DELETE FROM {} WHERE key=?".format(self.table), (key,))
+        self.client.commit()
```
jacobtomlinson: Need to await this

opsdroid/database/sqlite/__init__.py
```diff
@@ -126,8 +123,11 @@ def __init__(self, config, opsdroid=None):
...
 
-        self.client = _db
+    async def disconnect(self):
+        """Disconnect from the database."""
+        if self.client:
+            self.client.close()
```
jacobtomlinson: Need to await this

opsdroid/database/sqlite/__init__.py
```diff
@@ -82,14 +82,13 @@ def __init__(self, config, opsdroid=None):
...
-                "INSERT INTO {} VALUES (?, ?)".format(self.table), (key, json_data)
-            )
-
-        self.client = _db
+        cur = await self.client.cursor()
+        await cur.execute("DELETE FROM {} WHERE key=?".format(self.table), (key,))
+        await cur.execute(
+            "INSERT INTO {} VALUES (?, ?)".format(self.table), (key, json_data)
+        )
+        self.client.commit()
```
jacobtomlinson: This needs to be awaited

daniccan:
@jacobtomlinson The build seems to fail due to timeout. Anyway to resolve this?

jacobtomlinson:
@daniccan restarted the build.

PR #1270, opened 2019-11-25
Update regex pattern for the envvar construtor

### LINKED ISSUE #1240: Update regex pattern for the envvar construtor
You can use envvars on your configuration, but they need to follow this pattern `$ENVVARNAME` the regex pattern should be updated to allow users to use either `$ENVVARNAME` or `${ENVVARNAME}`. 

While we are at it we should allow users to use other characters as well like `_` or `-`. 

This change needs to be done on [opsdroid.loader.load_config_file](https://github.com/opsdroid/opsdroid/blob/674013037eab826640174407a73f8fed1a29b290/opsdroid/loader.py#L347)

benjpalmer:
# Description

Update environment variable regex pattern to be more inclusive of comon naming conventions. Create constant for env_var regex to prevent duplicating the pattern between testing and actual usage. Add tests for invalid and valid env variables.

Fixes #1240 

## Status
**READY** 

## Type of change

# How Has This Been Tested?

-Added `test_env_var_regex()` to ` tests.test_loader.TestLoader`
This test confirms valid test environment variables will match the updated regex pattern and inversely confirms invalid test environment variables will not match. No additional test configurations are needed. Test data is included within the test itself.

# Checklist:


tests/test_loader.py
```diff
@@ -165,6 +166,24 @@ def test_yaml_load_exploit(self):
...
+            {"env_var": "$OPS_DROID23", "match": False},
+            {"env_var": "$UPPER-AND-lower", "match": False},
+            {"env_var": '"MISSING_PREFIX"', "match": False},
+        ]
+        for d in test_data:
+            # Tests opsdroid constant ENV_VAR_REGEX for both valid and invalid environment variables.
+            self.assertRegex(d["env_var"], ENV_VAR_REGEX) if d[
+                "match"
+            ] else self.assertNotRegex(d["env_var"], ENV_VAR_REGEX)
+
```
jacobtomlinson: This is a nice implementation but this is a little hard to read. I don't think black has helped you here either.

Would you mind changing it to 

```python
if d["match"]:
    self.assertRegex(d["env_var"], ENV_VAR_REGEX)
else:
    self.assertNotRegex(d["env_var"], ENV_VAR_REGEX) 
```

tests/test_loader.py
```diff
@@ -165,6 +166,24 @@ def test_yaml_load_exploit(self):
...
+        test_data = [
+            {"env_var": "$OPS_DROID", "match": True},
+            {"env_var": "$OPS-DROID", "match": True},
+            {"env_var": "${OPS_DROID}", "match": True},
+            {"env_var": '"$OPSDROID_SLACK_TOKEN"', "match": True},
+            {"env_var": "$INVALID!_CHARACTERS@", "match": False},
+            {"env_var": "OPS_DROID", "match": False},
+            {"env_var": "$OPS_DROID23", "match": False},
+            {"env_var": "$UPPER-AND-lower", "match": False},
+            {"env_var": '"MISSING_PREFIX"', "match": False},
```
jacobtomlinson: Could you add a few more test cases here? Perhaps the following:

- All lowercase
- A simple sentence
- Some numbers (as a string)
- An long string with random characters including `$`s
- A `$` on its own
- An empty variable `${}`
- An empty string
- All symbols `${_}`

jacobtomlinson:
Also there are a couple of merge conflicts from another PR I just merged. If you are comfortable resolving them would you mind? Shout if you need guidance.

tests/test_loader.py
benjpalmer: Thanks for the feedback! I added more test data to cover the cases outlined above. This also led to a more exclusive regex pattern.

tests/test_loader.py
benjpalmer: No problem!

FabioRosado:
This is great! Thank you so much for working on this issue and making the changes that Jacob asked. I'm going to merge this PR now 😄 👍 

PR #1271, opened 2019-11-27
Load config from path and show active modules

### LINKED ISSUE #634: Pass path to config as cli argument
The new config file location appears to cause confusion with existing users as they end up in a state where opsdroid is using the new config but the old config still exists.

Here is some feedback via Gitter/Matrix.

> jacobtomlinson: I would be really interested to hear feedback on how we could improve this transition. Perhaps we should empty .opsdroid and put in a warning file.
>
> colonelkrud: Probably a start blocking error that asks you to check and remove the old path if it exists.. rename it to stop the error. That way you 100% know the person running the bot is aware of the change

FabioRosado:
# Description

I've been playing around with click and implemented the `list-connectors` idea that you mentioned in matrix when we were talking about it. Now that I look at it perhaps we should just have a single command 'list-modules' that will list all active connectors, db, parsers, etc. I'm thinking since the code is pretty much repeated for all the modules type.

I've also added the possibility to start opsdroid from a config path (#634), this also made possible to change the lint command and check validation of a config file from a path.

Since I am repeating the `if path` code twice perhaps this bit should be added to an helper function? 

I'm also playing around with the  -set flag as described in #634 but that would mean using the whole start code and just change the config before starting opsdroid, that's why I thought we could use a helper function (based on utils) to get the config loaded and then touch it if needed 🤔 

What do you think?

Fixes some of  #634

## Status
**READY** |~~**UNDER DEVELOPMENT**~~ | ~~**ON HOLD**~~

## Type of change

- Documentation (fix or adds documentation) - by docstrings

# How Has This Been Tested?

- Ran `opsdroid config -f <path> lint` - validated good/bad config
- Ran `opsdroid config start -f <path>` - started opsdroid with different config
- Ran opsdroid with bad config type - failed as expected
- Tox - green but missing some coverage

# Checklist:

- [x] I have made corresponding changes to the documentation (if applicable) - docstrings


opsdroid/cli/start.py
```diff
@@ -9,20 +9,23 @@
...
     check_dependencies()
 
-    config = load_config_file(
-        ["configuration.yaml", DEFAULT_CONFIG_PATH, "/etc/opsdroid/configuration.yaml"]
-    )
+    if path:
+        DEFAULT_CONFIG_LOCATIONS.insert(0, path)
+
+    config = load_config_file(DEFAULT_CONFIG_LOCATIONS)
+
```
jacobtomlinson: I would prefer that if a user sets the path flag we don't check the default locations at all. Because if they specify a bad path we will just move on to the default. This could be confusing because opsdroid would still start in this case.

opsdroid/cli/utils.py
```diff
@@ -160,3 +162,46 @@ def welcome_message(config):
...
+                )
+            )
+
+        if param == "parsers":
+            parsers = [parser for parser in config.get("parsers")]
+            click.echo(
+                "Parsers active in the configuration: {parsers_active}.".format(
+                    parsers_active=", ".join(parsers)
+                )
+            )
```
jacobtomlinson: You could condense all this down to `things = [thing for thing in config.get(param)] `. It would also be nice to get some more information here in a tabular format

We should echo each module on a separate line. For example if I have the `hello`, `dance` and a custom skill configured it would be nice to get some output like:

```
NAME     TYPE   MODE  CACHED  LOCATION
hello    skill  repo  yes     https://github.com/opsdroid/skill-hello.git
dance    skill  repo  yes     https://github.com/opsdroid/skill-dance.git
myskill  skill  path  no      /path/to/my/skill
```

Or if you have the slack connector configured you would get


```
NAME     TYPE       MODE    CACHED  LOCATION
slack    connector  module  yes     opsdroid.connector.slack
```

This kind of layout could make it nice to operate on from the command line with tools like `grep` and `xargs`.

opsdroid/configuration/validation.py
```diff
@@ -58,3 +58,17 @@ def validate_configuration(config, schema):
...
+
+def validate_data_type(data):
+    """Validate data type.
+
+    The configuration needs to be a dict. Since yaml loads any file
+    that can be decoded by utf-8, some files will be parsed and returned
+    as a single string. If data is not a dict we raise a TypeError.
+
+    """
+    if type(data) != dict:
```
jacobtomlinson: It is preferable to use `isinstance(thing, type)` over `type(thing) == type`

```suggestion
    if not isinstance(data, dict):
```

jacobtomlinson:
I've added some comments which should help. You're going in a good direction here.

I wouldn't worry about `--set` right now. Try and keep PRs small and don't try and solve too many things in one (I know I'm a hyppocrite here sometimes). There is [a helper](https://github.com/opsdroid/opsdroid/pull/1252/files#diff-994fe42274148c7b03ef72ad4843cc6aR311) in #1252 which will help you anyway.

opsdroid/cli/start.py
FabioRosado: If the user gives a path that doesn't exist then click will throw an error and won't load. It will be something like:

```
Usage: __main__.py start [OPTIONS] [PATH]
Try "__main__.py start --help" for help.

Error: Invalid value for "[PATH]": Path "/Users/fabiorosado/Desktop/testt.yaml" does not exist.
```

With that in mind, should I just load the path that the user gives and ignore the rest of the default configs? 🤔

jacobtomlinson: Ah fair enough. Clever click!

Yeah I would just ignore the rest.

opsdroid/cli/utils.py
FabioRosado: I thought this would be something easy to implement but came around and bit me.  Perhaps I should revert these changes a create a new PR for this 🤔 

I'm thinking that perhaps a way to do this would be to actually load the config and check if its cached and the location. What do you think?

FabioRosado:
I've marked this PR as ready for review since the cli options all work and Travis is happy. I'm wondering if the helper function should return a dict with meaningful keys instead of the tuple (or triple?)

Screenshot 2019-11-27 at 18 22 01

This is how the data is shown in the terminal. The only thing that I haven't figured was to check if a module is core and then change it, but that would mean touching the helper function again.

I've also made the -f flag for the list command in case users want to check modules from a single place.

The last question I got is - if we have a good way to show modules, should we perhaps just use a single command like `list-modules` and list all modules? 🤔

opsdroid/cli/config.py
```diff
@@ -58,35 +58,64 @@ def edit(ctx):
 
 @config.command()
 @click.pass_context
-@click.argument("path", required=False, type=click.Path(exists=True))
+@click.option(
+    "-f",
+    "path",
+    help="Validates opsdroid configuration from a path.",
+    type=click.Path(exists=True),
+)
```
jacobtomlinson: You've got a lot of code duplication here. I wonder if we can just set this option on the root level click group. It would make sense to have the `-f` flag is most places.

jacobtomlinson:
> The last question I got is - if we have a good way to show modules, should we perhaps just use a single command like list-modules and list all modules? 🤔

Yeah that sounds good.

FabioRosado:
I've removed all the duplicated commands and joined them on a single one: `list-modules`.  This will loop through the config and get the options of all the modules and then show them all at once.
I've added colour to the table descriptors as well since click will handle the case when showing colour is not possible and just shows the plain text.

This is the picture of the command now 
Screenshot 2019-11-28 at 16 14 30

I would still like to find a way to use just one -f flag and pass this to all the options but haven't figured out how to do that. I got 2 days of 12 hours shift so I won't be able to do much until Sunday 👎

jacobtomlinson:
Awesome thanks @FabioRosado.

If it really turns out to be difficult to do one `-f` flag you could always pre bake the decorator, that should at least avoid some code duplication.

```python
# Define once (maybe in opsdroid/cli/utils.py ) and import where you need it
path_option = click.option(
    "-f",
    "path",
    help="Lists all modules active from a configuration loaded from a path.",
    type=click.Path(exists=True),
)
```

```python
from opsdroid.cli.utils import path_option

# Use it on your commands
@cli.command()
@click.pass_context
@path_option
def click_function(ctx, path):
   pass
```

FabioRosado:
I might be missing something because when I add the path option to the config group I can use the -f but the subcommand won't run with that param. I did try to get ctx but don't know how to get the path to the subcommand 🤔

I'll keep reading the docs and try to figure out but meanwhile I used your suggestion to use the `@path_option`

FabioRosado:
Like I said on the matrix room, we just need to add path to `ctx.obj` and get that from the other subcommands. It's a pretty easy way to do it but not that obvious haha

This PR is ready to be merged now 😄

jacobtomlinson:
Good work thanks @FabioRosado

PR #1279, opened 2019-11-29
Slack interactions support

### LINKED ISSUE #1267: Slack Blocks interactive button support
It is currently not possible to take actions on button clicks within Slack Blocks (rich messages in Slack). We should add this.

## Slack Interactivity

_This section is probably useful for adding to the docs once we implement this._

When you click a button in a rich Slack message or use a message action (a [todo list app](https://slack.com/apps/A0NSTB8R3-todoist) may have an "add to list" action that can be performed on any message) Slack will send an event to a separate webhook endpoint. **This is a different endpoint to the Events API and the RTM websocket API. Therefore will require a separate implementation in opsdroid.**

To configure your Slack app for interactive events you need to:

- Open your app's [management dashboard](https://api.slack.com/apps)
- Click on _Interactive Components_ in the sidebar
- Toggle the _Interactivity_ switch on.
- Enter the https URL of your bot's slack interactivity endpoint

image

See the [Slack docs](https://api.slack.com/interactivity/handling#setup) for more info.

## Suggested implementation

To enable this feature we would need to add a new webhook endpoint in the Slack connector. You can find an example of doing this in the [Facebook connector](https://github.com/opsdroid/opsdroid/blob/ca3a7e0fd8d822e0d89af9ee6228e2dccf014ab8/opsdroid/connector/facebook/__init__.py#L39-L47). Perhaps something like `/connector/slack/interactions`.

We would also need to add an event handler for payloads being sent to the webhook. Again see the Facebook connector for [an example](https://github.com/opsdroid/opsdroid/blob/ca3a7e0fd8d822e0d89af9ee6228e2dccf014ab8/opsdroid/connector/facebook/__init__.py#L49-L79).

It looks like there are four main interaction types: `block_actions`, `message_actions`, `views_submissions` and `view_closed` (see the [Slack docs](https://api.slack.com/interactivity/handling#payloads) for more info). We should create an [event class](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/slack/events.py) for each one and construct the event objects to be parsed in the handler.

We could then use the `match_event` matcher to write skills which take actions based on these interactions.

Users of opsdroid would also need to expose their bot to the internet. They could do this quickly and easily using a service like [ngrok](https://ngrok.com/) or [serveo](http://serveo.net/).

daniccan:
# Description

Added support for Interactive Actions in Slack to the Slack Connector.

Fixes #1267 

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Existing and new Test Cases for Slack Connector
- A Skill that interacts with Slack based on the interactive actions.

# Checklist:

### DIFF
--- a/docs/connectors/slack.md
+++ b/docs/connectors/slack.md
@@ -106,12 +106,14 @@ When one of those entry points is triggered, a new aspect is introduced to the [
 
 For example, when you click a button in a rich Slack message or use a message action (a todo list app may have an "add to list" action that can be performed on any message) Slack will send an event to a separate webhook endpoint.
 
-To configure your Slack app for interactive events you need to:
+### Configure Slack App for Interactive Events
 
 - Open your app's [management dashboard](https://api.slack.com/apps)
 - Click on `Interactive Components` in the sidebar.
 - Toggle the `Interactivity` switch on.
-- Save the HTTPS URL (https://slackbot.yourdomain.com/connector/slack/interactions) of your bot's slack interactivity endpoint.
+- Save the HTTPS URL of your bot's slack interactivity endpoint (`/connector/slack/interactions`).
+
+*Note:* The bot's interactivity URL would be of the format: `https://slackbot.yourdomain.com/connector/slack/interactions`
 
 ### [block_actions](https://api.slack.com/reference/interaction-payloads/block-actions)
 
--- a/docs/connectors/slack.md
+++ b/docs/connectors/slack.md
@@ -112,7 +112,7 @@ For example, when you click a button in a rich Slack message or use a message ac
 - Click on `Interactive Components` in the sidebar.
 - Toggle the `Interactivity` switch on.
 - Save the HTTPS URL of your bot's slack interactivity endpoint (`/connector/slack/interactions`).
-    - *Example:* `https://slackbot.opsdroid.com/connector/slack/interactions`
+    - *Example:* `https://slackbot.example.com/connector/slack/interactions`
 
 ### [block_actions](https://api.slack.com/reference/interaction-payloads/block-actions)
 
@@ -125,7 +125,7 @@ class InteractionsSkill(Skill):
 
     @match_event(BlockActions)
     async def slack_interactions(self, event):
-        await self.opsdroid.send(Message("Block Actions interactivity has been triggered."))
+        await event.respond("Block Actions interactivity has been triggered.")
 ```
 
 ### [message_action](https://api.slack.com/reference/interaction-payloads/actions)
@@ -139,7 +139,7 @@ class InteractionsSkill(Skill):
 
     @match_event(MessageAction)
     async def slack_interactions(self, event):
-        await self.opsdroid.send(Message("Message Action interactivity has been triggered."))
+        await event.respond("Message Action interactivity has been triggered.")
 ```
 
 ### [view_submission](https://api.slack.com/reference/interaction-payloads/views#view_submission)
--- a/docs/connectors/slack.md
+++ b/docs/connectors/slack.md
@@ -97,3 +97,74 @@ class BlocksSkill(Skill):
 ```
 
 ![](https://user-images.githubusercontent.com/1610850/58658951-ac523300-8319-11e9-8c2a-011469a436d0.png)
+
+## Interactive Actions
+
+Slack apps can use [interactive features](https://api.slack.com/interactivity) to achieve much more than just one-way communication. Apps can implement a number of interaction entry points that allow users to intentionally invoke a response from the app.
+
+When one of those entry points is triggered, a new aspect is introduced to the [interaction transaction](https://api.slack.com/interactivity/handling) — the interaction payload. This payload is a bundle of information that explains the context of the user action, giving the app enough to construct a coherent response.
+
+For example, when you click a button in a rich Slack message or use a message action (a todo list app may have an "add to list" action that can be performed on any message) Slack will send an event to a separate webhook endpoint.
+
+To configure your Slack app for interactive events you need to:
+
+- Open your app's [management dashboard](https://api.slack.com/apps)
+- Click on `Interactive Components` in the sidebar.
+- Toggle the `Interactivity` switch on.
+- Save the HTTPS URL (https://slackbot.yourdomain.com/connector/slack/interactions) of your bot's slack interactivity endpoint.
+
+### [block_actions](https://api.slack.com/reference/interaction-payloads/block-actions)
+
+```python
+from opsdroid.skill import Skill
+from opsdroid.matchers import match_event
+from opsdroid.connector.slack.events import BlockActions
+
+class InteractionsSkill(Skill):
+
+    @match_event(BlockActions)
+    async def slack_interactions(self, payload):
+        await event.respond(Message("Block Actions interactivity has been triggered."))
+```
+
+### [message_action](https://api.slack.com/reference/interaction-payloads/actions)
+
+```python
+from opsdroid.skill import Skill
+from opsdroid.matchers import match_event
+from opsdroid.connector.slack.events import MessageAction
+
+class InteractionsSkill(Skill):
+
+    @match_event(MessageAction)
+    async def slack_interactions(self, payload):
+        await event.respond(Message("Message Action interactivity has been triggered."))
+```
+
+### [view_submission](https://api.slack.com/reference/interaction-payloads/views#view_submission)
+
+```python
+from opsdroid.skill import Skill
+from opsdroid.matchers import match_event
+from opsdroid.connector.slack.events import ViewSubmission
+
+class InteractionsSkill(Skill):
+
+    @match_event(ViewSubmission)
+    async def slack_interactions(self, payload):
+        await event.respond(Message("View Submission interactivity has been triggered."))
+```
+
+### [view_closed](https://api.slack.com/reference/interaction-payloads/views#view_closed)
+
+```python
+from opsdroid.skill import Skill
+from opsdroid.matchers import match_event
+from opsdroid.connector.slack.events import ViewClosed
+
+class InteractionsSkill(Skill):
+
+    @match_event(ViewClosed)
+    async def slack_interactions(self, payload):
+        await event.respond(Message("View Closed interactivity has been triggered."))
+```
--- a/docs/connectors/slack.md
+++ b/docs/connectors/slack.md
@@ -123,7 +123,7 @@ from opsdroid.connector.slack.events import BlockActions
 
 class InteractionsSkill(Skill):
 
-    @match_event(BlockActions)
+    @match_event(BlockActions, value="click_me_123")
     async def slack_interactions(self, event):
         await event.respond("Block Actions interactivity has been triggered.")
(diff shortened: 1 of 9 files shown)


docs/connectors/slack.md
```diff
@@ -97,3 +97,75 @@ class BlocksSkill(Skill):
...
+
+For example, when you click a button in a rich Slack message or use a message action (a todo list app may have an "add to list" action that can be performed on any message) Slack will send an event to a separate webhook endpoint.
+
+### Configure Slack App for Interactive Events
+
+- Open your app's [management dashboard](https://api.slack.com/apps)
+- Click on `Interactive Components` in the sidebar.
+- Toggle the `Interactivity` switch on.
+- Save the HTTPS URL of your bot's slack interactivity endpoint (`/connector/slack/interactions`).
+    - *Example:* `https://slackbot.opsdroid.com/connector/slack/interactions`
```
jacobtomlinson: Minor request here, could we use a generic domain rather than suggest we may provide some hosting or domain services.

```suggestion
    - *Example:* `https://slackbot.example.com/connector/slack/interactions`
```

opsdroid/connector/slack/events.py
```diff
@@ -42,3 +42,39 @@ def __init__(self, blocks, *args, **kwargs):
...
+        self.payload = payload
+
+
+class ViewClosed(Event):
+    """Event class to represent view_closed in Slack."""
+
+    def __init__(self, payload, *args, **kwargs):
+        """Create object with minimum properties."""
+        super().__init__(*args, **kwargs)
+        self.payload = payload
```
jacobtomlinson: As all these classes are the same other than the docstrings you could create a superclass with this `__init__` method and inherit all classes from that.

opsdroid/core.py
```diff
@@ -494,18 +494,18 @@ def _connector_names(self):  # noqa: D401
...
-                        )
+        unconstrained_skills = await self._constrain_skills(self.skills, event)
+        ranked_skills = await self.get_ranked_skills(unconstrained_skills, event)
+        if ranked_skills:
+            tasks.append(
+                self.eventloop.create_task(
+                    self.run_skill(
+                        ranked_skills[0]["skill"],
+                        ranked_skills[0]["config"],
+                        ranked_skills[0]["message"],
```
jacobtomlinson: Why has this indentation been removed here? Ranking should only happen on `Message` events.

opsdroid/connector/slack/events.py
jacobtomlinson: Also many of these actions are given a `response_url` which should be used when responding to the action. We may need to handle this in the connector.

```python
from opsdroid.matchers import match_event
from opsdroid.connector.slack.events import MessageAction

class InteractionsSkill(Skill):
    @match_event(MessageAction)
    async def slack_interactions(self, event):
        await event.respond("Message Action interactivity has been triggered.")  # What happens here?
```

docs/connectors/slack.md
daniccan: Done.

opsdroid/connector/slack/events.py
daniccan: Superclass added with `respond` method.

opsdroid/core.py
daniccan: Without this change, the `match_event` matcher is not working. Not sure if I'm missing something else.

daniccan:
@jacobtomlinson Are the changes related to `match_event` API where we create entities also needed as part of this PR?

opsdroid/connector/slack/events.py
```diff
@@ -1,6 +1,7 @@
 """Classes to describe different kinds of Slack specific event."""
 
 import json
+import requests
```
jacobtomlinson: Requests does not currently support asyncio. Therefore your calls to `requests.post` below will block the thread. Could you please use `aiohttp` instead?

opsdroid/connector/slack/events.py
```diff
@@ -42,3 +43,57 @@ def __init__(self, blocks, *args, **kwargs):
...
+
+class InteractiveAction(Event):
+    """Super class to represent Slack interactive actions."""
+
+    def __init__(self, payload, *args, **kwargs):
+        """Create object with minimum properties."""
+        super().__init__(*args, **kwargs)
+        self.payload = payload
+
+    async def respond(self, message):
```
jacobtomlinson: Users should be able to respond with any event type and if it is not supported it should handle that with a warning in the log.

You could check if `message` is a string here and if then call `super().respond`.

opsdroid/core.py
jacobtomlinson: Hmm it looks like `match_event` is being included in the NLU ranking, which it shouldn't. However this change breaks all the NLU ranking. I'll look at raising a PR to fix this.

daniccan:
> > Are the changes related to match_event API where we create entities also needed as part of this PR?
> 
> Yes I think so. Is that ok?

Sure. Just need clarification regarding the following,

```
Also I noticed when writing the above snippet that the block_actions payload contains a list of actions. This may mean multiple actions are bundled up into one webhook call. Therefore we may want to create a BlockAction event instead (just for one action) and parse an event for each action in the list.
```

Could you describe this a little more? Do I have to iterate the list of actions and create a BlockAction event for each action and parse it?

jacobtomlinson:
> Do I have to iterate the list of actions and create a BlockAction event for each action and parse it?

Yes please.

So if your payload looks like this I would like you to iterate over `payload["actions"]` and create an event for each one. You'll still need to take the user and channel (target) information from the top level payload but then the value and other info would be from each action.

```
payload = {
   "type": "block_actions",
   "team": {"id": "T9TK3CUKW", "domain": "example"},
   "user": {
         "id": "UA8RXUSPL",
         "username": "jtorrance",
         "team_id": "T9TK3CUKW",
   },
   "channel": {"id": "CBR2V3XEX", "name": "review-updates"},
   "actions": [
         {
            "action_id": "WaXA",
            "block_id": "=qXel",
            "text": {"type": "plain_text", "text": "View", "emoji": True},
            "value": "click_me_123",
            "type": "button",
            "action_ts": "1548426417.840180",
         }
   ],
}
```

daniccan:
How do I add the `value` to the `match_event` API in `matchers.py` file?

opsdroid/connector/slack/events.py
daniccan: Done.

opsdroid/connector/slack/events.py
daniccan: Done.

jacobtomlinson:
I would go for something like this.

```python
def match_event(event_type, **kwargs):
    """Return event type matcher.
    Decorator that calls skill based on passed event_type.
    Args:
        event_type (str): opsdroidstarted, message, typing, reaction, file, image
        **kwargs (dict): arbitrary kwargs to be added to the event matcher
    Returns:
        Decorated function
    """

    def matcher(func):
        """Add decorated function to list for event matching."""
        func = add_skill_attributes(func)
        func.matchers.append({"event_type": dict(type=event_type, **kwargs)})
        return func

    return matcher
```

daniccan:
Completed the `match_event` matcher changes.

jacobtomlinson:
You've done the first half as I suggested. So the skill will now have an `event_type` matcher dict attached with your kwargs in. But now we also need to check for this and filter when we parse the event matchers. https://github.com/opsdroid/opsdroid/blob/master/opsdroid/parsers/event_type.py#L32

jacobtomlinson:
#1286 should fix the indentation issue you were having.

daniccan:
Made the filter change. Is the change ok?

daniccan:
Can you merge #1286 in master and I'll sync my PR with the master.

opsdroid/core.py
jacobtomlinson: Will be fixed in #1286

opsdroid/parsers/event_type.py
```diff
@@ -8,34 +8,49 @@
 _LOGGER = logging.getLogger(__name__)
 
 
+async def match_event(event, event_opts):
+    event_type = event_opts.get("type", None)
+    event_entitites = event.entities.get("value", {})
```
jacobtomlinson: Do you need to do this here?

opsdroid/parsers/event_type.py
```diff
@@ -8,34 +8,49 @@
...
+                    " event representation.".format(event_type=event_type)
+                )
+                event_type = et
+
+        # TODO: Add option to match all subclasses as well
+        # if isinstance(event, event_type):
+        # pylint: disable=unidiomatic-typecheck
+        if type(event) is event_type:
+            event_value = event_opts.get("value", None)
+            entity_value = event_entitites.get("value", None)
```
jacobtomlinson: This feels a bit rigid here, you're only implementing the `"value"` option for your specific use case.

It would probably be better to iterate over the items in `event_opts` (except `type`) and check if the event has that entitiy and if the values match.

opsdroid/parsers/event_type.py
daniccan: Changed the position.

opsdroid/parsers/event_type.py
daniccan: Changes done.

opsdroid/parsers/event_type.py
```diff
@@ -8,34 +8,53 @@
...
+                    "{event_type} is not a valid opsdroid"
+                    " event representation.".format(event_type=event_type)
+                )
+            event_type = et
+
+        # TODO: Add option to match all subclasses as well
+        # if isinstance(event, event_type):
+        # pylint: disable=unidiomatic-typecheck
+        if type(event) is event_type:
+            for key in event_opts.keys():
```
jacobtomlinson: You don't need to specify `.keys()`. Iterating on a dict will do this automatically.

```suggestion
            for key in event_opts:
```

opsdroid/core.py
jacobtomlinson: Ok that PR is in now so could you update this with the master branch and restore this indentation.

opsdroid/parsers/event_type.py
daniccan: Change done.

daniccan:
Indentation is back now and tested the changes locally. However, made a line change in `core.py` to get the `match_event` matcher to work.

opsdroid/core.py
```diff
@@ -488,10 +488,10 @@ def _connector_names(self):  # noqa: D401
         """
         self.stats["messages_parsed"] = self.stats["messages_parsed"] + 1
         tasks = []
+        tasks.append(self.eventloop.create_task(parse_event_type(self, event)))
```
jacobtomlinson: Good spot with this. That was my mistake.

The `tasks.append(self.eventloop.create_task(parse_always(self, event)))` line should probably be up here too.

daniccan: Done.

jacobtomlinson:
Thanks for iterating with me on this!

I've tested the block action and message action locally and they seem to work nicely. I'm merging now.

PR #1289, opened 2019-12-03
Add build subcommand to load config without running bot

### LINKED ISSUE #244: Add --load-config flag
Need a flag for installing local skill dependencies when running a dockerized app. Should call the install new requirements.txt from skills loaded by `path` without running the bot

FabioRosado:
# Description

This is the initial work on the build subcommand. I was able to add a skill that contains a `requirement.txt` and the subcommand installed all the necessary things.

I've also added a `-d` flag to set the logging level to debug - not sure if this should always be active but it might be good to have the logs showing when running the subcommand so the user knows what's happening.

When I run the command an exception is raised:

```shell
py opsdroid config build
No databases in configuration. This will cause skills which store things in memory to lose data when opsdroid is restarted.
ERROR: Unhandled exception in opsdroid, exiting...
ERROR: Unhandled exception in opsdroid, exiting...
sys:1: RuntimeWarning: coroutine 'BaseEventLoop._create_server_getaddrinfo' was never awaited
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

After running the command in debug mode I can see that the logging message saying that the web server started just after exiting.  I'm not sure how to deal with the issue since this is an synchronous function.

Perhaps we should just run the Loader on the config? 🤔 

Fixes #244

## Status

## Type of change

# How Has This Been Tested?

- Ran the command `opsdroid config build` 
- Ran the command `opsdroid config build -d`
- Ran the command  `opsdroid config -f <path> build`
- Tox - green

# Checklist:

### DIFF
--- a/opsdroid/cli/config.py
+++ b/opsdroid/cli/config.py
@@ -3,11 +3,12 @@
 import click
 
 from opsdroid.cli.utils import (
+    build_config,
     edit_files,
-    warn_deprecated_cli_option,
-    validate_config,
     list_all_modules,
     path_option,
+    validate_config,
+    warn_deprecated_cli_option,
 )
 from opsdroid.const import EXAMPLE_CONFIG_FILE
 
@@ -125,3 +126,31 @@ def list_modules(ctx):
 
     """
     list_all_modules(ctx, ctx.obj, "config")
+
+
+@config.command()
+@click.pass_context
+@click.option(
+    "-d",
+    "verbose",
+    is_flag=True,
+    help="Turns logging level to debug to see all logging messages.",
+)
+def build(ctx, verbose):
+    """Load configuration, load modules and install dependencies.
+
+    This function loads the configuration and install all necessary
+    dependencies defined on a `requirements.txt` file inside the module.
+    If the flag `-d` is passed the logging level will be set as debug and
+    all logs will be shown to the user.
+
+
+    Args:
+        ctx (:obj:`click.Context`): The current click cli context.
+        verbose (bool): set the logging level to debug.
+
+    Returns:
+        int: the exit code. Always returns 0 in this case.
+
+    """
+    build_config(ctx, {"path": ctx.obj, "verbose": verbose}, "config")
--- a/opsdroid/cli/config.py
+++ b/opsdroid/cli/config.py
@@ -131,7 +131,7 @@ def list_modules(ctx):
 @config.command()
 @click.pass_context
 @click.option(
-    "-d",
+    "--verbose",
     "verbose",
     is_flag=True,
     help="Turns logging level to debug to see all logging messages.",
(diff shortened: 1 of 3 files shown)


opsdroid/cli/config.py
```diff
@@ -125,3 +126,31 @@ def list_modules(ctx):
 
     """
     list_all_modules(ctx, ctx.obj, "config")
+
+
+@config.command()
+@click.pass_context
+@click.option(
+    "-d",
+    "verbose",
```
jacobtomlinson: I feel like this flag should either be `--debug` or `--verbose`. Mixing the two is a little strange. I'd also like to reserve `-d` in case there is something better to do with it.

I also agree this should be available on all commands.

opsdroid/cli/utils.py
```diff
@@ -226,3 +228,41 @@ def list_all_modules(ctx, path, value):
...
+    check_dependencies()
+
+    config = load_config_file([path] if path else DEFAULT_CONFIG_LOCATIONS)
+
+    if params["verbose"]:
+        config["logging"] = {"level": "debug"}
+        configure_logging(config)
+
+    with OpsDroid(config=config) as opsdroid:
+        opsdroid.sync_load()
```
jacobtomlinson: You are right that we are running too much here. We could probably just run `opsdroid.loader.load_modules_from_config(config)`.

opsdroid/cli/config.py
FabioRosado: I've decided to go with `--verbose` here  because it seems a bit more logical since the user will get all the logs and know what's happening. If this flag is not set then the only thing shown to the user will be:

```
Opsdroid will build modules from config.
No databases in configuration. This will cause skills which store things in memory to lose data when opsdroid is restarted.
Opsdroid modules successfully built from config.
```

I'm not sure if it's worth to add a try/except and only show the successful message if nothing wrong happened 🤔

FabioRosado:
Using the loader works and opsdroid is able to build the whole config without raising any exception (on a valid config). 

I've also updated the docstrings since we build the documentation from them 😄

opsdroid/cli/config.py
jacobtomlinson: > I'm not sure if it's worth to add a try/except and only show the successful message if nothing wrong happened 🤔

Yes please!

FabioRosado: We are handling exceptions pretty well so when we try to run the subcommand no exception is raised since we are handling it elsewhere. So the solution I found was to suppress any exception - this will make the success message only show when things go right. 

When things go wrong opsdroid will just show the appropriate log message telling that there was a problem. As an example, if the config fails the validation no exception is raised, the success message doesn't show and the validation error message is shown in the terminal.

PR #1292, opened 2019-12-05
Fix logs command - open logs if given different name than default

FabioRosado:
# Description

As we talked in matrix, it could be good to have a command to clear all the logs collected from opsdroid. Since click allows you to ask for confirmation I thought it could be a good idea to include it in this subcommand. 

I've also had to split the logs into `view` since I was unable to use the command `logs clear`. 

## Status

## Type of change

# How Has This Been Tested?

- Tox - all green coverage 100%
- Ran opsdroid logs view - opens logs with vim
- Ran opsdroid logs clear - selected no - aborted message and logs untouched.
- Ran opsdroid logs clear - selected yes - success message and logs cleared.

# Checklist:

- [x] I have made corresponding changes to the documentation (if applicable) - docstring

### DIFF
--- a/opsdroid/cli/utils.py
+++ b/opsdroid/cli/utils.py
@@ -10,14 +10,16 @@ import sys
 import time
 import warnings
 
+from appdirs import user_log_dir
 from opsdroid.core import OpsDroid
 from opsdroid.configuration import load_config_file
 from opsdroid.const import (
     DEFAULT_LOG_FILENAME,
-    LOCALE_DIR,
     DEFAULT_LANGUAGE,
     DEFAULT_CONFIG_PATH,
     DEFAULT_CONFIG_LOCATIONS,
+    LOCALE_DIR,
+    NAME,
 )
 from opsdroid.helper import get_config_option
 from opsdroid.loader import Loader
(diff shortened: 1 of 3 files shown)


opsdroid/cli/utils.py
```diff
@@ -270,3 +270,31 @@ def build_config(ctx, params, value):
...
+            calling this function.
+
+    """
+    click.echo(click.style("WARNING:", bg="yellow", bold=True), nl=False)
+    click.confirm(
+        " Are you sure you wish to clear all your logs? This action cannot be undone.",
+        abort=True,
+    )
+
+    open(DEFAULT_LOG_FILENAME, "w").close()
```
jacobtomlinson: I think the logs location is configurable.

FabioRosado: Good point I was looking at the edit cli function. This made me realise that the command would only work if the log was using the default location as well. 

With this in mind should we handle the case where there are more than one log file in the logs folder? 🤔

jacobtomlinson: Maybe rather than adding these CLI options it would be better for opsdroid to rotate its own logs?

http://www.blog.pythonlibrary.org/2014/02/11/python-how-to-create-rotating-logs/

FabioRosado: Yeah that will be a better option, I'm going to start working on that. But since the edit cli doesn't work on logs that don't use the default name I'll use this PR to fix that and then raise a new one adding rotation 😄

PR #1294, opened 2019-12-07
Remove deprecated flags and add -f flag to logs

FabioRosado:
# Description

This PR removes the deprecated flags from the cli, it also adds two new subcommands to the logs command - head and tail.

The head command will print the first 20 lines - this can be changed if passed like `opsdroid logs head 100`

The tail will print the last 20 ones. I've also added the `-f` flag to the tail subcommand to follow the logs in realtime.

I've used the tailer dependency since it seem a good way to deal with it.

I know that we talked in matrix about settings just the -f flag on the logs but perhaps it could be useful to get either the first or last lines of the logs instead of getting them in real time only.

## Status

## Type of change

- Documentation (fix or adds documentation) - docstrings

# How Has This Been Tested?

- Tox - green and coverage ok
- Ran opsdroid logs - print all the logs
- Ran opsdroid logs head - shows either 20 lines or more if int is passed
- Ran opsdroid logs tail - shows either 20 lines or more if int is passed
- Ran opsdroid and opsdroid logs tail -f - got logs in real time.

# Checklist:

### DIFF
--- a/opsdroid/cli/__init__.py
+++ b/opsdroid/cli/__init__.py
@@ -2,66 +2,20 @@
 
 import click
 
-from opsdroid.cli.config import config, print_example_config
+from opsdroid.cli.config import config
 from opsdroid.cli.logs import logs
 from opsdroid.cli.start import start
-from opsdroid.cli.version import version, print_version
-from opsdroid.cli.utils import edit_files, warn_deprecated_cli_option
+from opsdroid.cli.version import version
 
 
-@click.group(invoke_without_command=True)
+@click.group()
 @click.pass_context
-@click.option(
-    "--gen-config",
-    is_flag=True,
-    callback=print_example_config,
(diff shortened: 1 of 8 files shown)


jacobtomlinson:
Thanks for this, glad to see the `-f` flag and the old flags gone.

I'm worried with the `head` and `tail` commands that we are just reproducing standard bash things in a non-bash way and that people wouldn't use it.

As someone who uses bash a lot I would generally run `opsdroid logs | head -n 100` if I wanted the head of the logs. I wouldn't even condifer checking if opsdroid has that functionality built in.

FabioRosado:
I see your point and agree with you, I think I just went a bit crazy with creating stuff with click haha

The next update of this PR I will drop the head/tail subcommands and include the `-f` flag to the logs like you suggested 😄

PR #1295, opened 2019-12-08
Authorization example

jcomeaux:
# Description

Provides a complete though simple example of user authorization

### DIFF
--- a/docs/skills/constraints.md
+++ b/docs/skills/constraints.md
@@ -23,6 +23,35 @@ class MySkill(Skill):
         await message.respond('Hey')
 ```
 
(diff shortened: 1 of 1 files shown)


FabioRosado:
Hello Joel thanks for raising this PR, I think Jacob was suggesting if you could write an example on how to deal with authentication on the example part of the docs - https://github.com/opsdroid/opsdroid/tree/master/docs/examples

jacobtomlinson:
Yeah thanks for raising this @jcomeaux but I was kinda hoping you'd write this up as a standalong example rather than update the constraints docs.

Would you be up for doing this? I think it would be a really useful example.

jcomeaux:
Yep, no problem!  I'll just close this PR out and send another...

PR #1296, opened 2019-12-09
stand alone authorization example

jcomeaux:
# Description

Adds standalone authorization example of supreme awesomeniss!

### DIFF
--- a/docs/examples/authorization.md
+++ b/docs/examples/authorization.md
@@ -1,0 +1,32 @@
+# Authorization
+
+Combining [constrain_users](https://docs.opsdroid.dev/en/stable/skills/constraints.html#constrain-users) and a simple user list, authorization for a particular skill is straightforward.
+
+In this example, we use an AWS IAM group for authorization. Note, you would need to reload should the group members change.
+
+```python
+import boto3
+from opsdroid.skill import Skill
+from opsdroid.matchers import match_regex
+from opsdroid.constraints import constrain_users
+
+def _get_admins():
+    admins = []
+    try:
+        groupname = "opsdroid_admins"
+        client = boto3.client("iam")
+        group = client.get_group(GroupName=groupname)
+        admins = [user.get("UserName") for user in group.get("Users")]
+    except Exception as e:
+        logging.exception("unable to get admin group members!")
+    return admins
+
+admins = _get_admins()
+
+class AdminSkill(Skill):
+    @match_regex(r'stop all the things!')
+    @constrain_users(admins)
+    async def stop_everything(self, message):
+        # do admin stuff
+        await message.respond('done!')
+```


jcomeaux:
Okay, i used your example from issue 1198 and that works, however, i have several functions within that skill.  So the code below seems impossibly clunky I guess? As far as loading in credentials to access authorization info like AWS IAM, putting that in a config file isn't best practice.  Using ec2 role permissions or something loaded in via the secrets manager would be the way to go.

I appreciate your patience, but it's beginning to feel like it may make sense to make an auth module a first class extension like skills or connectors? or is that crazy? :)

```
from opsdroid.constraints import constrain_users

class MySkill(Skill):
    def __init__(self, opsdroid, config):
        super(MySkill, self).__init__(opsdroid, config)
        
        # Get your admin users somehow
        users = ['admin']  

        # Dynamically add the constraint
        self.admin_skill_1 = constrain_users(users)(self.admin_skill_1)
        self.admin_skill_2 = constrain_users(users)(self.admin_skill_2)
        self.admin_skill_3 = constrain_users(users)(self.admin_skill_3)

    @match_regex(r"Delete everything")
    async def admin_skill_1(event):
        # Do some admin task
        pass

    @match_regex(r"Do the admin 2 thing")
    async def admin_skill_2(event):
        pass
```

jcomeaux:
@jacobtomlinson curious about your thoughts on my comment above; given that my code is pretty janky though, my feelings wouldn't be hurt if we close this one as not example-worthy 😏

jacobtomlinson:
> So the code below seems impossibly clunky I guess?

I see what you mean. It's not terrible, but it could be better.

I had seen you do an example like this, did it not work?

```
from opsdroid.constraints import constrain_users

class MySkill(Skill):
    admins = []

    def __init__(self, opsdroid, config):
        super(MySkill, self).__init__(opsdroid, config)
        
        # Get your admin users somehow
        self.admins = ['admin']  

    @match_regex(r"Delete everything")
    @constrain_users(admins)
    async def admin_skill_1(event):
        # Do some admin task
        pass

    @match_regex(r"Do the admin 2 thing")
    @constrain_users(admins)
    async def admin_skill_2(event):
        pass
```

> As far as loading in credentials to access authorization info like AWS IAM, putting that in a config file isn't best practice.

While I agree this isn't best practice that doesn't mean it should be impossible. What if my bot isn't running on AWS but I want to use IAM? Other non-AWS authentication may expose credentials via environment variables (12 factor app style), user's may have already included them in their config using these (you can reference env vars in your config).

I'm totally on the same page that our examples should show best practices. But they also need to be simple and understandable to novices who may not know about AWS specific details.

> given that my code is pretty janky though, my feelings wouldn't be hurt if we close this one as not example-worthy

We are well on the way to getting a good example here! Please don't be discouraged by the review process.

jcomeaux:
No, not frustrated! Happy to try and help out.  Gotcha on the "AWS magic" aspect...I'll work out a couple of approaches for getting the auth info that should work for most people.  

...and no, the code that I tried before that you're referencing does not work I think because, as you say, the `constrain_user` decorator is processed too early.  I'll continue to work on it!

jcomeaux:
Okay, still not frustrated ( 🙂) but am probably at the limits of my abilities with python.  This is this best I could come up with so far.  I guess whether I get the user list in `__init__( )` or do it this way, it's still going to result in an I/O operation when the class is loaded, and this way, at least the `lru_cache` prevents an AWS api hit every time the skill is called.  Still requires a reload if the group members change, but could use something like `cachetools` or similar lib to add cache expire/ttl.

If this is still way off of the right path I'll update the example and include several methods of configuring boto3...otherwise, I'm at a bit of a loss here.  

```python
from functools import lru_cache
import logging
import json

import boto3
from opsdroid.matchers import match_witai
from opsdroid.constraints import constrain_users

@lru_cache()
def get_admins():
    admins = []
    try:
        groupname = "aws_iam_admin_group"
        client = boto3.client("iam")
        group = client.get_group(GroupName=groupname)
        admins = [user.get("UserName") for user in group.get("Users")]
    except Exception as e:
        logging.exception("unable to get admin group members!")
    return admins

class AdminSkill(Skill):
    @constrain_users(get_admins())
    @match_witai("admin_job")
    async def admin_job(self, opsdroid, config, message):
        # do admin_job work
        response.append("admin job is done!")
        await message.respond("\n".join(sorted(response)))
```

PR #1311, opened 2019-12-21
fix(logging): UnboundLocalError when config['logging']['extended'] is…

snahor:
# Description

If config['logging']['extended'] == False, we were getting an UnboundLocalError

Set formatter before attempting to get the value from the configuration, on KeyError just ignore.

Fixes #<issue>

## Status
**READY**

## Type of change

# How Has This Been Tested?

Manual test, set extended to True/False, also removed it from the config file.

# Checklist:

### DIFF
--- a/opsdroid/logging.py
+++ b/opsdroid/logging.py
@@ -80,13 +80,14 @@ def configure_logging(config):
 
     rootlogger.setLevel(log_level)
 
+    formatter = logging.Formatter("%(levelname)s %(name)s: %(message)s")
     try:
         if config["logging"]["extended"]:
             formatter = logging.Formatter(
                 "%(levelname)s %(name)s.%(funcName)s(): %(message)s"
             )
     except KeyError:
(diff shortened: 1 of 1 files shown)


snahor:
The CI seems to be failing for a style error, I will fix it later.

opsdroid/logging.py
```diff
@@ -80,13 +80,14 @@ def configure_logging(config):
 
     rootlogger.setLevel(log_level)
 
+    formatter = logging.Formatter("%(levelname)s %(name)s: %(message)s")
     try:
```
FabioRosado: I'd suggest this to be changed to something like this to avoid using the try/except: pass.

```suggestion
    with contextlib.suppress(KeyError):
        if config["logging"]["extended"]:
             formatter = logging.Formatter(
                 "%(levelname)s %(name)s.%(funcName)s(): %(message)s"
             )
```

With contextlib.suppress, the KeyError gets suppressed and will automatically pass.

PR #1336, opened 2020-01-11
Allow mxc URLs in matrix messages

Cadair:
This is a quick bugfix which allows you to embed images in html messages when using matrix. I think this is a new feature of bleach.


SolarDrew:
LGTM :+1:

PR #1339, opened 2020-01-17
Add logs rotation

FabioRosado:
# Description

I've been meant to add rotation to logs for a while now, finally I've sat down and got working on it. This is a pretty simple change but works as expected. 

I've decided on 50kb size +/- mostly because it should be the size of 1000 lines. I'd be happy to change this if you prefer something else 😄 

We could add a timed rotation as well but I'm not sure if that should be added or not what do you think?

Fixes #<issue>

## Status

## Type of change

# How Has This Been Tested?

- Tox - all green
- Ran opsdroid and watch logs being rotated

# Checklist:

### DIFF
--- a/opsdroid/logging.py
+++ b/opsdroid/logging.py
@@ -4,6 +4,7 @@ import os
 import logging
 import contextlib
 
+from logging.handlers import RotatingFileHandler
 from opsdroid.const import DEFAULT_LOG_FILENAME, __version__
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


opsdroid/logging.py
```diff
@@ -112,6 +113,8 @@ def configure_logging(config):
         with contextlib.suppress(KeyError):
             file_handler.addFilter(ParsingFilter(config, config["logging"]["filter"]))
 
+        rotation_handler = RotatingFileHandler(logfile_path, maxBytes=config["logging"].get('file-size', 50))
```
jacobtomlinson: Looks like `RotatingFileHandler` takes bytes not KB. So this will rotate at 50 bytes.

Let's bump this up, I recommend 50MB (`50e6`) would probably be a good default.

FabioRosado: Uh that was a fail haha I'll change that to 50mb

FabioRosado:
This PR turn out a bit of a mess apologies for that.

jacobtomlinson:
No worries, thanks for this!

Issue #1347, opened 2020-01-30
Skill configuration issues


quulah:
# Description
I'm working on a skill which should have the ability to pull the room and connector from it's config. This issue is partly a feature request, and maybe bit of a bug report or at least the documentation may need some additions.

## Steps to Reproduce
Skill `__init__()`:
```
class CVESkill(Skill):
    def __init__(self, opsdroid, config):
        super(CVESkill, self).__init__(opsdroid, config)
        ...
        # Dynamically add a decorator based on config
        # See: https://github.com/opsdroid/opsdroid/issues/1198
        expression = self.config.get("cron", {}).get("expression", "15 * * * *")
        self.schedule = match_crontab(expression)(self.schedule)
```

```
@match_regex("^cve$", case_sensitive=False)
async def schedule(self, message):
...
    if not message:
        if self.config.get("cron", {}).get("connector"):
            # Waiting for: https://github.com/opsdroid/opsdroid/issues/1345
            [connector] = [
                x
                for x in self.opsdroid.connectors
                if x.name == self.config.get("cron", {}).get("connector")
            ]
        else:
            connector = self.opsdroid.default_connector

        room = self.config.get("cron", {}).get(
            "room", self.opsdroid.default_connector.default_target
        )

        message = Message("", None, room, connector)

    await message.respond(f"Searching CVEs with queries: {', '.join(queries)}")
...
```

## Expected Functionality
The above should work and push the message from the cron trigger to #somethingelse. When triggered by hand with the regex match, it should write to the room being asked from.

The feature request part of this issue would be out-of-the-box configurability of the connector and room per skill. Also, it would be nice to have some sort of default configuration along the lines of `CONFIG_SCHEMA`. I'm currently using the `self.config.get()` and setting some sane values.

It's also a bit tricky to configure the decorators, as you need to do it like #1198 states. Perhaps it would be worthwhile to add this in the documentation.

The bug report / cry for help is that I originally had the room / connector selection in the `__init__()` of the skill, but this just causes the whole bot to hang when starting. I'd like to do it that way to have all the configuration handling in one place. As seen below, there's also something wrong with the current approach as the connector ends up being `None` for some reason.

## Experienced Functionality

When the room / connector selection is in the `schedule()` method:
```
hemmo_1  | DEBUG opsdroid.parsers.crontab: Running crontab skills at Thu Jan 30 08:29:00 2020.
hemmo_1  | ERROR opsdroid.core: Exception when running skill 'cve'.
[...]

PR #1348, opened 2020-01-30
Update validated config with defaults

jacobtomlinson:
# Description

When setting your `CONFIG_SCHEMA` with voluptuous it is possible to set default values for the configuration. However we are not actually capturing the output of the `validate` command which returns the validated config with the defaults applied.

This PR returns the validated config and updates it so you can set default config values in your skills like this:

```python
from voluptuous import Required

CONFIG_SCHEMA = {Required("foo", default="bar"): str}
```

Related to #1347

## Status
**READY** 

## Type of change

# How Has This Been Tested?

I've added a new test which loads in a skill that has a default configuration option set and asserts that value is correct in the `config` object.

# Checklist:

### DIFF
--- a/docs/skills/index.md
+++ b/docs/skills/index.md
@@ -41,6 +41,18 @@ skills:
 
 For more information about the various ways you can package skills and other modules and tell opsdroid about them see the [packaging section](../packaging).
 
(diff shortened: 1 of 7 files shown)


FabioRosado:
Well spotted, I remember I used default values when looking at voluptuous but then didn't included them for some reason 🤷‍♂️

PR #1355, opened 2020-02-04
update telegram connector for message event changes

### LINKED ISSUE #1354: Telegram connector needs update for event message
# Description
The telegram connector uses message.target['id'] instead of message.target.  This leads to problems when trying to use default_target per normal configuration and usage of the event flow logic.

## Steps to Reproduce
Setup telegram as the default connector... call the telegram connector.send with a target...  
```
await self.opsdroid.send(Message(text='hello', target='<useridhere>') 
```
The default_target is also always None.
```
await self.opsdroid.send(Message(text='hello')  
```
You can hack around it with..
```
sillytarget = { 'id': <useridhere> }
await self.opsdroid.send(Message(text='hello', target=sillytarget) 
```
## Expected Functionality
message.target should work like the other core connectors

## Experienced Functionality
Errors out.
```
opsdroid    | ERROR opsdroid.connector.telegram.send_message(): Unable to respond.   
```
## Versions

- opsdroid: latest/stable
- python 3.7.6
- docker image: opsdroid/opsdroid:latest

## Configuration File
```yaml
connectors:
## Telegram (core)
  telegram:
    token: "......"
   # optional
    update-interval: 0.5  # Interval between checking for messages
    whitelisted-users:  # List of users who can speak to the bot, if not set anyone can speak
      - ......
```
## Additional Details
None

TonyRoomZ:
# Description
Fixes #1354 

## Status
**READY**

## Type of change

# How Has This Been Tested?
- tox
- test against hello skill for regression
- created a webhook skill per documentation example to verify default_target in Message event now works using the current telegram configuration example

# Checklist:

I have not written a specfic test for this case, If I do, let me know.

### DIFF
--- a/opsdroid/connector/telegram/__init__.py
+++ b/opsdroid/connector/telegram/__init__.py
@@ -167,9 +167,13 @@ class ConnectorTelegram(Connector):
         """
         for result in response["result"]:
             _LOGGER.debug(result)
+
             if result.get("edited_message", None):
                 result["message"] = result.pop("edited_message")
-            if "channel" in result["message"]["chat"]["type"]:
+            if result.get("channel_post", None) or result.get(
+                "edited_channel_post", None
+            ):
+                self.latest_update = result["update_id"] + 1
                 _LOGGER.debug(
                     _("Channel message parsing not supported " "- Ignoring message.")
                 )
--- a/opsdroid/connector/telegram/__init__.py
+++ b/opsdroid/connector/telegram/__init__.py
@@ -34,9 +34,9 @@ class ConnectorTelegram(Connector):
         self.name = "telegram"
         self.opsdroid = opsdroid
         self.latest_update = None
-        self.default_target = None
         self.listening = True
         self.default_user = config.get("default-user", None)
+        self.default_target = self.default_user
         self.whitelisted_users = config.get("whitelisted-users", None)
         self.update_interval = config.get("update-interval", 1)
         self.session = None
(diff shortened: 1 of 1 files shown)


TonyRoomZ:
I went ahead and tested the change on a channel.  Turns out the ``result['message']'[chat']['id']``  isn't the best of checks is we want to ignore them, so I went ahead and fixed that...  I am thinking it might be a good idea to look at using a well supported telegram library as a new telegram connector basis.  What we have now works, but we might want to be able to use more of its platform features later.  I might take a stab at that later, but for now, this is fixed :smile:

jacobtomlinson:
Thanks @TonyRoomZ! 

The only issue we often have with using third party libraries is the lack of asyncio support. We've used many libraries in the past which are not well supported and have been bitten a few too many times. Therefore our default position is to manually interact with the API unless the service has a first party library that supports asyncio or an extremely popular third party library. 

This was more of an issue when we started as asyncio support was pretty new. Many first party libraries now support it. So if there is one for Telegram then we can definitely discuss it. But I recommend you open an issue first to discuss it before going to the effort of raising a PR. Hope that's ok!

I'm going to leave this PR for @FabioRosado to merge as he mostly looks after the Telegram connector, but it looks good to me!

TonyRoomZ:
@jacobtomlinson I understand the lack of asyncio support.. In this case I was specifically pondering integration with [telethon](https://github.com/LonamiWebs/Telethon) which is written using asyncio.  At least I will take a stab at it for a couple days.  I started a branch for it.... couldn't hurt to try :grinning:

jacobtomlinson:
Telethon looks well maintained and popular, so I would be more than happy to depend on it. Go for it!

PR #1369, opened 2020-02-16
Create event for users joining a room in Matrix

awesome-michael:
# Description
Listens to matrix state events and creates a JoinRoom event when a `m.room.member` event with content `"membership": "join"` occurs.

## Status
**READY**

## Type of change

# How Has This Been Tested?
Tested with this skill:
```
@match_event(JoinRoom)
    async def hello_join(self, event):
        text = random.choice(
            ["Welcome {}", "Hey {}, nice to meet you"]).format(event.user)
        await event.respond(Message(text))
```

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -108,11 +108,7 @@ class ConnectorMatrix(Connector):
             "event_format": "client",
             "account_data": {"limit": 0, "types": []},
             "presence": {"limit": 0, "types": []},
-            "room": {
-                "account_data": {"types": []},
-                "ephemeral": {"types": []},
-                "state": {"types": []},
-            },
+            "room": {"account_data": {"types": []}, "ephemeral": {"types": []}},
         }
 
     async def make_filter(self, api):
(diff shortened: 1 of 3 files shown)


opsdroid/events.py
```diff
@@ -451,6 +451,9 @@ def __init__(self, description, *args, **kwargs):  # noqa: D107
 class JoinRoom(Event):
     """Event class to represent a user joining a room."""
 
+    def __init__(self, *args, **kwargs):  # noqa: D107
```
Cadair: Do we need an `__init__` here if it's not doing anything?

awesome-michael: Was not sure if its need to call the `super.init`. As no additional parameters are needed.. what would be the correct constructor?

awesome-michael:
> Looks good, it would be great to add a test for this, especially as it's a new field in the sync. Do you have the time to do that?

Added a test. Could you please review it?

opsdroid/events.py
jacobtomlinson: If you omit the `__init__` it will call the super init so don't worry. You can remove this.

Cadair:
Thanks a lot @awesome-michael :rocket:

PR #1375, opened 2020-02-22
Add matcher that fires when bot is mentioned

awesome-michael:
# Description
Adds a matcher that listens for messages in that the bot is mentioned. Currently only for matrix connector. I would add slack mentions if I knew how to get the bots slack intern user id.

## Status
**READY**

## Type of change

# Checklist:

### DIFF
--- a/opsdroid/parsers/mention.py
+++ b/opsdroid/parsers/mention.py
@@ -1,13 +1,24 @@
 """A helper function for parsing and executing mention skills."""
 
 import logging
+import re
 
 
 _LOGGER = logging.getLogger(__name__)
 
 
 async def parse_mention(opsdroid, skills, message):
-    """Parse a message if the bot is mentioned."""
+    """Parse a message if the bot is mentioned.
+
+    Args:
+        opsdroid (OpsDroid): An instance of opsdroid.core.
+        skills (list): A list containing all skills available.
+        message(object): An instance of events.message.
+
+    Return:
+        Either empty list or a list containing all matched skills.
+    """
+
     matched_skills = []
     mentions = {}
 
@@ -19,11 +30,9 @@ async def parse_mention(opsdroid, skills, message):
         for matcher in skill.matchers:
             if "mention" in matcher:
                 if "matrix" in mentions:
-                    if "formatted_body" in message.raw_event["content"] and (
-                        '<a href="https://matrix.to/#/' + mentions["matrix"] + '">'
-                        in message.raw_event["content"]["formatted_body"]
-                        or "<a href='https://matrix.to/#/" + mentions["matrix"] + "'>"
-                        in message.raw_event["content"]["formatted_body"]
+                    if "formatted_body" in message.raw_event["content"] and re.search(
+                        "https://matrix.to/#/" + mentions["matrix"],
+                        message.raw_event["content"]["formatted_body"],
                     ):
                         matched_skills.append(
                             {
--- a/opsdroid/parsers/mention.py
+++ b/opsdroid/parsers/mention.py
@@ -1,0 +1,37 @@
+"""A helper function for parsing and executing mention skills."""
+
+import logging
+
+
+_LOGGER = logging.getLogger(__name__)
+
+
+async def parse_mention(opsdroid, skills, message):
+    """Parse a message if the bot is mentioned."""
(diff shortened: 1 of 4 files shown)


opsdroid/parsers/mention.py
```diff
@@ -0,0 +1,37 @@
+"""A helper function for parsing and executing mention skills."""
+
+import logging
+
+
+_LOGGER = logging.getLogger(__name__)
+
+
+async def parse_mention(opsdroid, skills, message):
+    """Parse a message if the bot is mentioned."""
```
FabioRosado: Could we turn this into  a google style docstring?

opsdroid/parsers/mention.py
```diff
@@ -0,0 +1,37 @@
...
+    for connector in opsdroid.connectors:
+        if connector.config["name"] == "matrix":
+            mentions.update({"matrix": connector.config["mxid"]})
+
+    for skill in skills:
+        for matcher in skill.matchers:
+            if "mention" in matcher:
+                if "matrix" in mentions:
+                    if "formatted_body" in message.raw_event["content"] and (
+                        '<a href="https://matrix.to/#/' + mentions["matrix"] + '">'
```
FabioRosado: Am I wrong to think that the difference between these two is that one uses double quotes after `href=` and the other uses a single quote?

A way to simplify this could be by using format or fstrings. This could be turn into 
`'<a href="https://matrix.to/#/{mentions}">'.format(mentions=mentions["matrix"])`.

Or... use regex to check if `message.raw_event["content"]["formatted_body"]` matches any of `https://matrix.to/#/ mentions["matrix"]`.

Either way it should work, I'd go with regex just to avoid the same check just with different quotes.

opsdroid/parsers/mention.py
awesome-michael: Could you give an example? Unfortunately I don't know the definition of a google style docstring

opsdroid/parsers/mention.py
awesome-michael: You're right. Was a last-minute patch because I noticed the test uses single quotes. Think I'll go for a regex

opsdroid/parsers/mention.py
FabioRosado: The dialogflow parser has an example of google style docstring on [this line] (https://github.com/opsdroid/opsdroid/blob/7282a5da065cbe73212245f55b7cc4b3537ebd37/opsdroid/parsers/dialogflow.py#L13)

I've added some explanation but we just need to add this:
```
    Args:
        opsdroid (OpsDroid): An instance of opsdroid.core.
        skills (list): A list containing all skills available.
        message(object): An instance of events.message.
        config (dict): configuration settings from the
            file config.yaml.
    Return:
        Either empty list or a list containing all matched skills.
```

Cadair:
I haven't had a chance to look at this properly, but how about having a method on connectors which detects mentions for itself in events? This could be called by the matcher, or even just run all the time and emit events with a specific type?

FabioRosado:
Hello @awesome-michael I've reopened this PR, any chance that you have some time to check the comments and see if you have the time to update the PR? I'm keen on merging this PR 😄 👍

awesome-michael:
Hey,
added the suggested regex a while ago, the docstring may be missing. I waited on the result of your internal discussion about @Cadair 's comment if this should be a matcher or a function of the connectors.

FabioRosado:
Fair enough I think I've missed this so apologies for that! I think it could be a great thing to perhaps run all the time and send the event when the bot is mentioned 🤔

Cadair:
I think that's the best idea in theory. (An `OpsdroidMention` event which has `linked_event` set to the event that contains the mention.) The downside of this is that you could potentially get a skill triggered on the `Message` event as well as the `OpsdroidMention` event.

Another way to think about mentions (that just came into my brain) is that Events contain them, i.e. a `Message` event could contain five `Mention` events of various different users that were mentioned in that event. This seems like it would scale better into detecting all mentions not just mentions of the bot, and also potentially scale into other things, like links to other rooms / channels / groups etc.

To do that we would need to extend the base event class with some kind of `contains` attribute and then have a matcher which can match on the contents of contains (also maybe a specific version for bot mentions to make it easier).

PR #1396, opened 2020-03-12
Add option to use access token instead of mxid+password for matrix connector

### LINKED ISSUE #813: support providing an access token in the configuration for the matrix connector rather than the password
This would alleviate the need to plain text the password in the config, and the access token can be deactivated. This would be a case of detecting the key in the config and skipping the login call

Tyagdit:
# Description

The matrix connector needed the user's password in plaintext for its configuration. Changed that to allow either the mxid+password or access token.

Fixes #813 

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Tried the default Hello skill with matrix - bot connected successfully and replied as expected.

# Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -542,17 +542,28 @@ Let's take the example of our matrix connector. Inside the module we set the con
 
 ```python
 from voluptuous import Required
-
-CONFIG_SCHEMA = {
-    Required("credentials"): dict,
-    Required("rooms"): dict,
-    "homeserver": str,
-    "nick": str,
-    "room_specific_nicks": bool,
-}
+CONFIG_SCHEMA = Any(
+    {
+        Required("mxid"): str,
+        Required("password"): str,
+        Required("rooms"): dict,
+        "homeserver": str,
+        "nick": str,
+        "room_specific_nicks": bool,
+    },
+    {
+        Required("token"): str,
+        Required("rooms"): dict,
+        "homeserver": str,
+        "nick": str,
+        "room_specific_nicks": bool,
+    },
+)
 ```
 
-As you can see 'credentials` and `rooms` are required fields for this connector and we expect them to be dictionaries.
+As you can see 'mxid`, 'password', 'token' and `rooms` are required fields for this connector and we expect them to be strings and a dictionary.
+
+You can use the Any() method to return one of either dictionaries to provide alternative required fields
 
 Since we don't need to explicitly declare a value as Optional we can just write the expected value and type.
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -544,8 +544,7 @@ Let's take the example of our matrix connector. Inside the module we set the con
 from voluptuous import Required
 
 CONFIG_SCHEMA = {
-    Required("mxid"): str,
-    Required("password"): str,
+    Required("credentials"): dict,
     Required("rooms"): dict,
     "homeserver": str,
     "nick": str,
@@ -553,7 +552,7 @@ CONFIG_SCHEMA = {
 }
 ```
 
-As you can see `mxid`, `password` and `rooms` are required fields for this connector and we expect them to be either strings or a dictionary.
+As you can see 'credentials` and `rooms` are required fields for this connector and we expect them to be dictionaries.
 
 Since we don't need to explicitly declare a value as Optional we can just write the expected value and type.
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -554,14 +554,14 @@ CONFIG_SCHEMA = Any(
     {
         Required("token"): str,
         Required("rooms"): dict,
-        "homeserver": str,
+        Required("homeserver"): str,
         "nick": str,
         "room_specific_nicks": bool,
     },
 )
 ```
 
-As you can see 'mxid`, 'password', 'token' and `rooms` are required fields for this connector and we expect them to be strings and a dictionary.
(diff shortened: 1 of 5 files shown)


Cadair:
We should be allowing either username + password or access token. (You don't actually need to specify username if you specify access token iirc).

Tyagdit:
You don't need mxid to login but we need it for getting the nickname and creating the filter and other stuff and I can't seem to find a way to retrieve the mxid by using just the token.

Cadair:
You can use this APi call: https://matrix.org/docs/spec/client_server/r0.5.0#get-matrix-client-r0-account-whoami

docs/connectors/matrix.md
```diff
@@ -14,8 +14,11 @@ To use this connector you will need to have a Matrix account, and login using yo
 connectors:
   matrix:
     # Required
-    mxid: "@username:matrix.org"
-    token: "mytoken"
+    credentials:
+      'token': 'mytoken'  # Your matrix Access Token
+      # Alternatively, you could use your matrix username and password (not recommended)
```
Cadair:
```suggestion
      # Alternatively, you could use your matrix username and password
```

opsdroid/configuration/example_configuration.yaml
```diff
@@ -134,8 +134,11 @@ connectors:
 #  ## Matrix (core)
 #  matrix:
 #    # Required
-#    mxid: "@username:matrix.org"
-#    token: "mytoken"
+#    credentials:
+#      'token': 'mytoken'  # Your matrix Access Token
+#      # Alternatively, you could use your matrix username and password (not recommended)
```
Cadair:
```suggestion
#      # Alternatively, you could use your matrix username and password
```

docs/connectors/matrix.md
```diff
@@ -14,8 +14,11 @@ To use this connector you will need to have a Matrix account, and login using yo
 connectors:
   matrix:
     # Required
-    mxid: "@username:matrix.org"
-    token: "mytoken"
+    credentials:
```
Cadair: I like this, but I think given it's a breaking change it would be easier to to remove this level of nesting. (So that people who have `mxid` and `password` at the moment don't have to change anything). And we just validate that either `mxid` and `password`, or `token`.

Cadair:
@Tyagdit You likely to pick this back up? :smile:

docs/configuration.md
```diff
@@ -542,17 +542,28 @@ Let's take the example of our matrix connector. Inside the module we set the con
...
+        Required("password"): str,
+        Required("rooms"): dict,
+        "homeserver": str,
+        "nick": str,
+        "room_specific_nicks": bool,
+    },
+    {
+        Required("token"): str,
+        Required("rooms"): dict,
+        "homeserver": str,
```
Cadair: If we are using token rather than mxid, we are actually going to have to require homeserver as we can't do a well-known lookup for it.

docs/connectors/matrix.md
```diff
@@ -6,16 +6,18 @@ Maintained by [@SolarDrew](https://github.com/SolarDrew).
 
 ## Requirements
 
-To use this connector you will need to have a Matrix account, and login using your Matrix username (mxid) and password.
+To use this connector you will need to have a Matrix account, and login using your Matrix username (mxid) and access token. You can find the access token in the **Advanced** section of the **Help & About** tab in the **Settings** of the official <Riot.im> web client.
```
Cadair: This needs rewording. I think we should be recommending username + password first as it's the easiest way.

docs/connectors/matrix.md
```diff
@@ -6,16 +6,18 @@ Maintained by [@SolarDrew](https://github.com/SolarDrew).
...
 ## Configuration
 
 ```yaml
 connectors:
   matrix:
     # Required
-    mxid: "@username:matrix.org"
-    password: "mypassword"
+    'token': 'mytoken'  # Your matrix Access Token
+    # Alternatively, you could use your matrix username and password
```
Cadair: flip this around to be username + password first then token.

opsdroid/connector/matrix/connector.py
```diff
@@ -22,14 +22,23 @@
...
+        Required("password"): str,
+        Required("rooms"): dict,
+        "homeserver": str,
+        "nick": str,
+        "room_specific_nicks": bool,
+    },
+    {
+        Required("token"): str,
+        Required("rooms"): dict,
+        "homeserver": str,
```
Cadair: heh, this is the actual code, homeserver is required with token.

docs/connectors/matrix.md
```diff
@@ -6,25 +6,26 @@ Maintained by [@SolarDrew](https://github.com/SolarDrew).
...
-    # A dictionary of rooms to connect to
-    # One of these have to be named 'main'
+    # Alternatively, you could use your matrix access token
+    'token': 'mytoken'
+    # A dictionary of multiple rooms
+    # One of these should be named 'main'
     rooms:
       'main': '#matrix:matrix.org'
       'other': '#riot:matrix.org'
+    homeserver: "https://matrix.org" # Optional if you've specified mxid and password or if the homeserver is https://matrix.org
```
Cadair: We haven't actually implemented well-known support yet, so it's only optional if the server is matrix.org

Tyagdit: Shoud we make it optional and just explain that it defaults to matrix.org if not specified?

Cadair: tbh I would rather make it always required, but that's a bit of a breaking change. Let's document it as if it were required in the user docs when specifying a username and password, but leave the default for now. For token it should always be required.

PR #1399, opened 2020-03-14
Skill constraints for all event types

### LINKED ISSUE #1381: Constraints don't apply to non-Message events
# Description
It seems that constrains don't apply to non-Message events.

## Steps to Reproduce
```python
class ExampleSkill(Skill):

    @match_event(JoinRoom)
    @constrain_rooms(['room1'])
    async def user_join(self, event):
        print('User joined {}'.format(event.target))
```

## Expected Functionality
The `user_join()` skill method would only run if a user joined `room1`

## Experienced Functionality
The `user_join()` skill method is run when a user joins any room the connector is configured to join.

## Versions
- **Opsdroid version:** 7282a5da06
- **Python version:** 3.7.6
- **OS/Docker version:** Linux/Debian (testing)

## Configuration File
```yaml
connector:
  - name: matrix
    mxid: '@user:matrix.org'
    password: xxx
    nick: 'User'
    rooms:
      'room1': <room id>
      'room2': <room id>
```

thesseyren:
# Description

Fixes #1381 
This code patch is checks all constraints of all skills for the relevant event when `parse_event_type` triggered.

~By the way, `@constrain_rooms` for the skills isn't works with room aliases which are specified in configuration file (because `message.target` and `rooms` doesn't match [here](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/constraints.py#L34), I guess). It works with exact room ids (at least for Matrix connector). First of all, is this behavior a bug or a feature? And if it is a bug, should be fixed in this PR?~

This code patch adds `lookup_target` method for connectors which converts a room name to corresponding room id, and with this, skill constraints works.

## Status
**READY**

## Type of change

# How Has This Been Tested?

Tested with current Matrix connector like:
- Added two public rooms to localhost matrix home-server
- Wrote a skill like the skill in the issue

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -210,12 +210,10 @@ class ConnectorMatrix(Connector):
             except Exception:  # pylint: disable=W0703
                 _LOGGER.exception(_("Matrix sync error."))
 
-    def lookup_target(self, alias):
-        """Convert a room alias to the corresponding room ID."""
-        try:
-            return self.room_ids[alias]
-        except KeyError:
-            return alias
+    def lookup_target(self, room):
+        """Convert name or alias of a room to the corresponding room ID."""
+        room = self.get_roomname(room)
+        return self.room_ids.get(room, room)
 
     async def get_nick(self, roomid, mxid):
         """
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -210,6 +210,13 @@ class ConnectorMatrix(Connector):
             except Exception:  # pylint: disable=W0703
                 _LOGGER.exception(_("Matrix sync error."))
 
+    def lookup_target(self, alias):
+        """Convert a room alias to the corresponding room ID."""
+        try:
+            return self.room_ids[alias]
+        except KeyError:
+            return alias
+
     async def get_nick(self, roomid, mxid):
         """
         Get nickname from user ID.
--- a/opsdroid/constraints.py
+++ b/opsdroid/constraints.py
@@ -32,9 +32,7 @@ def constrain_rooms(rooms, invert=False):
         def constraint_callback(message, rooms=rooms):
             """Check if the room is correct."""
             if hasattr(message.connector, "lookup_target"):
-                return message.target in list(
-                    map(message.connector.lookup_target, rooms)
-                )
+                rooms = list(map(message.connector.lookup_target, rooms))
 
             return message.target in rooms
 
--- a/opsdroid/constraints.py
+++ b/opsdroid/constraints.py
@@ -7,7 +7,7 @@ having a matcher which matches the current message.
 import logging
 from functools import wraps
 
-from opsdroid.helper import add_skill_attributes, lookup_target
+from opsdroid.helper import add_skill_attributes
 
 
 _LOGGER = logging.getLogger(__name__)
@@ -31,7 +31,12 @@ def constrain_rooms(rooms, invert=False):
 
         def constraint_callback(message, rooms=rooms):
             """Check if the room is correct."""
-            return message.target in lookup_target(message, rooms)
+            if hasattr(message.connector, "lookup_target"):
+                return message.target in list(
+                    map(message.connector.lookup_target, rooms)
+                )
+
+            return message.target in rooms
 
         func = add_skill_attributes(func)
         if invert:
--- a/opsdroid/constraints.py
+++ b/opsdroid/constraints.py
@@ -7,7 +7,7 @@ having a matcher which matches the current message.
 import logging
 from functools import wraps
 
-from opsdroid.helper import add_skill_attributes
+from opsdroid.helper import add_skill_attributes, lookup_target
 
 
 _LOGGER = logging.getLogger(__name__)
@@ -31,7 +31,7 @@ def constrain_rooms(rooms, invert=False):
 
         def constraint_callback(message, rooms=rooms):
             """Check if the room is correct."""
-            return message.target in rooms
+            return message.target in lookup_target(message, rooms)
 
         func = add_skill_attributes(func)
         if invert:
--- a/opsdroid/helper.py
+++ b/opsdroid/helper.py
@@ -325,3 +325,30 @@ register_json_type(
         dct["hour"], dct["minute"], dct["second"], dct["microsecond"]
     ),
 )
+
+
+def lookup_target(event, aliases):
+    """Convert room aliases to ID's.
+
+    Converts chat room aliases to room ID's with connector object of event.
+    Translation based on `event.connector.room_ids` dict object.
+
+    Args:
+        event (object): An Event object.
+        aliases (list): List of room aliases.
+
+    Returns:
+        list: List of room ID's.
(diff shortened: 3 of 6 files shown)


FabioRosado:
Hello @thesseyren  thank you for taking the time to work on this issue and start working on the fix. 

I also think that `@contrain_rooms` should work on room aliases. Although I'm not sure how this might affect the different connectors that work a bit different than matrix like telegram 🤔

Cadair:
This PR and #1375 both provide some motivation for us to expand the base API of a connector. This one could use a method like `lookup_target` which can take an arbitrary string and returns whatever ID the connector uses for `.target`. #1375 wants a similar kind of concept for resolving user ids.

What do people think about adding a set of standardised methods like this connectors when we find needs for them? It would obviously make implementing a connector a little more involved, but I think it's worth it for the added functionality for Skill authors?

thesseyren:
I added a method which translates list of room aliases (comes from constraint method) to list of room ID's. What do you think?

opsdroid/helper.py
```diff
@@ -325,3 +325,30 @@ def register_json_type(type_cls, fields, decode_fn):
         dct["hour"], dct["minute"], dct["second"], dct["microsecond"]
     ),
 )
+
+
+def lookup_target(event, aliases):
```
Cadair: because this is matrix specific I think it should go on the matrix connector as a method. We could then implement a similar one for slack to convert from IDs like `C128371890` to `#general`. The matcher would then just call the method on whatever connector is attached to the event.

opsdroid/constraints.py
```diff
@@ -31,7 +31,7 @@ def constraint_decorator(func):
 
         def constraint_callback(message, rooms=rooms):
             """Check if the room is correct."""
-            return message.target in rooms
+            return message.target in lookup_target(message, rooms)
```
Cadair: This line would then become
```suggestion
            return message.target in message.connector.lookup_target(targets)
```

Although I feel like the connector method should be singular, i.e. it tries to convert a human-friendly name to an ID. So maybe something like this:
```suggestion
            return message.target in list(map(message.connector.lookup_target, rooms))
```
(Where `lookup_target` takes one target alias only)

opsdroid/connector/matrix/connector.py
```diff
@@ -210,6 +210,13 @@ def filter_json(self):
             except Exception:  # pylint: disable=W0703
                 _LOGGER.exception(_("Matrix sync error."))
 
+    def lookup_target(self, alias):
+        """Convert a room alias to the corresponding room ID."""
+        try:
+            return self.room_ids[alias]
+        except KeyError:
+            return alias
```
```suggestion
        return self.room_ids.get(alias, alias)
```

opsdroid/constraints.py
```diff
@@ -31,7 +31,12 @@ def constraint_decorator(func):
 
         def constraint_callback(message, rooms=rooms):
             """Check if the room is correct."""
-            return message.target in lookup_target(message, rooms)
+            if hasattr(message.connector, "lookup_target"):
+                return message.target in list(
+                    map(message.connector.lookup_target, rooms)
+                )
```
Cadair: If you do this here, then it will be evaluated below
```suggestion
                rooms = list(map(message.connector.lookup_target, rooms))
```

opsdroid/connector/matrix/connector.py
```diff
@@ -210,6 +210,13 @@ def filter_json(self):
             except Exception:  # pylint: disable=W0703
                 _LOGGER.exception(_("Matrix sync error."))
 
+    def lookup_target(self, alias):
+        """Convert a room alias to the corresponding room ID."""
```
Cadair: Can this also lookup the config names for these rooms? i.e. `'main'`? If so the docstring is a little misleading, and if not it would be cool if it could. As that allows skill developers to have different rooms which the user can then map to actual matrix rooms / aliases.

opsdroid/connector/matrix/connector.py
thesseyren: Open source world really teaches me something... Thanks a lot.

opsdroid/constraints.py
thesseyren: Yea, what a silly mistake...

opsdroid/connector/matrix/connector.py
```diff
@@ -210,12 +210,10 @@ def filter_json(self):
...
                 _LOGGER.exception(_("Matrix sync error."))
 
-    def lookup_target(self, alias):
-        """Convert a room alias to the corresponding room ID."""
-        try:
-            return self.room_ids[alias]
-        except KeyError:
-            return alias
+    def lookup_target(self, room):
+        """Convert name or alias of a room to the corresponding room ID."""
```
thesseyren: Let's check that I understand it correctly:
- Room name is a string like `main` which comes from the configuration file.
- Room alias is a Matrix specific thing, like: `#main:matrix.org`
I edited docstring in scope of these.

>Can this also lookup the config names for these rooms?

Yes it can. But, I added a call of `get_roomname`, because when a skill developer constrains a method with `#public:matrix.org` instead of just `main`, it should work too, imo. What do you think? Should I revoke this?

Cadair: Yeah I agree, having both is the way forward.

For clarity, I think the input to `lookup_target` should be able to be any of:

* "name" like `'main'`
* matrix room alias like `#matrix:matrix.org`
* a room id (`lookup_target` is a no-op)

opsdroid/connector/matrix/connector.py
```diff
@@ -210,6 +210,11 @@ def filter_json(self):
             except Exception:  # pylint: disable=W0703
                 _LOGGER.exception(_("Matrix sync error."))
 
+    def lookup_target(self, room):
+        """Convert name or alias of a room to the corresponding room ID."""
+        room = self.get_roomname(room)
+        return self.room_ids.get(room, room)
```
Cadair: I am pretty sure this will work as expected (all be it maybe not in the most efficient possible way), but it will need a bunch of tests to check :laughing:

opsdroid/constraints.py
Cadair: No worries, It still worked, only a marginal improvement!

thesseyren:
I know I need to add a test for [`parse_event_type`](https://github.com/opsdroid/opsdroid/blob/f6182c67081527a9e0d5579fc7e47e2a81697f40/opsdroid/parsers/event_type.py#L54) because I change it's behavior but I couldn't find any way to do it. Method returns nothing to test. Can you give me a hint about it?

Cadair:
I think you should be able to copy and modify this test: https://github.com/opsdroid/opsdroid/blob/master/tests/test_constraints.py#L24 which should test the rest of the code paths.

jacobtomlinson:
Just bumping this. Any change it is going to be continued?

thesseyren:
Hi @jacobtomlinson, no it's ready to merge i think.

jacobtomlinson:
Thanks @thesseyren. 

@Cadair could you confirm you are happy with the changes?

jacobtomlinson:
Thanks @Cadair and @thesseyren

PR #1418, opened 2020-04-03
Rewrite matrix connector to use matrix-nio

awesome-michael:
# Description

Rewrite of the matrix connector to use the matrix-nio library. This is _not_ tested with encryption enabled due to the fact that I was not able to install the olm library properly. But the connector is ported to nio and enabling encryption should be a much easier step now.

Please review and test a lot to ensure everything is working as expected. I have adapted all unit tests (and the are all passing) but I did't test all the opsdroid features because for some I don't know how they are supposed to work and I assume there a some I don't know of.

## Status
**READY**

## Type of change

# Checklist:


Cadair:
bad stale bot.

opsdroid/connector/matrix/connector.py
```diff
@@ -413,45 +435,46 @@ async def _send_room_creation(self, creation_event):
...
-                    f"A room with the alias {address_event.address} already exists."
-                )
+        return await self.connection.room_put_state(
+            address_event.target, "m.room.aliases", address_event.address
+        )
+        # except MatrixRequestError as err:
+        #    if err.code == 409:
+        #        _LOGGER.warning(
+        #            f"A room with the alias {address_event.address} already exists."
+        #        )
```
SolarDrew, line 452: Was there a particular reason for removing this warning?

opsdroid/connector/matrix/connector.py
```diff
@@ -413,45 +435,46 @@ async def _send_room_creation(self, creation_event):
...
+        # try:
+        return await self.connection.room_invite(
+            invite_event.target, invite_event.user_id
+        )
+        # except MatrixRequestError as err:
+        #    content = json.loads(err.content)
+        #    if err.code == 403 and "is already in the room" in content["error"]:
+        #        _LOGGER.info(
+        #            f"{invite_event.user_id} is already in the room, ignoring."
+        #        )
```
SolarDrew, line 471: As above, I feel like both of these warnings are useful for the user, but perhaps I'm missing something in how nio handles these fail cases for the api calls?

opsdroid/connector/matrix/connector.py
awesome-michael: Nope, added them in again

SolarDrew:
Thanks for the updates on this @awesome-michael . I'm happy with your changes now in themselves, unless @Cadair has any concerns. Your branch will need updating for the latest changes in master though, as it looks like you've got some conflicts - please could you pull from upstream/master and push those changes here, then we'll see if there's anything left to do on this.

opsdroid/connector/matrix/connector.py
```diff
@@ -224,18 +237,24 @@ async def get_nick(self, roomid, mxid):
         """
         if self.room_specific_nicks:
             try:
-                return await self.connection.get_room_displayname(roomid, mxid)
+                res = await self.connection.get_displayname(mxid)
```
Tyagdit, line 240: Here it should be a room specific name. Right now its just returning the default username.

opsdroid/connector/matrix/connector.py
```diff
@@ -224,18 +237,24 @@ async def get_nick(self, roomid, mxid):
...
+                if isinstance(res, nio.ProfileGetDisplayNameError):
+                    logging.exception("Failed to lookup nick for %s.", mxid)
+                    return mxid
+                if res.displayname is None:
+                    return mxid
+                return res.displayname
             except Exception:  # pylint: disable=W0703
                 # Fallback to the non-room specific one
                 logging.exception("Failed to lookup room specific nick for %s.", mxid)
```
Tyagdit, line 249: The exception is being dealt with within the try clause itself, plus I don't think the try clause would actually throw any exceptions at all.

opsdroid/connector/matrix/connector.py
```diff
@@ -224,18 +237,24 @@ async def get_nick(self, roomid, mxid):
...
-            # Log the error if it's not the 404 from the user not having a nick
-            if mre.code != 404:
-                logging.exception("Failed to lookup nick for %s.", mxid)
+        res = await self.connection.get_displayname(mxid)
+        if isinstance(res, nio.ProfileGetDisplayNameError):
+            logging.exception("Failed to lookup nick for %s.", mxid)
             return mxid
+        if res.displayname is None:
+            return mxid
+        return res.displayname
```
Tyagdit, line 257: The change above also makes this part redundant.

opsdroid/connector/matrix/connector.py
```diff
@@ -111,100 +109,115 @@ def filter_json(self):
...
+                self.room_ids[roomname] = response.room_id
+
         self.connection = mapi
 
         # Create a filter now, saves time on each later sync
-        self.filter_id = await self.make_filter(mapi)
+        # self.filter_id = await self.make_filter(mapi, self.filter_json)
+
+        # first_filter_id = await self.make_filter(mapi, '{ "room": { "timeline" : { "limit" : 1 } } }')
+        # print(first_filter_id)
```
Tyagdit, line 148: Is there a reason for removing the filter? I think it's important to only get events that we can handle.

opsdroid/connector/matrix/connector.py
```diff
@@ -111,100 +109,115 @@ def filter_json(self):
...
-        session = aiohttp.ClientSession(trust_env=True)
-        mapi = AsyncHTTPAPI(self.homeserver, session)
+        config = nio.ClientConfig(encryption_enabled=False, pickle_key="")
+        mapi = nio.AsyncClient(self.homeserver, self.mxid, config=config)
 
-        self.session = session
         login_response = await mapi.login(
-            "m.login.password", user=self.mxid, password=self.password
+            password=self.password, device_name="opsdroid"
         )
```
Tyagdit, line 128: I think it would be better to not hard code the device name, maybe we could use the nick from the config instead

Cadair: I think this is a sane default, but we might want to consider allowing it to be configured. In general with bot accounts I think you are only likely to have the single device.

Tyagdit: Yeah that seems like a better way to do it

Tyagdit:
Hey @awesome-michael any chance you'd get back on this or would you like for us to pick this up?

jacobtomlinson:
I think enough time has passed that I would support someone else taking over the PR.

@Tyagdit if you want to raise another PR which builds on this one I'd be happy to close this in favour.

awesome-michael:
Hey.. sorry I have little time at the moment. I could look into your comments. The room-specific displaynames are a concept I don't know and couldn't find info to that in nio so far. I tried to implement the filters but they didn't work so I commented them out and think forgot them.

opsdroid/connector/matrix/connector.py
awesome-michael: I think you're right. I'll remove the exception

opsdroid/connector/matrix/connector.py
awesome-michael: I'm sorry I can't find anything in the matrix spec how room specific displaynames can be set or read. Can you point me to the corresponding part of the documentation?

Tyagdit: I dont think there's a direct way to do this. Seems like you'll have to match the mxid with the usernames in the list returned by [this function](https://matrix-nio.readthedocs.io/en/latest/nio.html?highlight=members#nio.AsyncClient.joined_members) and get the corresponding displayname.

opsdroid/connector/matrix/connector.py
awesome-michael: I'm having trouble to use the predefined filters with their IDs because nio always returns a "Invalid filter ID". The homeserver accepts the generated filter IDs. Maybe its a bug in nio. Should I just send the filters with every sync request or wait until the problem is solved?

Tyagdit: I'm unable to reproduce this error. Instead there was a problem with the `filter_json` property, it needs to be serialized into json in this case, passing it directly results in a bad request. Maybe that's why the filter ID isn't being recognized? because it works properly when the filter is serialized.

awesome-michael: I also had the problem with `filter_json` but ignored it for the moment because the serialized json made a problem in the listen function. I tried a lot with the first filter but always ran in the "Invalid filter ID" response from the homeserver. How did you serialize the json?

Tyagdit: `json.dumps()` worked for me and I'm not getting "invalid filter ID" with `first_filter_id` either

opsdroid/connector/matrix/connector.py
```diff
@@ -223,19 +246,27 @@ async def get_nick(self, roomid, mxid):
...
-        try:
-            return await self.connection.get_display_name(mxid)
-        except MatrixRequestError as mre:
-            # Log the error if it's not the 404 from the user not having a nick
-            if mre.code != 404:
-                logging.exception("Failed to lookup nick for %s.", mxid)
+            res = await self.connection.joined_members(roomid)
+            if isinstance(res, nio.JoinedMembersError):
+                logging.exception("Failed to lookup room members for %s.", roomid)
+                return mxid
```
Cadair, line 255: I think this is a behaviour change. Before if this lookup failed we would fall back to the profile, not the mxid.

opsdroid/connector/matrix/connector.py
```diff
@@ -223,19 +246,27 @@ async def get_nick(self, roomid, mxid):
...
+
+            for member in res.members:
+                if member.user_id == mxid:
+                    return member.display_name
```
Cadair, line 259: I am not sure if this is better or not :thinking: take it or leave it.
```suggestion
            nick = list(filter(lambda id: id == mxid, res.members))
            if nick:
                return nick[0].display_name
```

opsdroid/connector/matrix/connector.py
```diff
@@ -223,19 +246,27 @@ async def get_nick(self, roomid, mxid):
...
+
+            return mxid
+
+        res = await self.connection.get_displayname(mxid)
+        if isinstance(res, nio.ProfileGetDisplayNameError):
+            logging.exception("Failed to lookup nick for %s.", mxid)
```
Cadair, line 265: I think that `.exception` uses the current stacktrace when inside an `except:` block. I am not sure how you should pass the exception object to it in this case.

Also I think you need to use `_LOGGER` not `logging` to make sure the logging line knows the calling scope.

opsdroid/connector/matrix/connector.py
```diff
@@ -394,17 +419,26 @@ async def _send_file(self, file_event):
...
     @register_event(events.NewRoom)
     @ensure_room_id_and_send
     async def _send_room_creation(self, creation_event):
         params = creation_event.room_params
         params = params.get("matrix", params)
-        response = await self.connection.create_room(**params)
-        room_id = response["room_id"]
+        response = await self.connection.room_create(**params)
+        if isinstance(response, nio.RoomCreateError):
+            print("Error while creating the room")
```
Cadair, line 440: should be a logger call at error.

opsdroid/connector/matrix/connector.py
```diff
@@ -414,45 +448,50 @@ async def _send_room_creation(self, creation_event):
...
-                address_event.target, address_event.address
-            )
-        except MatrixRequestError as err:
-            if err.code == 409:
+        res = await self.connection.room_put_state(
+            address_event.target, "m.room.aliases", address_event.address
+        )
+
+        if isinstance(res, nio.RoomPutStateError):
+            print("Error setting room address: " + res.message)
```
Cadair, line 463: The previous behaviour here seems wrong. I think we should raise the exception unless it's a `409` in which case we should log and then silently continue.

opsdroid/connector/matrix/connector.py
```diff
@@ -414,45 +448,50 @@ async def _send_room_creation(self, creation_event):
...
-            if err.code == 403 and "is already in the room" in content["error"]:
+        res = await self.connection.room_invite(
+            invite_event.target, invite_event.user_id
+        )
+
+        if isinstance(res, nio.RoomInviteError):
+            if res.status_code == 403 and "is already in the room" in res.message:
                 _LOGGER.info(
                     f"{invite_event.user_id} is already in the room, ignoring."
                 )
```
Cadair, line 487: As above we should throw the exception if we don't understand it.
```suggestion
        if isinstance(res, nio.RoomInviteError):
            if res.status_code == 403 and "is already in the room" in res.message:
                _LOGGER.info(
                    f"{invite_event.user_id} is already in the room, ignoring."
                )
            else:
                raise res
```

opsdroid/connector/matrix/connector.py
awesome-michael: In my opinion the previous version is more understandable ... so I would leave it as is

opsdroid/connector/matrix/connector.py
```diff
@@ -104,107 +104,130 @@ def _process_rooms_dict(self, rooms):
...
+
+        mapi.token = login_response.access_token
         mapi.sync_token = None
 
         for roomname, room in self.rooms.items():
-            response = await mapi.join_room(room["alias"])
-            self.room_ids[roomname] = response["room_id"]
+            response = await mapi.join(room["alias"])
+            if isinstance(response, nio.JoinError):
+                print("Error while joining room: " + roomname)
```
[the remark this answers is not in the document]
awesome-michael: Will opsdroid stop running if an exception is raised here? Then I think we should just log because joining the other rooms may work well.

opsdroid/connector/matrix/connector.py
```diff
@@ -223,20 +246,35 @@ async def get_nick(self, roomid, mxid):
...
-        except MatrixRequestError as mre:
-            # Log the error if it's not the 404 from the user not having a nick
-            if mre.code != 404:
-                logging.exception("Failed to lookup nick for %s.", mxid)
             return mxid
 
+        res = await self.connection.get_displayname(mxid)
+        if isinstance(res, nio.ProfileGetDisplayNameError):
+            _LOGGER.error("Failed to lookup nick for %s.", mxid)
+            return mxid
```
SolarDrew, line 273: Since this "if error: log error, return" structure is coming up a lot here, I wonder if it's worth defining a function or a context manager to wrap it, just to make the code in the connector a little tidier? Probably not worth worrying about for this PR but something to consider going forward.

opsdroid/connector/matrix/connector.py
```diff
@@ -414,45 +455,53 @@ async def _send_room_creation(self, creation_event):
...
-                invite_event.target, invite_event.user_id
-            )
-        except MatrixRequestError as err:
-            content = json.loads(err.content)
-            if err.code == 403 and "is already in the room" in content["error"]:
+        res = await self.connection.room_invite(
+            invite_event.target, invite_event.user_id
+        )
+
+        if isinstance(res, nio.RoomInviteError):
```
SolarDrew, line 490: The test failures seem to be suggesting that `nio.RoomInvitationError` doesn't inherit from `Exception`, and it's causing problems. Is this the correct class to use here?

awesome-michael: seems that all the nio ErrorResponse classes don't inherit from `Exception`. Should I replace those passages with something like
```
if isinstance(res, nio.RoomInviteError):
  _LOGGER.error(res.message)
  raise
```
?

Cadair: I think that makes sense then. Are there other properties on those error classes which are worth putting in the log? http code etc?

opsdroid/connector/matrix/connector.py
```diff
@@ -104,107 +104,130 @@ def _process_rooms_dict(self, rooms):
...
-        mapi = AsyncHTTPAPI(self.homeserver, session)
+        config = nio.ClientConfig(encryption_enabled=False, pickle_key="")
+        mapi = nio.AsyncClient(self.homeserver, self.mxid, config=config)
 
-        self.session = session
         login_response = await mapi.login(
-            "m.login.password", user=self.mxid, password=self.password
+            password=self.password, device_name=self.device_name
         )
-        mapi.token = login_response["access_token"]
```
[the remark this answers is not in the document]
Cadair: oh wait that never got merged :facepalm:

opsdroid/connector/matrix/connector.py
```diff
@@ -104,107 +104,133 @@ def _process_rooms_dict(self, rooms):
...
+
+        mapi.token = login_response.access_token
         mapi.sync_token = None
 
         for roomname, room in self.rooms.items():
-            response = await mapi.join_room(room["alias"])
-            self.room_ids[roomname] = response["room_id"]
+            response = await mapi.join(room["alias"])
+            if isinstance(response, nio.JoinError):
+                _LOGGER.info("Error while joining room: " + roomname)
```
Cadair, line 149: I feel like this should probably be at `error`? It's an action the user has specified in the config file?

SolarDrew:
Looks like the tests are all passing now so it just needs a couple more to keep the coverage up and then I think this is there. Looks like it's mostly error-catching that's not included in the current tests, so we should make sure those errors are raised and check that they're caught and handled correctly.

opsdroid/connector/matrix/connector.py
```diff
@@ -104,107 +104,133 @@ def _process_rooms_dict(self, rooms):
...
+                            # event is encrypted
+                            event = await self.connection.decrypt_event(event)
+                        return await self._event_creator.create_event(
+                            event.source, roomid
+                        )
 
     async def listen(self):  # pragma: no cover
         """Listen for new messages from the chat service."""
         while True:  # pylint: disable=R1702
             try:
```
Cadair, line 217: Is this `try` block still relevant with `nio`? It looks like the pattern with nio is to return errors rather than raise them, so I assume the best thing to do would be to remove this?

opsdroid/connector/matrix/connector.py
```diff
@@ -104,107 +104,133 @@ def _process_rooms_dict(self, rooms):
...
                 response = await self.connection.sync(
-                    self.connection.sync_token,
-                    timeout_ms=int(60 * 1e3),  # 1m in ms
-                    filter=self.filter_id,
+                    timeout=int(60 * 1e3),  # 1m in ms
+                    sync_filter=self.filter_id,
+                    since=self.connection.sync_token,
                 )
+                if isinstance(response, nio.SyncError):
```
Cadair, line 223: Is this type also returned if you get a http error like a 504, 522 etc? If not we should also handle those cases.

opsdroid/connector/matrix/connector.py
```diff
@@ -394,17 +429,26 @@ async def _send_file(self, file_event):
...
     @register_event(events.NewRoom)
     @ensure_room_id_and_send
     async def _send_room_creation(self, creation_event):
         params = creation_event.room_params
         params = params.get("matrix", params)
-        response = await self.connection.create_room(**params)
-        room_id = response["room_id"]
+        response = await self.connection.room_create(**params)
+        if isinstance(response, nio.RoomCreateError):
+            _LOGGER.error("Error while creating the room")
```
Cadair, line 450: This should also log why.

Cadair:
@awesome-michael Thanks so much for continuing to work on this. Just so you know, @Tyagdit is starting the GSOC coding period today, and is planning on working off your work here, to get things like e2ee going etc. So we would like to get this merged as soon as practical.

It's up to you if you want to keep iterating on this, or if you want to hand it over to @Tyagdit to get it across the line, either in this or another PR. If you are pressed for time, or have had enough of nit picky comments feel free to leave it up to @Tyagdit, but if you would like the satisfaction of getting it cleaned up yourself then it's all yours :grin:

opsdroid/connector/matrix/connector.py
awesome-michael: afaik http errors are also returned as errors

awesome-michael:
Is it acceptable that the one single line that would handle encrypted events is not covered by tests yet?

opsdroid/connector/matrix/connector.py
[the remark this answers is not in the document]
SolarDrew: Yes I think it will. My preference would be to raise the error - some rooms may work but without further info from the user I think we have to assume that the bot needs to join every room in the config. (Some way to specify higher or lower priority of rooms might be a good way to address this, but that's a different PR.)

SolarDrew:
I'm fine with that given that should be covered by the work @Tyagdit is doing.

Cadair:
> Is it acceptable that the one single line that would handle encrypted events is not covered by tests yet?

I am pretty sure that encrypted events wont work as this PR stands anyway, so you could also just comment it out. (but yeah it's fine as @SolarDrew says)

opsdroid/connector/matrix/connector.py
```diff
@@ -104,111 +103,136 @@ def _process_rooms_dict(self, rooms):
...
+        )
 
-    async def make_filter(self, api):
+    async def make_filter(self, api, fjson):
         """Make a filter on the server for future syncs."""
-        fjson = self.filter_json
-        resp = await api.create_filter(user_id=self.mxid, filter_params=fjson)
-        return resp["filter_id"]
+        path = f"/_matrix/client/r0/user/{self.mxid}/filter"
+        headers = {"Authorization": "Bearer {api.token}"}
```
Tyagdit, line 122: This needs to be an f-string

opsdroid/connector/matrix/connector.py
```diff
@@ -47,7 +45,7 @@ async def ensure_room_id(self, event):
             event.target = self.room_ids[event.target]
 
         if not event.target.startswith("!"):
-            event.target = await self.connection.get_room_id(event.target)
+            event.target = await self.connection.request_room_key(event.target)
```
Cadair, line 48: This isn't the right function. I think you need [`resolve_room_alias`](https://github.com/poljar/matrix-nio/blob/053be26ceca5a24422b471187f4d0e7d044807b5/nio/client/async_client.py#L1636)

opsdroid/connector/matrix/connector.py
```diff
@@ -104,111 +103,136 @@ def _process_rooms_dict(self, rooms):
...
-        for roomid, room in response["rooms"]["invite"].items():
+        for roomid, roomInfo in response.rooms.invite.items():
             # Process the invite list to extract the person who invited us.
             invite_event = [
                 e
-                for e in room["invite_state"]["events"]
-                if "invite" == e.get("content", {}).get("membership")
+                for e in roomInfo.invite_state
+                if isinstance(e, nio.InviteMemberEvent)
+                if "invite" == e.membership
```
Cadair, line 193: This is real picky, but it' reads better this way round :grin: 
```suggestion
                if e.membership == "invite"
```

opsdroid/connector/matrix/connector.py
```diff
@@ -104,111 +103,136 @@ def _process_rooms_dict(self, rooms):
...
-                for event in room["timeline"]["events"]:
-                    if event["sender"] != self.mxid:
-                        return await self._event_creator.create_event(event, roomid)
+        for roomid, roomInfo in response.rooms.join.items():
+            if roomInfo.timeline:
+                for event in roomInfo.timeline.events:
+                    if event.sender != self.mxid:
+                        if isinstance(event, nio.MegolmEvent):
+                            # event is encrypted
+                            event = await self.connection.decrypt_event(event)
```
Cadair, line 213:
```suggestion
                            # event = await self.connection.decrypt_event(event)
                            continue
```

opsdroid/connector/matrix/connector.py
```diff
@@ -104,111 +103,136 @@ def _process_rooms_dict(self, rooms):
...
+                        return await self._event_creator.create_event(
```
Cadair, line 214: I know this was like this before, but should this actually be return here? What happens if there is more than one event in the sync response?!

(we probably need to open an issue for this and not worry about it in this PR).

opsdroid/connector/matrix/connector.py
```diff
@@ -223,20 +247,35 @@ async def get_nick(self, roomid, mxid):
         setting.
         """
         if self.room_specific_nicks:
-            try:
-                return await self.connection.get_room_displayname(roomid, mxid)
-            except Exception:  # pylint: disable=W0703
-                # Fallback to the non-room specific one
-                logging.exception("Failed to lookup room specific nick for %s.", mxid)
+            if roomid is None:
+                return mxid
```
Cadair, line 251: I would move this up into the top if statement, and fall back to not-roomspecific nicks if room is `None`.
```suggestion
```

opsdroid/connector/matrix/connector.py
```diff
@@ -223,20 +247,35 @@ async def get_nick(self, roomid, mxid):
         setting.
         """
         if self.room_specific_nicks:
```
Cadair, line 249:
```suggestion
        if self.room_specific_nicks and roomid is not None:
```

opsdroid/connector/matrix/connector.py
```diff
@@ -223,20 +247,35 @@ async def get_nick(self, roomid, mxid):
...
+                if isinstance(res, nio.ProfileGetDisplayNameError):
+                    _LOGGER.error("Failed to lookup nick for %s.", mxid)
+                    return mxid
+                if res.displayname is None:
+                    return mxid
+                return res.displayname
+
+            for member in res.members:
+                if member.user_id == mxid:
+                    return member.display_name
```
Cadair, line 267: I think you can re-write this logic so that it drops out of the top `if self.room_specific_nicks` block to resolve the global profile nick, rather than rewriting this here.

```suggestion
            else:
                for member in res.members:
                    if member.user_id == mxid:
                        return member.display_name
```

opsdroid/connector/matrix/connector.py
awesome-michael: I think you're right. The return seems not to be correct here

opsdroid/connector/matrix/connector.py
awesome-michael: You're right

opsdroid/connector/matrix/connector.py
awesome-michael: hmm not sure how to write this as every function has its own error class that is returned

opsdroid/connector/matrix/connector.py
awesome-michael: A sorry.. no the return is needed so yeah we should probably open an issue for that

opsdroid/connector/matrix/connector.py
awesome-michael: Anyone has an idea why the one test fails? Seems like the error is not logged but I have no clue why

awesome-michael:
Anyone has an idea why the one test fails? Seems like the error is not logged but I have no clue why

Cadair:
It looks like the caplog test for the resolve room alias isn't working, it hasn't captured any log lines for the 400 error?

awesome-michael:
No.. it seems that nothing is logged. But I don't know why

Cadair:
I couldn't see it either so I checked the branch out locally. It was because the mock wasn't being reset from the lines above which tested the working case, so I moved the error test into it's own test to make sure things were torn down properly.

opsdroid/connector/matrix/connector.py
```diff
@@ -414,45 +458,57 @@ async def _send_room_creation(self, creation_event):
     @register_event(events.RoomName)
     @ensure_room_id_and_send
     async def _send_room_name_set(self, name_event):
-        return await self.connection.set_room_name(name_event.target, name_event.name)
+        return await self.connection.room_put_state(
```
Cadair, line 461: I am taking this as a generic example of a pattern here.

Previously before nio if a call like this errored, the error would raise and be caught in the logs as the outgoing send would fail. I am not sure what the best solution to this is, if all the `nio.XXXError` classes have a common parent we could write a decorator which catches them in the return and logs them at error?

(We can probably punt this to an issue post merge)

Cadair:
I think there are some issues that need raising after we merge this:

- https://github.com/opsdroid/opsdroid/pull/1418#discussion_r436615260
- https://github.com/opsdroid/opsdroid/pull/1418#discussion_r435176024
- https://github.com/opsdroid/opsdroid/pull/1418#discussion_r432383617

opsdroid/connector/matrix/connector.py
```diff
@@ -104,111 +109,139 @@ def _process_rooms_dict(self, rooms):
...
-                raise
-            except Exception:  # pylint: disable=W0703
-                _LOGGER.exception(_("Matrix sync error."))
+            response = await self.connection.sync(
+                timeout=int(60 * 1e3),  # 1m in ms
+                sync_filter=self.filter_id,
+                since=self.connection.sync_token,
+            )
+            if isinstance(response, nio.SyncError):
+                _LOGGER.error(f"Error during sync: {response.message}")
```
Cadair, line 236:
```suggestion
                _LOGGER.error(f"Error during sync: {response.message} (status code {response.status_code})")
```

opsdroid/connector/matrix/connector.py
```diff
@@ -104,111 +109,139 @@ def _process_rooms_dict(self, rooms):
...
         """Make a filter on the server for future syncs."""
-        fjson = self.filter_json
-        resp = await api.create_filter(user_id=self.mxid, filter_params=fjson)
-        return resp["filter_id"]
+        path = f"/_matrix/client/r0/user/{self.mxid}/filter"
+        headers = {"Authorization": f"Bearer {api.token}"}
+        resp = await api.send(method="post", path=path, data=fjson, headers=headers)
+
+        resp_json = await resp.json()
+        return resp_json["filter_id"]
```
Cadair, line 132: I tested this branch out locally and this caused all my syncs to fail:
```suggestion
        return int(resp_json["filter_id"])
```

opsdroid/connector/matrix/create_events.py
```diff
@@ -30,7 +30,8 @@ class MatrixEventCreator(events.EventCreator):
 
     async def create_event_from_eventid(self, eventid, roomid):
         """Return an ``Event`` based on an event id in a room."""
-        event_json = await self.connector.connection.get_event_in_room(roomid, eventid)
+        room_context = await self.connector.connection.room_context(roomid, eventid, 1)
+        event_json = room_context.event
```
Tyagdit, line 34: `room_context.event` returns a `RoomMessageText` instead of a dict
```suggestion
        event_json = room_context.event.source
```

Also [down here](https://github.com/opsdroid/opsdroid/pull/1418/files#diff-9d595b55ddf1583136a938382e129321R122) in `_file_kwargs()` the function is not part of nio, instead the url can be retrieved like so 
```
url = event["content"]["url"]
```

opsdroid/connector/matrix/connector.py
```diff
@@ -104,111 +109,139 @@ def _process_rooms_dict(self, rooms):
...
+            if roomInfo.timeline:
+                for event in roomInfo.timeline.events:
+                    if event.sender != self.mxid:
+                        # if isinstance(event, nio.MegolmEvent):
+                        # event is encrypted
+                        # event = await self.connection.decrypt_event(event)
+
+                        return await self._event_creator.create_event(
+                            event.source, roomid
+                        )
```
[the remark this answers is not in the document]
Cadair: The event creator really needs the content dict, as it's supposed to set it as `raw_event` on the generated `opsdroid.Event`. Maybe it would be better to pass that through?

opsdroid/connector/matrix/create_events.py
```diff
@@ -119,7 +119,7 @@ async def create_message(self, event, roomid):
         return events.Message(**kwargs)
 
     async def _file_kwargs(self, event, roomid):
-        url = self.connector.connection.get_download_url(event["content"]["url"])
+        url = event["content"]["url"]
```
Cadair, line 122: Am I correct in assuming this will be the `mxc://` url? This needs to be converted into a https URL, for people to be able to interact with it in a sensible way.

Cadair: Which is what that function used to do.

Cadair: I think you need https://github.com/poljar/matrix-nio/blob/master/nio/client/async_client.py#L768 :smile:

opsdroid/connector/matrix/connector.py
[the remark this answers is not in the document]
Tyagdit: You mean opsdroid expects `raw_event` to have a `content` dict? 
In that case we could do its as 
```
event.source["content"] = event.content
```
Or does content need more stuff in it?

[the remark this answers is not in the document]
Cadair: I mean `raw_event` should be the content dict. What is `event.source` and `event.content`? I am not really familiar with these nio classes.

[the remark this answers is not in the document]
Tyagdit: I think source is the info about the response and content is the actual body of the event. In other nio event classes content is part of source but not in this case, not sure why. For all the other events we've been passing source as raw_event since that's where the stuff like sender and event type is

Cadair:
I have opened #1529 as a catchall for followups from this PR, as I expect there will be a few.

Cadair:
@SolarDrew I hand this to you to stamp and merge.

Cadair:
Thank you so much for all your work on this @awesome-michael it's a fantastic change :rocket: :tada:

PR #1437, opened 2020-04-19
Implement a Generic Event type for all unsupported matrix events

Cadair:
# Description

This PR implements a `GenericMatrixRoomEvent` class which is accepted for send or emitted when we receive an event over matrix which we don't support.

## Status
**Done**

## Type of change

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/events.py
+++ b/opsdroid/connector/matrix/events.py
@@ -59,3 +59,12 @@ class MatrixRoomAvatar(MatrixStateEvent):
         key = "m.room.avatar"
         content = {"url": url}
         super().__init__(key, content, *args, **kwargs)
+
+
+class UnknownMatrixRoomEvent(Event):
+    """A matrix event which we don't understand."""
+
+    def __init__(self, event_type, content, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.content = content
+        self.event_type = event_type
--- a/opsdroid/connector/matrix/events.py
+++ b/opsdroid/connector/matrix/events.py
@@ -21,12 +21,12 @@ class GenericMatrixRoomEvent(Event):
     matrix specific events.
     """
 
-    def __init__(self, event_type, content, *args, **kwargs):
+    def __init__(self, event_type, content, *args, **kwargs):  # noqa: D107
         super().__init__(*args, **kwargs)
         self.content = content
         self.event_type = event_type
 
-    def __repr__(self):  # noqa: D107
+    def __repr__(self):  # noqa: D105
         return f"<GenericMatrixRoomEvent(room_id={self.target}, event_type={self.event_type}, content={self.content})>"
 
 
@@ -37,7 +37,7 @@ class MatrixStateEvent(GenericMatrixRoomEvent):
         super().__init__(*args, **kwargs)
         self.state_key = state_key
 
-    def __repr__(self):  # noqa: D107
+    def __repr__(self):  # noqa: D105
         return f"<MatrixStateEvent(room_id={self.target}, event_type={self.event_type}, state_key={self.state_key}, content={self.content})>"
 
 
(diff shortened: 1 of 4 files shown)


opsdroid/connector/matrix/events.py
```diff
@@ -12,50 +14,76 @@
...
+        self.event_type = event_type
+
+    def __repr__(self):  # noqa: D107
+        return f"<GenericMatrixRoomEvent(room_id={self.target}, event_type={self.event_type}, content={self.content})>"
+
+
+class MatrixStateEvent(GenericMatrixRoomEvent):
+    """A Generic matrix state event."""
+
+    def __init__(self, *args, state_key=None, **kwargs):  # noqa: D107
```
Cadair, line 36: This is a small ish breaking change for `MatrixStateEvent` as it no longer accepts "key=" as a kwarg (it's still the same positional arg) it's now a more correct "event_type".

Cadair:
I still need to actually test this IRL

opsdroid/connector/matrix/events.py
```diff
@@ -12,50 +14,76 @@
...
     """Send power levels."""
 
+    # Deprecate old name for event_type
+    @property
+    def key(self):
+        warnings.warn(
+            "key has been renamed event_type to reduce confusion with state_key",
+            DeprecationWarning,
+        )
+        return self.event_type
```
SolarDrew, line 54: Nice :+1:

tests/test_connector_matrix.py
```diff
@@ -1052,3 +1066,28 @@ async def test_create_joinroom(self):
...
+    def custom_json(self):
+        return {
+            "content": {"hello": "world"},
+            "event_id": "$15573463541827394vczPd:localhost",
+            "origin_server_ts": 1557346354253,
+            "room_id": "!test:localhost",
+            "sender": "@neo:matrix.org",
+            "type": "opsdroid.dev",
+            "unsigned": {"age": 48926251},
+            "user_id": "@nso:matrix.org",
```
SolarDrew, line 1080: Not @neo?

tests/test_connector_matrix.py
```diff
@@ -1052,3 +1066,28 @@ async def test_create_joinroom(self):
...
+            "age": 48926251,
+        }
+
+    async def test_create_generic(self):
+        event = await self.event_creator.create_event(self.custom_json, "hello")
+        assert isinstance(event, matrix_events.GenericMatrixRoomEvent)
+        assert event.user == "Rabbit Hole"
```
SolarDrew, line 1087: This could just be me forgetting how the connector works but where is it getting this from?

opsdroid/connector/matrix/events.py
Cadair: it broke the database, but I pushed a fix which works either way already.

tests/test_connector_matrix.py
Cadair: it's mocked in the event_creator property

tests/test_connector_matrix.py
Cadair: this shouldn't actually be here so I removed it and all other instances of it.

Cadair:
Background: A state_event in matrix is just a Room event which has `state_key` in the top level of the event.

> So, to be clear, if the connector gets an event it doesn't recognise as an explicitly supported event type, it will try to construct it as a generic event?

Yes, and if it's a state event (`state_key` is in the event), then it will construct a `MatrixStateEvent` (new commit pushed this morning).

> And when you send a generic event to matrix it'll send it as a state event?

No, it will send it as a generic event. If you send a `MatrixStateEvent` event it gets handled separately. However, I can't accept subclasses of `GenericMatrixRoomEvent` because that also includes `MatrixStateEvent`.

Finally, this is a useful tool, but needs to be used with caution, because if we ever add support for an event, it will break anyone using this to react to them while they are unsupported, but I think this is also fine.

SolarDrew:
>  if we ever add support for an event, it will break anyone using this to react to them while they are unsupported, but I think this is also fine.

Yeah I think if we break anyone's workaround by adding explicit support for something they'll have limited scope for complaint there.

LGTM :+1:

PR #1466, opened 2020-05-12
Reload opsdroid if a local module is modified

jacobtomlinson:
If you write your own opsdroid skill and configure it via the `path` option that path will now be watched and if any changes are made to python files within that path opsdroid will automatically reload to pick up those changes.

```yaml
...

skills:
  test:
    path: /path/to/my/skill
```

Modifying any Python file within `/path/to/my/skill` will trigger a reload.

---

For this to work I also had to handle importing a broken module better, the import exception is now logged with an error.

If `/path/to/my/skill/__init__.py` looks something like:

```python
import loging  # Should be logging

...
```

we now see:

```
ERROR opsdroid.loader: The following exception was raised while importing skill opsdroid-modules.skill.test
ERROR opsdroid.loader: No module named 'loging'
ERROR opsdroid.loader: Failed to load skill: opsdroid-modules.skill.test.
ERROR opsdroid.loader: Module test failed to import.
```

### DIFF
--- a/opsdroid/loader.py
+++ b/opsdroid/loader.py
@@ -101,9 +101,15 @@ class Loader:
                 continue
 
         if module_spec:
-            module = Loader.import_module_from_spec(module_spec)
-            _LOGGER.debug(_("Loaded %s: %s."), config["type"], config["module_path"])
-            return module
+            try:
+                module = Loader.import_module_from_spec(module_spec)
+            except Exception as e:
+                _LOGGER.error(str(e))
+            else:
+                _LOGGER.debug(
+                    _("Loaded %s: %s."), config["type"], config["module_path"]
+                )
+                return module
 
         _LOGGER.error(
             _("Failed to load %s: %s."), config["type"], config["module_path"]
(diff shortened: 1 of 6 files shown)


opsdroid/loader.py
```diff
@@ -101,9 +101,20 @@ def import_module(config):
                 continue
 
         if module_spec:
-            module = Loader.import_module_from_spec(module_spec)
-            _LOGGER.debug(_("Loaded %s: %s."), config["type"], config["module_path"])
-            return module
+            try:
+                module = Loader.import_module_from_spec(module_spec)
+            except Exception as e:
+                _LOGGER.error(
```
FabioRosado, line 107: Perhaps this should be 
```suggestion
                _LOGGER.error(_("The following exception was raised while importing %s %s"), config["type"], config["module_path"])
```
due to the i18n/i10n (never know which is which haha) we are doing?

jacobtomlinson:
Thanks for the review @FabioRosado I had missed that. I've commited your suggestion and I'll get this merged on passing CI.

PR #1472, opened 2020-05-14
Replace CI on Travis and AppVeyor with GitHub Actions 

jacobtomlinson:
CI has been pretty slow lately.

This is a test to see how easy it would be to switch our CI to GitHub Actions for more parallelism and hopefully less queueing.


jacobtomlinson:
Ok this seems to work pretty nicely. I'm inclined to merge this and disable Travis and AppVeyor completely.

Before I do it would be good to get some thoughts from the other maintainers.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -U tox tox-gh-actions
+    - name: Run tests
+      run: tox
+    - name: Upload coverage report
+      uses: codecov/codecov-action@v1
+      if: ${{ matrix.os == 'ubuntu-latest' && matrix.python-version == '3.8' }}
```
Cadair, line 26: can't you just upload them all with codecov and it unions them?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,26 @@
...
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
```
Cadair, line 25: is running bdist wheel what we actually want here? Does that create a universal wheel?

.github/workflows/ci.yml
jacobtomlinson: I guess so, just seemed wasteful.

.github/workflows/deploy.yml
jacobtomlinson: Hmm good point. This was just copy/pasta from the GitHub docs. What would you recommend here?

jacobtomlinson:
```suggestion
        python setup.py sdist
```

.github/workflows/ci.yml
jacobtomlinson:
```suggestion
```

FabioRosado:
This looks so great! I really want to dig into github actions and this PR made me want to play around with them even more! I'm pretty happy to ditch Travis, it has served us well but yeah we are having a lot of issues with it. Not only it takes a long time to run sometimes it doesn't report the status to github and things just stay untouched until something unstucks the report (still dont know why or how)


So disabling travis I'm up for it. Appveyor doesn't bother me, but we had it because of the windows compatibility, if we ditch appveyor will we risk of breaking opsdroid on windows? Also is opsdroid working 100% on windows to even matter? 🤔

jacobtomlinson:
Yeah Travis has been causing enough problems that I'm ready to try something else.

GitHub Actions also runs on Windows and MacOS. So we should be better off!

FabioRosado:
Oh awesome! Then yeah I'd say move the whole thing to github actions should be easier to maintain as well 👍

PR #1497, opened 2020-05-25
Add Twitch Connector 

FabioRosado:
# Description

This is the initial commit for the Twitch connector, there is still a lot to be done  like writing the tests and create the documentation for it. I'm raising this draft PR to get some feedback on the word done and if I can improve things a bit more.

This connector is pretty big because I'm using a lot of things. Twitch has different ways to do stuff so we are using webhooks to get 3 events, then calling the API to update channel titles and websockets to listen to the chat.

I thought about what other bots do and what could be cool to include and that's the reason this connector grew so much haha

## Things still to be done

- [x] Update documentation
- [x] Write tests for the whole connector
- [x] Test most of the features on Twitch (did some but still need further testing)
- [x] Write the `skill-twitch` to handle different events and actions
- [x] Figure out why websockets throw exception when exiting opsdroid
- [ ] Add ssl support to connect to websockets - still need to figure out how to do that.
- [x] Check if every logger contains `_()`
- [x] Figure out and implement `hub.secret` check when handling webhook response - prevent folks from sending data that wasn't sent by Twitch.
- [x] Connect to the chat service only when streamer went online.
- [x] Handle aiohttp status code != 200. Currently I'm assuming everything works fine - that is BAAAAAD

## Notes

- `config['forward-url'] - Added this so we can still run opsdroid on localhost and use ngrok or serveo forwarding urls as a callback when subscribing to webhooks.
- `token` expires after 24 hours so we will refresh token when it expires and will re-subscribe to webhooks
- `twitch.json` - saved data on appdir `data` to keep the oauth token and refresh token saved so we can get/set
- We should really only connect to the twitch irc whenever a streamer goes live, it's unlikely that we will do anything with the chat until we are online. I need to check if the webhook will still receive data from Twitch after the oauth token expired (but the webhook lease is still going), that way we can keep the 'refresh' token code on the websockets connect method

Every time we send something through websocket the command is very similar, should we just create a helper method to send messages to the websocket? Will this improve readability?

## Status

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:


.gitignore
```diff
@@ -106,4 +106,6 @@ ENV/
 site/
 Pipfile
 Pipfile.lock
-.icloud
\ No newline at end of file
+.icloud
+
+TODO
```
jacobtomlinson, line 111: These should probably be in your [global .gitignore](https://stackoverflow.com/a/7335487/1003288)

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,576 @@
...
+
+
+def build_api_url(client_id, client_secret, authorization, redirect):
+    """Build api url to authenticate and refresh token."""
+    clientid = f"?client_id={client_id}"
+    secret = f"&client_secret={client_secret}"
+    grant = f"&grant_type={authorization}"
+    redirect = f"&redirect_uri={redirect}"
+
+    return f"{TWITCH_API_ENDPOINT}{clientid}{secret}{grant}{redirect}"
```
jacobtomlinson, line 39: Have you seen the [docs on passing URL paramaters with aiohttp](https://docs.aiohttp.org/en/stable/client_quickstart.html#passing-parameters-in-urls)? You don't need to build urls like this, you can pass a dictionary of parameters in the `get` method and it will do it for you.

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,576 @@
...
+
+async def get_user_id(channel, token, client_id):
+    """Calls twitch api to get broadcaster user id.
+    
+    A lot of webhooks expect you to pass your user id in order to get the 
+    notification when a user subscribes or folllows the broadcaster
+    channel.
+    
+    """
+    url = f"https://api.twitch.tv/helix/users?login={channel}"
```
jacobtomlinson, line 49: The `login` here should probably also be passed as a parameter. That way special characters will be handled for you.

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,576 @@
...
+    async with aiohttp.ClientSession() as session:
+        response = await session.get(url, headers={
+            'Authorization': f"Bearer {token}",
+            'Client-ID': client_id
+        })
+        
+        if response.status != 200:
```
jacobtomlinson, line 56: I recommend changing this to >=400 as there are many valid reasons why you would get a non 200 error with a valid response body. A temporary redirect for example.

```suggestion
        if response.status >= 400:
```

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,576 @@
...
+        self.server = "ws://irc-ws.chat.twitch.tv"
+        self.port = "80"
+        self.forward_url = config['forward-url']
+        # Setup routes for webhooks subscription
+        self.opsdroid.web_server.web_app.router.add_get(
+            f"/connector/{self.name}", self.handle_challenge
+        )
+        self.opsdroid.web_server.web_app.router.add_post(
+            f"/connector/{self.name}", self.twitch_webhook_handler
+        )
```
jacobtomlinson, line 92: These should be in the `connect` method. The `__init__` shouldn't do anything other than setting attributes of the instance.

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,576 @@
...
+        Returns:
+           aiohttp.web.Response: if request contains `hub.challenge` we return it, otherwise return status 500.
+        
+        """
+        
+        challenge = request.rel_url.query.get('hub.challenge')
+
+        if challenge:
+            return aiohttp.web.Response(text=challenge)
+        return aiohttp.web.Response(status=500)
```
jacobtomlinson, line 293: Maybe log here to show the user the challenge failed.

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,576 @@
...
+        Args:
+            request (aiohttp.web.Request): Request made to the post route created for webhook subscription.
+
+        Return:
+            aiohttp.web.Response: Send a `received` message and status 200 - Twitch will keep sending the event if 
+            it doesn't get the 200 status code. 
+
+        """
+        payload = await request.json()
+        data = payload.get('data')[0]
```
jacobtomlinson, line 320: If the list is only ever one item it is more Pythonic to unpack at assingment like this

```suggestion
        [data] = payload.get('data')
```

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,576 @@
...
+                    message=data['event_data']['message']
+                )
+                
+                await self.opsdroid.parse(user_subscription)
+                
+        except KeyError:
+            if not payload.get('data'):
+                # When the stream goes offline, Twitch will return `data: []`
+                # if data is none we assume the streamer whent offline
+                stream_ended = twitch_event.StreamEnded(connector=self)
```
jacobtomlinson, line 369: You don't seem to be parsing this event.

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,576 @@
...
+        We also need to check if whatever we received from the websocket is indeed a text
+        message that we need to parse, so we use a simple if statement to check if the data
+        received matches a text message.
+        
+        """
+        
+        try:
+            resp = await self.websocket.recv()
+        except websockets.ConnectionClosed:
+            await self.reconnect()
```
jacobtomlinson, line 447: I would reconnect and then return here. That way you go round the `while self.listening` loop and come back in at the top of the try again. Otherwise if the next `self.websocket.recv` also gets a `websockets.ConnectionClosed` exception it will not be caught.
```suggestion
            await self.reconnect()
            return
        else:
```

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,576 @@
...
+        """
+        async with aiohttp.ClientSession() as session:
+            headers = {
+                "Client-ID": self.client_id, 
+                "Authorization": f"OAuth {self.token}",
+                "Accept": "application/vnd.twitchtv.v5+json"
+            
+            }
+            status = event.status.replace(" ", "+")
+            url = f"https://api.twitch.tv/kraken/channels/{self.user_id}?channel[status]={status}"
```
jacobtomlinson, line 547: Should this URL be a constant too?

.gitignore
FabioRosado: You are right and didn't knew about global .gitignore haha thanks for pointing it out 😄

FabioRosado:
Hello Jacob I've resolved all of the comments that you've made and did some testing of the connector on my end on stream as well. Did some more refactoring and decided to add a `always-listening` flag to the config, if this flag is set opsdroid will keep trying to connect and reconnect to the twitch server.

The `always-listening` flag defaults to false so the default behavior of the connector is to listen to the webhook and only start listening to the chat whenever a broadcaster went live. This seemed like the best approach since the chat will be empty when the broadcaster isn't there and twitch keps closing the websocket connection.

I've been thinking about the webhooks. Currently Twitch asks you to set a lease for the subscription to expire. I assume that Twitch will keep sending events even if our oauth expired (but need to test this out on my end).

I'm thinking to somehow add a crontab to run a method that will trigger a token refresh. Should this be done in the connector or on a skill? 🤔

jacobtomlinson:
I would do it in the listen. You are looping there right? So put a check in that says something like `if time_since_last_check > 1hr: refresh()`.

FabioRosado:
Sorry for the late reply, I've refactored the code a bit and we are not really using the loop to listen the websockets since I made the decision to listen for the "is live" event sent through the webhook and then connect to the chat service

FabioRosado:
I've been livetesting the connector and the bot on stream for the past 3 streams and I'm trying to squash some annoying bugs. Twitch seems to close the websockets more often when I'm live (even if opsdroid is open on a digital ocean droplet).

I tried to do the `return` on the reconnect() but that made the websockets hang and wasn't able to reconnect to the websocket server. Also when it hanged I couldn't shut the bot with CTRL+C for some reason 🤔

docs/connectors/twitch.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Configuration
+
+```yaml
+connectors:
+  twitch:
+    # required
+    code: "hfu923hfks02nd2821jfislf"  # Code obtained from the first oauth step
+    client-id: "e0asdj48jfkspod0284"
+    client-secret: "kdksd0458j93847j"
```
jacobtomlinson, line 19: I would maybe change the example values to something like `MYTWITCHCODE`, `MYCLIENTID`, `MYCLIENTSECRET`, etc.

docs/connectors/twitch.md
```diff
@@ -0,0 +1,78 @@
...
+    channel: theflyingdev # Broadcaster channel
+    redirect: http://localhost # Url to be passed to get oath token - defaults to localhost
+    foward-url: 'http://94jfsd9ff.ngrok.io' # Either an URL provided by a forwarding service or an exposed ip address
```
jacobtomlinson, line 22: In the GitHub connector rewrite I've added a `base-url` configuration option to the `web` config, so I would quite like to use that consistently between connectors. More and more connectors are moving to this model so it would be good to configure it in one place.

```yaml
web:
  base-url: https://url-to-opsdroid.com
```

docs/connectors/twitch.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Setup Twitch App
+
+You need to create a [Twitch App](https://dev.twitch.tv/console/apps) to use the Twitch Connector. Click the `+ Register Your Application` button, give this app a name and a redirect url - using `http://localhost` is fine. Once created, you can click the `Manage` button and get your `client-id`, you can then get a `client-secret` by pressing the `New Secret` button (keep this secret safe as it won't be shown to you again).
+
+## Getting OAuth code
+
+Twitch OAuth has two steps, first you need to make a `GET` request to a specific URL to obtain a `code`. After you've received the code, you need to make a `POST` request to the same URL and Twitch will send you an `access_token` and `refresh_token`. 
+
+After a certain period, the `access_token` will expire and you have to make a new request with your `refresh_token` to re-authenticate yourself.
```
jacobtomlinson, line 35: Does the user need to do this themselves, or can the connector handle this for them?

docs/connectors/twitch.md
```diff
@@ -0,0 +1,78 @@
...
+
+```
+https://id.twitch.tv/oauth2/authorize?client_id=<your client id>&redirect_uri=http://localhost&response_type=code&scope=channel:read:subscriptions+channel_subscriptions+analytics:read:games+chat:read+chat:edit+viewing_activity_read+channel_feed_read+channel_editor+channel:read:subscriptions+user:read:broadcast+user:edit:broadcast+user:edit:follows+channel:moderate+channel_read
+```
+
+## Usage
+
+The connector will subscribe to three webhooks - follower alert, stream status and subscribers. These webhooks allow Twitch to make a post request with data of certain actions, these actions will trigger events specific to the Twitch connector. By default, opsdroid will only connect to the chat service when the streamer is live, but you can set the optional configuration setting `always-listening` to true and make opsdroid connect to the chat whenever opsdroid is active.
```
jacobtomlinson, line 68: I don't think users will care about details like opsdroid subscribing to three webhooks. They will just case about the kind of events that will come down the connector.

docs/connectors/twitch.md
jacobtomlinson: User's don't care about how the tokens work or how opsdroid refreshes them. They just need simple steps to get set up.

FabioRosado:
Very good points haha will change things around and add some examples 😄 
The rst code is not working at the moment not sure how to make it work atm

docs/connectors/twitch.md
jacobtomlinson: To be clear this isn't done yet.

FabioRosado:
Btw after using digital ocean for a few days I've noticed that for some reason ubuntu keeps killing opsdroid, so I'm afraid websockets is giving us a memory leak of some sort.  Not sure how to fix now but might give it a go and try aiohttp websockets support, maybe its a library things - I think slack was using the same websockets lib? 🤔

jacobtomlinson:
Yeah the old slack connector had a memory leak. It seemed to happen when the socket was closed from the remote end and a reconnect happened. Not sure why. 

We ended up switching to the official Slack python library when they added asyncio support and we haven't seen the problem since. So might be worth looking at how they implemented the websockets support. [Looks like they use aiohttp `ws_connect`. ](https://github.com/slackapi/python-slackclient/blob/master/slack/rtm/client.py#L330)

FabioRosado:
That's what I thought as well, since it seems to be happening everytime the websocket gets closed by Twitch, so I thought it might be an issue with how I'm trying to reconnect or it's a websocket problem.

Thank you for pointing the slack library so I can see how they implement the reconnect will play around and try to solve the issue 👍

FabioRosado:
Apologies for all of these pushes, I've tested things out and the connector seems to work okay. Still going to test the websocket connect tomorrow on stream to see if I fixed all the issues. It will probably take me a while to write the tests for it and update the docs but I'm happy with the code so far.

Also still need to test the renewal logic on a long session of opsdroid.

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,738 @@
...
+        a user leaves the chat channel.
+        
+        """
+        _LOGGER.info(_("Connecting to Twitch IRC Server."))
+
+        async with aiohttp.ClientSession() as session:
+            async with session.ws_connect(f"{self.server}:{self.port}", heartbeat=600) as websocket:
+                self.websocket = websocket
+                await self.send_handshake()
+                await self.get_messages_loop()
```
FabioRosado, line 277: Apparently when I do `raise aiohttp.ServerDisconnectedError`  inside the `get_messages_loop` this will throw an "exception not caught", I would expect that the try/catch inside the `listen` method should handle that but doesn't seem like it.

So I might have to re-raise the same exception here, what do you think about it? 🤔

FabioRosado: Just another update, apparently trying to send `pong` to Twitch with `websocket.pong()` doesn't work because Twitch wants us to send it as a string. So changing the reply to `webscket.send_str("PONG")` seems to have fixed the issue of the connection dropping constantly around 6 minutes.

Now that makes me a bit afraid that the whole reconnect is still messed up and might not be able to catch the exception. But also makes me realise that probably the issue might have been happening because Twitch allowed us to reconnect to the service but then didn't get a proper reply and kind of put us on a timeout? So maybe the `aiohttp.ClientSession` request got an error message and since I'm not catching anything there it just stayed silet.

Anyway apologies for the rambling and I will need to test this further to see if we get that exception eventually. Will push the code to the digital ocean droplet and let it run for a while to test things out.

FabioRosado:
Apologies for having forced push this, but when I pulled the changed I showed up on all the previous commits so yeah 🤷‍♂️ 

I've finished writing the tests and they all pass but I will change them when you submit your github PR, figured that theres no point in doing the same mockapi work when you did it already 😄   I will also add the Twitch const vars to the connector class like you did in the github one but I figured that I could do that when I update the tests.

On the docs end, I've added some examples and tried to use the sphinx format to show up the classes for the connector and events - figured it would be easier to maintain the docs if we change the code itself and the docstrings 🤔 Not sure if they are working properly because I keep getting that error that I mentioned on matrix that it couldn't find opsdroid.connector.twitch.

I'm also not sure what that doc test failure means 
```Warning, treated as error:
/home/runner/work/opsdroid/opsdroid/opsdroid/connector/twitch/__init__.py:docstring of opsdroid.connector.twitch.ConnectorTwitch.twitch_webhook_handler:10:Inline interpreted text or phrase reference start-string without end-string.```

Finally, coverage is saying that some code on the tests is not being covered, do I need to do something to prevent coverage from looking into the test folder? 🤔 

This is a massive PR sorry for that D:

**Comment by FabioRosado** [comment]
Well I thought the issue with the docs(docstring) was because I split up that line, but that wasn't it. Can't really understand what 
```/home/runner/work/opsdroid/opsdroid/opsdroid/connector/twitch/__init__.py:docstring of opsdroid.connector.twitch.ConnectorTwitch.twitch_webhook_handler:10:Inline interpreted text or phrase reference start-string without end-string.```

this means. I thought it was a line that was missing punctuation but doesn't seem like it 🤔

File: opsdroid/connector/twitch/__init__.py [hunk 16]
```diff
@@ -0,0 +1,769 @@
...
+
+    async def twitch_webhook_handler(self, request):
+        """Handle event from Twitch webhooks.
+
+        This method will handle events when they are pushed to the webhook post route. Each webhook will
+        send a different kind of payload so we can handle each event and trigger the right opsdroid event
+        for the received payload.
+
+        For follow events the payload will contain `from_id`(broadcaster id), `from_username`(broadcaster username)
+        `to_id`(follower id), `to_name` (follower name) and `followed_at` (timestamp).
```
jacobtomlinson, line 370:
```suggestion
        `to_id`(follower id), `to_name`(follower name) and `followed_at`(timestamp).
```

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,769 @@
...
+
+        For stream changes a lot more things are returned but we only really care about `type`(if live/offline)
+        `title`(stream title).
```
jacobtomlinson, line 373:
```suggestion
        For stream changes a lot more things are returned but we only really care about 
        `type`(if live/offline) `title`(stream title).
```

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,769 @@
...
+        return aiohttp.web.Response(status=500)
+
+    async def twitch_webhook_handler(self, request):
+        """Handle event from Twitch webhooks.
+
+        This method will handle events when they are pushed to the webhook post route. Each webhook will
+        send a different kind of payload so we can handle each event and trigger the right opsdroid event
+        for the received payload.
+
+        For follow events the payload will contain `from_id`(broadcaster id), `from_username`(broadcaster username)
```
Cadair, line 369: @jacobtomlinson is correct these need to be double backticks, also I would put a space after the close backticks.

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,769 @@
...
+        `to_id`(follower id), `to_name`(follower name) and `followed_at`(timestamp).
+
+        For stream changes a lot more things are returned but we only really care about `type`(if live/offline)
+        `title`(stream title).
+
+        For subscriptions events we will want to know `event_type`, `timestamp`, `event_data.plan_name`, `event_data.is_gift`,
+        `event_data.tier`, `event_data.username` and `event_data.gifter_name`.
```
Cadair: Try this:
```suggestion
        For follow events the payload will contain ``from_id`` (broadcaster id), ``from_username`` (broadcaster username)
        ``to_id`` (follower id), ``to_name`` (follower name) and ``followed_at`` (timestamp).

        For stream changes a lot more things are returned but we only really care about ``type`` (if live/offline)
        ``title`` (stream title).

        For subscriptions events we will want to know ``event_type``, ``timestamp``, ``event_data.plan_name``, ``event_data.is_gift``,
        ``event_data.tier``, ``event_data.username`` and ``event_data.gifter_name``.
```

FabioRosado:
Got to say, adding two backticks and a space its a very annoying thing haha but it's fine 😛  I just need to get used to that, I've added two backticks on all of them and also added the space, lets see if that will fix it.

Cadair:
in rst single backtick is a (or can be) a cross-reference.

FabioRosado:
Ah okay makes sense, and know that I know why it was complaining it makes sense why it was saying start-end issues. Good to see that the thing passes now haha

opsdroid/connector/twitch/__init__.py
```diff
@@ -0,0 +1,769 @@
...
+
+        valid = await self.validate_request(request, self.webhook_secret)
+        payload = await request.json()
+
+        if valid:
+            try:
+                [data] = payload.get("data")
+                _LOGGER.debug(_("Got event from Twitch - %s") % data)
+
+                if data.get("followed_at"):
```
Cadair, line 394: have you considered using our handy [`EventCreator`](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/events.py#L19) to replace this list of ifs?

docs/connectors/twitch.md
```diff
@@ -0,0 +1,265 @@
...
+### Events Available
+
+The Twitch Connector contains 10 events that you can use on your custom made skill. Some of these events are triggered automatically whenever an action happens on twitch - for example when a user follows your channel. Others you will have to trigger on a skill - for example, to delete a specific message.
+
+#### Automatic Events
+
+These events are triggered by opsdroid whenever something happens on twitch.
+
+```eval_rst
+.. autoclass:: opsdroid.connector.twitch.events.UserJoinedChat
```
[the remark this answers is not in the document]
FabioRosado: You know what? I didn't even think about checking the ones that we already have implemented in opsdroid 🤦  But yeah that makes much more sense!  I am going to update this in a second and submit an update

docs/connectors/twitch.md
```diff
@@ -0,0 +1,265 @@
...
+    :members:
+```
+
+```eval_rst
+.. autoclass:: opsdroid.connector.twitch.events.CreateClip
+    :members:
+```
+
+```eval_rst
+.. autoclass:: opsdroid.connector.twitch.events.DeleteMessage
```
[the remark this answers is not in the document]
FabioRosado: Sounds good, do you think I should add this event together with this PR? I guess that would make sense right? 🤔

opsdroid/connector/twitch/__init__.py
FabioRosado: Again I didn't even remember about that but yeah that would be a better way to do it haha I am going to have a look on how you use the EventCreator in matrix and check how to update that huge list of ifs 😁

docs/connectors/twitch.md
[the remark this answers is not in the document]
Cadair: yeah, we have a bunch of core events which haven't grown support in more than one connector at the moment. Basically, we added things there if it felt like we could use them elsewhere.

opsdroid/connector/twitch/__init__.py
Cadair: we also use it in slack as well (although it's a little more hacky there).

jacobtomlinson: I'm never sure how I feel about the `EventCreator`. It introduces a large amount of [indirection](https://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection) and I often struggle debugging in that area because you have to track so many function calls and pointers in your head.

FabioRosado: What do you think it might be the best option here? I think on the stream started we need to keep it as it is since I am triggering the listening method on when the user started streaming 🤔  

I guess I could add one for the follow and sub events 🤔

Cadair: At it's core the event creator is a dictionary of `type: function` with a helper to know where to read type from. I do agree that the slack implementation isn't great due to the way the slack API thing works, but by the time we got there we were pretty done with that PR lol.

I am interested in better ways we could do that, I personally have a pretty strong dislike of a couple of hundred lines of if statements.

jacobtomlinson: > I personally have a pretty strong dislike of a couple of hundred lines of if statements

Yup I also agree with this. Hence why I'm on the fence about patterns like this.

jacobtomlinson: I would potentially argue that the horrible long if statement is very readable compared to the event creator. And making things accessible to beginner developers is a core principal of opsdroid.

But I'd love to find a third option.

opsdroid/events.py
```diff
@@ -478,3 +482,14 @@ class PinMessage(Event):
 
 class UnpinMessage(Event):
     """Event to represent unpinning a message or other event."""
+
+
+class DeleteMessage(Event):
+    """Event to represent deleting a message or other event."""
+
+    def __init__(self, id, *args, **kwargs):  # noqa: D107
```
Cadair, line 490: I think `id` should just be `linked_event`?

FabioRosado: sounds good, in the Twitch case, we need to pass the message id to delete it.  I'm assuming that the linked_event represents the whole Message event right?

opsdroid/connector/twitch/__init__.py
FabioRosado: I took a stab at EventCreator but doesn't seem like a great way to use in this case. One of the biggest issues with it is the `dispatch_key` - Twitch will not use a single key that we can use to create events from,  I've tried sending data as it is, the content of the list inside data and even tried to add a default key and use that, but it didn't work.

I also got to be honest and say that I don't fully understand the EventCreator code. When trying to create the event just from the `UserFollowed` and passing the `followed_at` key as `dispatch_key` it seems that this event was never created anyway.

If I figure out how to use the EventCreator we could potentially reduce the amount of if when receiving a subscriber event - but the other events will either have to stay as they are or create another event creator to handle the other types which seems it would be just double work imo.

Cadair: fair enough.

jacobtomlinson:
So glad to see this go in. It has been a lot of work. Many thanks @FabioRosado.

PR #1499, opened 2020-05-25
regex_matcher: add findall matching_conditions

silvio:
# Description

The difference between "search" / "fullmatch" and "findall" is the
returning object. The first two return a match-object and the latter
returns a list of strings. The idea is to create a "dictonary" from this
and return it.

## Status
**UNDER DEVELOPMENT**

## Type of change

could be:

# How Has This Been Tested?

see in <https://github.com/silvio/azure-devops-opsdroid>

- New entry tested: functional
- old entry `@match_parse` also tested: functional as before

# Checklist:

### DIFF
--- a/opsdroid/parsers/regex.py
+++ b/opsdroid/parsers/regex.py
@@ -23,10 +23,29 @@ async def match_regex(text, opts):
 
     if opts["matching_condition"].lower() == "search":
         matched_regex = regex.search(opts["expression"], text, is_case_sensitive())
+        if matched_regex != None:
+            matched_regex = matched_regex.groupdict()
     elif opts["matching_condition"].lower() == "fullmatch":
         matched_regex = regex.fullmatch(opts["expression"], text, is_case_sensitive())
+        if matched_regex != None:
+            matched_regex = matched_regex.groupdict()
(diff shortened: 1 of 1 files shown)


silvio:
Can I run the regular expression tests on my own?

silvio:
Not needed anymore

jacobtomlinson:
Would be interested to know why you don't need this any more. Can you give us an update?

silvio:
@jacobtomlinson The use case is that I want to find on many similar elements of a regex expression. My current approach is to trigger to only one (the first one) and then execute the actual finding process in the skill.

jacobtomlinson:
@silvio It sounds like you are running up against the limitations of regex. This is why we support NLU matchers like Rasa in opsdroid. Doing statistical NLP matching can help you create more advanced and robust skills.

silvio:
@jacobtomlinson Thanks for the pointer. Have you an example for me, a skill which is using this?

jacobtomlinson:
@silvio I'd start with the Rasa docs https://docs.opsdroid.dev/en/stable/skills/matchers/rasanlu.html

Issue #1529, opened 2020-06-16
Improvements to the Matrix Connector now that it's using nio


Cadair:
I am opening this as a placeholder for a bunch of smaller tasks that need investigating or doing after the merge of #1418, as is usual with a large PR somethings got delayed to follow ups.

- [x] Handle all cases of (send) methods returning a `nio.Error` instance where previously errors would be raised. Proposed solution: write a decorator which converts a return Error to a raised error. (https://github.com/opsdroid/opsdroid/pull/1418#discussion_r436615260)

- [ ] Handle multiple events in one sync response. (https://github.com/opsdroid/opsdroid/pull/1418#discussion_r435176024)

- [x] Investigate generalising the logging of errors from nio in non-send methods. Could potentially use the same solution as the send methods above. (https://github.com/opsdroid/opsdroid/pull/1418#discussion_r432383617)

- [x] Ensuring 100% test coverage and test improvements, it's clear from the refactoring work that a bunch of our tests are not great tests of the behaviour only good tests of the code. Given nio provides clients which do no no io :wink: maybe we could look at developing tests based on nio classes?

- [x] User documentation! We just completely changed the type of `ConnectorMatrix.connection` we need to document it.

- [ ] Provide some information in the logs if events don't decrypt properly, and if we can get it out of nio, why they didn't decrypt.

- [ ] Improve logging of some transport errors with sync, currently they show as unknown errors.

Also this rewrite needs a lot of testing before we release it onto the world.

Tyagdit:
- [ ] Sending files using URLs doesn't work correctly, returns a TypeError which needs fixing
- [ ] Initial sync is being rate limited for some reason, needs some fiddling with sync tokens or device IDs
- [ ] Receiving membership events isn't proper

Cadair:
I have moved the remaining three issues to #1712 #1713 and #1714

PR #1538, opened 2020-06-16
Fix File/Image and membership events in Matrix Connector

Tyagdit:
# Description

Sending files/images and receiving membership events works. Updated matrix-nio version and removed redundant ClientConfig

Part of fixes listed in  #1529 

## Status
**UNDER DEVELOPMENT** 

## Type of change

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -1,16 +1,12 @@
 """Connector for Matrix (https://matrix.org)."""
 
 import re
-import json
 import logging
 import functools
-from concurrent.futures import CancelledError
 from urllib.parse import urlparse
 
 import aiohttp
 
-from matrix_api_async.api_asyncio import AsyncHTTPAPI
-from matrix_client.errors import MatrixRequestError
 from voluptuous import Required
 
 from opsdroid.connector import Connector, register_event
@@ -20,6 +16,8 @@ from .html_cleaner import clean
 from .create_events import MatrixEventCreator
 from . import events as matrixevents
 
+import nio
+import json
 
 _LOGGER = logging.getLogger(__name__)
 CONFIG_SCHEMA = {
@@ -47,7 +45,13 @@ def ensure_room_id_and_send(func):
             event.target = self.room_ids[event.target]
 
         if not event.target.startswith("!"):
-            event.target = await self.connection.get_room_id(event.target)
+            response = await self.connection.room_resolve_alias(event.target)
+            if isinstance(response, nio.RoomResolveAliasError):
+                _LOGGER.error(
+                    f"Error resolving room id for {event.target}: {response.message} (status code {response.status_code})"
+                )
+            else:
+                event.target = response.room_id
 
         try:
             return await func(self, event)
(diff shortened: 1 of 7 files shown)


opsdroid/connector/matrix/connector.py
```diff
@@ -405,11 +404,14 @@ async def _file_to_mxc_url(self, file_event):
                 mxc_url = file_event.url
 
         if not mxc_url:
+            upload_file = await file_event.get_file_bytes()
+            if isinstance(upload_file, bytes):
+                upload_file = BufferedReader(BytesIO(upload_file))
```
Cadair, line 409: I am not very familiar with the `io` classes, but what is `BufferedReader` providing here? Am I correct in thinking that `BytesIO` already provides a file-like input and that's all we need to pass to `nio`?

tests/test_connector_matrix.py
```diff
@@ -633,21 +637,21 @@ async def test_respond_image(self):
...
 
             patched_send.assert_called_once_with(
+                room_id="!test:localhost",
+                message_type="m.room.message",
                 content={
                     "body": "opsdroid_upload",
                     "info": {"w": 1, "h": 1, "mimetype": "image/gif", "size": 26},
                     "msgtype": "m.image",
-                    "url": "mxc://aurl",
+                    "url": {"content_uri": "mxc://aurl"},
```
Cadair, line 653: I don't think this change is correct. The form of the content dict here is just as described in the spec: https://matrix.org/docs/spec/client_server/r0.6.0#m-image unless I am missing something?

tests/test_connector_matrix.py
```diff
@@ -688,21 +692,21 @@ async def test_respond_file(self):
...
 
             patched_send.assert_called_once_with(
+                room_id="!test:localhost",
+                message_type="m.room.message",
                 content={
                     "body": "opsdroid_upload",
                     "info": {},
                     "msgtype": "m.file",
-                    "url": "mxc://aurl",
+                    "url": {"content_uri": "mxc://aurl"},
```
Cadair, line 708: same here.

opsdroid/connector/matrix/connector.py
Tyagdit: BytesIO is a file-like but the upload function expects `TextIOBase, BufferedReader, BufferedRandom, FileIO, AsyncBufferedReader or AsyncTextIOWrapper` and `bytes` cant be directly casted to these types

Cadair: Oh, what types it actually supports in `io` are really bizzare to me, I might ask in the nio room.

It might actually be easier to pass it a callable, which returns the bytes directly, I think this will work: `upload_file = lambda x, y: upload_file`.

Cadair:
Looks good, can you look into the test fails and bump the coverage till codecov is happy?

requirements.txt
```diff
@@ -9,7 +9,7 @@ certifi==2020.4.5.2
 click==7.1.2
 emoji==0.5.4
 mattermostdriver==7.0.1
-matrix-nio==0.10.0
+matrix-nio==0.13.0
```
Cadair, line 12:
```suggestion
matrix-nio==0.14.0
```

tests/test_connector_matrix.py
```diff
@@ -342,6 +393,21 @@ async def test_parse_sync_response(self):
             )
             assert returned_message.raw_event == raw_message
 
+            returned_message = await self.connector._parse_sync_response(
```
Cadair, line 396: assert the type of `returned_message` here?

Issue #1546, opened 2020-06-23
GitHub Connector uses deprecated query params when retrieving authenticated user


nkuik:
# Description

The GitHub connector uses a deprecated query params method to authenticate to GitHub for retrieving the username. Besides being less-than-ideal in sending the token in the query, this approach will eventually be unsupported by GitHub. 

## Steps to Reproduce

1. Setup the configuration for using the opsdroid GitHub connector (by adding the token in `configuration.yaml`).
2. Start opsdroid.
3. Wait for an email from GitHub notifying that a deprecated way of using the `user` API was used (if the token you used is connected to your GitHub account.

Here is an example of GitHub's warning email:

<img width="1401" alt="Screenshot 2020-06-23 at 11 28 56" src="https://user-images.githubusercontent.com/14183156/85394074-4cdf4a80-b54e-11ea-8ab8-0c6b80d260c5.png">

## Expected Functionality

Opsdroid should authenticate to GitHub without a deprecated method of doing so.

## Experienced Functionality

When starting up Opsdroid with the GitHub connector, I later received an email from GitHub warning that my GitHub token had been used with the GitHub API using a deprecated authentication strategy.

## Versions
- **Opsdroid version:** 0.19.0
- **Python version:** 3.8.3
- **OS/Docker version:** macOS Catalina (10.15.3)

## Configuration File

```yaml
skills:
  hello: {}

connectors:
  github:
    token: <my token>
```

## Additional Details

None.

jacobtomlinson:
Thanks for raising this. I made a start on rewriting this connector recently but lost me branch due to a corrupt SSD. I intend on taking another run at it soon.

I noticed the same issue, the authentication method we are using has been deprecated. It would be great to get this updated.

PR #1547, opened 2020-06-23
Make additional GitHub api request before making user request

nkuik:
# Description

This is a fix to a deprecated way of determining the authenticated GitHub user that is implemented by the Opsdroid GitHub connector. Instead of trying to retrieve the user in one request to the GitHub API (using the access token in query params), there are now two requests that are made:

1. One to authenticate with the token provided in `configuration.yaml`
2. An additional request to determine the username of the user authenticated in the first request

Fixes: #1546 

## Status
**READY**

I think?

## Type of change

# How Has This Been Tested?

I have not added any additional tests, but I have modified an existing test so that it continues to pass.

- `tests/test_connector_github.py::TestConnectorGitHubAsync::test_connect_failure`

# Checklist:

- [X] I have made corresponding changes to the documentation (not applicable)
- [X] I have added tests that prove my fix is effective or that my feature works (not applicable)

### DIFF
--- a/opsdroid/connector/github/__init__.py
+++ b/opsdroid/connector/github/__init__.py
@@ -32,16 +32,20 @@ class ConnectorGitHub(Connector):
 
     async def connect(self):
         """Connect to GitHub."""
-        url = "{}/user?access_token={}".format(GITHUB_API_URL, self.github_token)
-        async with aiohttp.ClientSession(trust_env=True) as session:
-            response = await session.get(url)
-            if response.status >= 300:
-                _LOGGER.error(_("Error connecting to GitHub: %s."), response.text())
+        url = "{}".format(GITHUB_API_URL)
+        headers = {"Authorization": "token {}".format(self.github_token)}
+        async with aiohttp.ClientSession(trust_env=True, headers=headers) as session:
+            auth_response = await session.get(url)
+            if auth_response.status >= 300:
+                response_text = await auth_response.text()
+                _LOGGER.error(_("Error connecting to GitHub: %s."), response_text)
                 return False
+            user_url = "{}/user".format(GITHUB_API_URL)
+            user_response = await session.get(user_url)
             _LOGGER.debug(_("Reading bot information..."))
-            bot_data = await response.json()
-        _LOGGER.debug(_("Done."))
-        self.github_username = bot_data["login"]
+            bot_data = await user_response.json()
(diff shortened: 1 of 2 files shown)


tests/test_connector_github.py
```diff
@@ -4,12 +4,10 @@
 
 import asyncio
 import unittest
-import unittest.mock as mock
```
nkuik, line 7: Remove unused imports.

opsdroid/connector/github/__init__.py
```diff
@@ -32,16 +32,20 @@ def __init__(self, config, opsdroid=None):
 
     async def connect(self):
         """Connect to GitHub."""
-        url = "{}/user?access_token={}".format(GITHUB_API_URL, self.github_token)
-        async with aiohttp.ClientSession(trust_env=True) as session:
-            response = await session.get(url)
-            if response.status >= 300:
-                _LOGGER.error(_("Error connecting to GitHub: %s."), response.text())
+        url = "{}".format(GITHUB_API_URL)
+        headers = {"Authorization": "token {}".format(self.github_token)}
```
nkuik, line 36: f-Strings might be preferred here.

jacobtomlinson: Yeah we would prefer f-strings moving forwards.

opsdroid/connector/github/__init__.py
```diff
@@ -32,16 +32,20 @@ def __init__(self, config, opsdroid=None):
...
+        async with aiohttp.ClientSession(trust_env=True, headers=headers) as session:
+            auth_response = await session.get(url)
+            if auth_response.status >= 300:
+                response_text = await auth_response.text()
+                _LOGGER.error(_("Error connecting to GitHub: %s."), response_text)
```
jacobtomlinson, line 41: What is the purpose of this request? Just to check auth? You could do the same with the `/user` call and reduce the number of requests made when connecting.

nkuik: I thought I initially had issues with doing the single request to `/user`, but I was mistaken. I switched it back to the single request.

jacobtomlinson:
Awesome work many thanks!

For future reference please do not force-push to your branch when adding extra commits. It makes it hard for the reviewer to see what has changed. Just a regular push will be fine.

PR #1548, opened 2020-06-23
Add E2EE functionality to matrix connector

### LINKED ISSUE #992: Add support for end to end encryption in Matrix Connector
I tried opsdroid in a docker container to connect to a matrix server. So far everything works fine. But I realized that opsdroid does react in unencrypted rooms only. I assume that there is no end-to-end-encryption support built in the matrix connector. Is there a possibility to enable encryption in opsdroid?

Tyagdit:
# Description

Adds steps needed to send and receive encrypted messages through matrix. The listen method basically mimics `AsyncClient.sync_forever`. Few additions to the configuration which we could find alternatives for. 

Fixes #992 and some parts of #1529 

## Status
 **UNDER DEVELOPMENT**

## TODO

- [x] Received encrypted events have slightly different formats, needs some tweaking
- [x] Config validation
- [x] `sync` sometimes returns older events which need to be filtered somehow, might be fixed by handling multiple events in a single sync
- [ ] Rarely, received events stop getting decrypted, possibly due to some conflict in the store, needs further investigation
- [ ] Add guide/script for installing libolm
- [x] Add tests
- [x] Add docs

## Type of change

## Checklist:


opsdroid/connector/matrix/connector.py
```diff
@@ -214,13 +252,8 @@ async def _parse_sync_response(self, response):
             if roomInfo.timeline:
                 for event in roomInfo.timeline.events:
                     if event.sender != self.mxid:
-                        # if isinstance(event, nio.MegolmEvent):
-                        # event is encrypted
-                        # event = await self.connection.decrypt_event(event)
```
Cadair, line 219: Is this done for us elsewhere?

opsdroid/connector/matrix/connector.py
```diff
@@ -241,6 +274,17 @@ async def listen(self):  # pragma: no cover
 
             _LOGGER.debug(_("Matrix sync request returned."))
 
+            await self.connection.send_to_device_messages()
+
+            if self.connection.should_upload_keys:
+                await self.connection.keys_upload()
+            if self.connection.should_query_keys:
+                await self.connection.keys_query()
+            if self.connection.should_claim_keys:
```
Cadair, line 283: Should this not also be done as part of the initial sync? If so This could be moved into a helper method?

opsdroid/connector/matrix/connector.py
```diff
@@ -145,6 +159,11 @@ async def connect(self):
             )
             return
 
+        if not self.device_id:
```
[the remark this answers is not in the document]
Tyagdit: It would but if we use a newly created device the device verification wouldn't work since the store doesn't get populated

opsdroid/connector/matrix/connector.py
Tyagdit: The sync function decrypts the events

opsdroid/connector/matrix/connector.py
```diff
@@ -24,9 +24,11 @@
     Required("mxid"): str,
     Required("password"): str,
     Required("rooms"): dict,
+    Required("store_path"): str,
```
Cadair, line 27: let's default this to [`DEFAULT_ROOT_PATH`](https://github.com/opsdroid/opsdroid/blob/6d7a256e88c88c2b7ae9524c5a3c41ebb5848d8a/opsdroid/const.py#L11)`/matrix` for now.

opsdroid/connector/matrix/connector.py
```diff
@@ -177,6 +197,24 @@ async def connect(self):
 
         self.connection.sync_token = response.next_batch
 
+        # TODO: Device verification
+        # for user in self.verify_users:
+        #    for olm_device in mapi.device_store[user].values():
+        #        if not olm_device.verified:
+        #            mapi.verify_device(olm_device)
```
Cadair, line 204: Probably best to just drop this and we can come back to it.
```suggestion
```

opsdroid/connector/matrix/connector.py
```diff
@@ -323,6 +367,7 @@ async def _send_message(self, message):
             self._get_formatted_message_body(
                 message.text, msgtype=self.message_type(message.target)
             ),
+            ignore_unverified_devices=True,
```
Cadair, line 370: leaving this as an instance variable is probably a good idea. I can see that this will likely become a flag in the future. Could be private (`_`)?

opsdroid/connector/matrix/connector.py
```diff
@@ -74,7 +76,7 @@ def __init__(self, config, opsdroid=None):  # noqa: D107
         self.room_ids = {}
         self.default_target = self.rooms["main"]["alias"]
         self.mxid = config["mxid"]
-        self.nick = config.get("nick", None)
+        self.nick = config.get("nick")
```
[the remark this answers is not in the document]
Tyagdit: `config.get()` would return `None` by default anyway

opsdroid/connector/matrix/connector.py
```diff
@@ -83,6 +85,8 @@ def __init__(self, config, opsdroid=None):  # noqa: D107
         self.filter_id = None
         self.connection = None
         self.device_name = config.get("device_name", "opsdroid")
+        self.device_id = config.get("device_id")
```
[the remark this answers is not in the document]
Tyagdit: We need a device_id that belongs to the user, so we'd have to get it from the config anyway

opsdroid/connector/matrix/connector.py
```diff
@@ -83,6 +86,11 @@ def __init__(self, config, opsdroid=None):  # noqa: D107
...
+        self.store_path = config.get(
+            "store_path", str(Path(const.DEFAULT_ROOT_PATH).joinpath("matrix"))
+        )
+        self.__ignore_unverified__ = True
```
Cadair, line 93: Just a single underscore is convention for private variables. `__thing__` is generally only used for python internal methods etc.
```suggestion
        self._ignore_unverified = True
```

opsdroid/connector/matrix/connector.py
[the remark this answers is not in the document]
Cadair: What do you mean when you say "belongs to the user"? I am pretty sure that the HS will generate a random device ID if one isn't specified at login.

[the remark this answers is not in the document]
Cadair: See the response section of https://matrix.org/docs/spec/client_server/r0.5.0#post-matrix-client-r0-login

opsdroid/connector/matrix/connector.py
Cadair: In which case we should delete this block.

opsdroid/connector/matrix/connector.py
```diff
@@ -145,6 +165,11 @@ async def connect(self):
             )
             return
 
+        if not self.device_id:
```
Cadair, line 168: By the time we make it here if the user hasn't given a device id, we have one assigned to us. Does this get added to the store? Could we save it with the rest of the store and restore it next time opsdroid is started, this would at least reduce the amount of devices we create?

Tyagdit: I couldn't find a way to access the store through nio without first creating a client object and creating it assigns a new device_id

opsdroid/connector/matrix/connector.py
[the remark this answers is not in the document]
Tyagdit: What I meant was if we were to use a device id that already exists as a default, that id would have to be one that belongs to the user's account.

opsdroid/connector/matrix/connector.py
```diff
@@ -83,6 +85,11 @@ def __init__(self, config, opsdroid=None):  # noqa: D107
         self.filter_id = None
         self.connection = None
         self.device_name = config.get("device_name", "opsdroid")
+        self.device_id = config.get("device_id", "OPSDROID")
```
Cadair, line 88:
```suggestion
        self.device_id = config.get("device_id", "opsdroid")
```

opsdroid/connector/matrix/connector.py
```diff
@@ -133,8 +140,23 @@ async def make_filter(self, api, fjson):
...
-        config = nio.AsyncClientConfig(encryption_enabled=False, pickle_key="")
-        mapi = nio.AsyncClient(self.homeserver, self.mxid, config=config)
+
+        if not Path(self.store_path).is_dir():
+            Path(self.store_path).mkdir()
+
+        config = nio.AsyncClientConfig(
+            encryption_enabled=True,
+            pickle_key="",
+            store=nio.store.SqliteStore,
```
Cadair, line 150: why sqlite and not default? What's the advantages / disadvantages of both?

opsdroid/connector/matrix/connector.py
```diff
@@ -133,8 +140,23 @@ async def make_filter(self, api, fjson):
...
+            store_name="test_store",
```
Cadair, line 151:
```suggestion
            store_name="opsdroid",
```
why not name this something like this?

opsdroid/connector/matrix/connector.py
```diff
@@ -177,13 +201,28 @@ async def connect(self):
...
         if self.nick:
             display_name = await self.connection.get_displayname(self.mxid)
             if display_name != self.nick:
                 await self.connection.set_displayname(self.nick)
 
     async def disconnect(self):
         """Close the matrix session."""
+        if self.device_id == "OPSDROID":
+            await self.connection.logout()
+            Path(self.store_path, "test_store").unlink()
```
Cadair, line 225: Why? Why not leave it there for next time?

opsdroid/connector/matrix/connector.py
```diff
@@ -133,8 +141,22 @@ async def make_filter(self, api, fjson):
...
-        config = nio.AsyncClientConfig(encryption_enabled=False, pickle_key="")
-        mapi = nio.AsyncClient(self.homeserver, self.mxid, config=config)
+
+        if self._allow_encryption and not Path(self.store_path).is_dir():
+            Path(self.store_path).mkdir()
+
+        config = nio.AsyncClientConfig(
+            encryption_enabled=self._allow_encryption,
+            pickle_key="",
+            store_name="opsdroid.db" if self._allow_encryption else "",
```
Cadair, line 151: does this store have any value if e2ee isn't enabled?

opsdroid/connector/matrix/connector.py
```diff
@@ -177,6 +201,19 @@ async def connect(self):
 
         self.connection.sync_token = response.next_batch
 
+        if self._allow_encryption:
```
Cadair, line 204: Does this break if encryption is disabled? If not can we just continue calling it?

opsdroid/connector/matrix/connector.py
Tyagdit: nope it doesn't get created at all

opsdroid/connector/matrix/connector.py
Tyagdit: i dont think it breaks anything

opsdroid/connector/matrix/connector.py
Cadair: then do we need this conditional?

opsdroid/connector/matrix/connector.py
Cadair: We might as well save the if then?

opsdroid/connector/matrix/connector.py
Tyagdit: if we pass a non-empty string it tries to load the store which then errors out

Cadair: Only if encryption is disabled?

Tyagdit: yeah

tests/test_connector_matrix.py
```diff
@@ -213,11 +217,11 @@ def filter_json(self):
...
-    def setUp(self):
-        """Basic setting up for tests"""
-        self.connector = setup_connector()
-        self.api = nio.AsyncClient("https://notaurl.com", None)
-        self.connector.connection = self.api
+    #    def setUp(self):
+    #        """Basic setting up for tests"""
+    connector = setup_connector()
+    api = nio.AsyncClient("https://notaurl.com", None)
+    connector.connection = api
```
Cadair, line 224: These should be a fixture so they are recreated new for each test.

tests/test_connector_matrix.py
```diff
@@ -249,6 +253,20 @@ async def test_connect(self):
...
+            api_string.format("keys_upload")
+        ) as patched_keys_upload, amock.patch(
+            api_string.format("keys_query")
+        ) as patched_keys_query, amock.patch(
+            "pathlib.Path.mkdir"
+        ) as patched_mkdir, amock.patch(
+            "pathlib.Path.is_dir"
+        ) as patched_is_dir, amock.patch(
             "aiohttp.ClientSession"
         ) as patch_cs, OpsDroid() as _:
```
Cadair, line 271: oh my this list of mocks has got out of control :rofl: 

Do you fancy refactoring this to use the mocker fixture? You can make fixtures which use mocker to mock things or groups of things and then pass them to tests, so for this you could mock all the e2ee related things in one fixture like this:

```python
def e2ee_mock(mocker):
    mocker.patch(api_string.format("send_to_device_messages"))
    ....
```
the only thing with this approach is returning the patched object for later assertions, maybe a dict of `{'api_string': patched}` would be the best way to handle that.

Or if that's a little too hacky, you could at least use the mocker fixture in this test to move all of these from being context managers to function calls which would make it read a lot nicer.

tests/test_connector_matrix.py
```diff
@@ -1165,12 +1334,13 @@ async def test_send_generic_event(self):
...
-        self.connector = setup_connector()
-        self.api = nio.AsyncClient("https://notaurl.com", None)
-        self.connector.connection = self.api
+@pytest.mark.asyncio
+class TestEventCreatorMatrixAsync:
+    """Basic setting up for tests"""
+
+    connector = setup_connector()
+    api = nio.AsyncClient("https://notaurl.com", None)
+    connector.connection = api
```
Cadair, line 1343: Again needs to be a fixture (probably could be the same fixture).

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py36, py37, py38, lint, docker
```
Cadair, line 2: @jacobtomlinson We have had to do some hacks here to test the e2ee functionality while waiting for upstream to provide builds. Do you have thoughts on this and the changes to the github actions config file?

jacobtomlinson: This feels very matrix specific. Does this mean I will have to run something like `tox -e py38-noe2e` when testing locally?

Cadair:
> Why are we switching from ubuntu-latest to ubuntu-20.04?

We need olm 3 which is only in the 20.04 version of the GH action VMs.

> How much does olm add to the Docker image?

I don't know but it shouldn't be large, and it's worth it for ease of use for matrix.

> Is there a way to avoid renaming the tox envs?

Working on it.

tox.ini
Cadair: It seems to be a limitation of the gh-actions tox plugin. I have got to what I think is a working compromise, where on the CI we run `noe2e` on windows to override the behaviour of the tox plugin, but locally all `py3x` commands should work just fine.

SolarDrew:
No further questions yer honour :+1:

PR #1558, opened 2020-07-03
Improve test machinery and rewrite GitHub connector tests

jacobtomlinson:
# Description

This is a big one sorry folks! I started rewriting the GitHub connector tests and anded up building a whole `opsdroid.testing` submodule to help make tests nicer.

## `opsdroid.testing`
Created `opsdroid.testing` submodule with fixtures and utilities for testing. Top level fixtures are now imported from here by `conftest.py` so they can also be used by others. Also contains the following utilities.

### `ExternalAPIMockServer`

Created `ExternalAPIMockServer` utility for mocking external APIs (see https://www.youtube.com/watch?v=MDI1X6B2Ya8). The general idea with this class is to allow you to push expected responses onto a stack for each API call you expect your test to make. Then as your tests make those calls each response is popped from the stack.

Here is some example usage from the docs.

```python
import pytest
import aiohttp

from opsdroid.testing import ExternalAPIMockServer

@pytest.mark.asyncio
async def test_example():
    # Construct the mock API server and push on a test method
    mock_api = ExternalAPIMockServer()
    mock_api.add_response("/test", "GET", None, 200)

    # Create a closure function. We will have our mock_api run this concurrently with the
    # web server later.
    async def test_closure():
        # Make an HTTP request to our mock_api
        async with aiohttp.ClientSession() as session:
            async with session.get(f"{mock_api.base_url}/test") as resp:

                # Assert that it gives the expected responses
                assert resp.status == 200
                assert mock_api.called("/test")

        # Return True to we can ensure the closure was run
        return True

    # Run the test closure. This will start the web server, and once it is running await
    # the closure. Then when the closure finishes the server will be stopped again and the
    # return from the closure will be passed back.
    assert await mock_api.run_test(test)
```

### `run_unit_test`

Added a helper which runs opsdroid and a test closure together at the same time, then shuts down and cleans up opsdroid after the test finishes.

Example from the docs

```python
import pytest
from opsdroid.testing import (
    opsdroid,
    run_unit_test,
    MINIMAL_CONFIG
    )

@pytest.mark.asyncio
async def test_example(opsdroid):
    # Using the opsdrid fixture we load it with the
    # minimal example config
    await opsdroid.load(config=MINIMAL_CONFIG)

    # Check that opsdroid is not currently running
    assert not opsdroid.is_running()

    # Define an awaitable closure which asserts that
    # opsdroid is now running
    async def test():
        assert opsdroid.is_running()
        return True  # So that we can assert our run test

    # Run our closure against opsdroid. This will start opsdroid,
    # await our closure and then stop opsdroid again.
    assert await run_unit_test(opsdroid, test)
```

### `call_endpoint`

Utility which takes a running opsdroid and calls an API endpoint with a predefined request. Useful alongside `run_unit_test` to fake external webhook requests. The nice thing here is you don't need to know what address opsdroid is running at.

Example from the docs.

```python
import pytest
from opsdroid.testing import (
    opsdroid,
    call_endpoint,
    run_unit_test,
    MINIMAL_CONFIG
    )

@pytest.mark.asyncio
async def test_example(opsdroid):
    # Using the opsdrid fixture we load it with the
    # minimal example config
    await opsdroid.load(config=MINIMAL_CONFIG)

    async def test():
        # Call our endpoint by just passing
        # opsdroid, the endpoint and the method
        resp = await call_endpoint(opsdroid, "/stats", "GET")

        # Make assertions that opsdroid responded successfully
        assert resp.status == 200
        return True

    assert await run_unit_test(opsdroid, test)
```

## Core refactors

I've refactored some of the core to make things more testable. Notably I've split `load` and `unload` unto `load`, `start`, `stop` and `unload`. I've also broken some functionality out into functions like `register_skill` to give better entrypoints in the tests.

## GitHub Connector test rewrite

I've rewritten all of the GitHub connector tests using these new tools and refactors to make them more readable and also less brittle. 

Take this example. 

- We mock out the API auth
- Write a skill inline that is full of assertions
- Register the skill with opsdroid
- Write a test closure which hits the connector pretending to be a GitHub webhook
- Run everything through together

```python
@pytest.mark.asyncio
async def test_receive_issue(opsdroid, connector, mock_api):
    """Test a PR create event creates a message and parses it."""
    mock_api.add_response(
        "/user", "GET", get_response_path("github_user.json"), status=200
    )

    @match_event(Message)
    async def test_skill(opsdroid, config, event):
        assert event.connector.name == "github"
        assert event.text == "test"
        assert event.target == "opsdroid/opsdroid#740"

    opsdroid.register_skill(test_skill, config={"name": "test"})

    async def test():
        resp = await call_endpoint(
            opsdroid,
            "/connector/github",
            "POST",
            data=get_webhook_payload("github_issue_payload.json"),
        )
        assert resp.status == 201

    await mock_api.run_test(run_unit_test, opsdroid, test)
```

The result of this last line means the Mock API server will be started. Then opsdroid will be started which will call the mock API. Then our test closure will be run which will hit the connector and trigger opsdroid to run the skill, the skill will assert things and then everything will clean up.

# Checklist:


opsdroid/testing/__init__.py
```diff
@@ -0,0 +1,380 @@
...
+                # Construct the mock API server and push on a test method
+                mock_api = ExternalAPIMockServer()
+                mock_api.add_response("/test", "GET", None, 200)
+
+                # Create a closure function. We will have our mock_api run this concurrently with the
+                # web server later.
+                async def test_closure():
+                    # Make an HTTP request to our mock_api
+                    async with aiohttp.ClientSession() as session:
+                        async with session.get(f"{mock_api.base_url}/test") as resp:
```
Cadair, line 53: where is this `mock_api` coming from here?

Cadair: doesn't that need to passed in as a fixture on line 43?

Cadair: oh :facepalm:

jacobtomlinson:
@FabioRosado I've only added GET and POST because that's all I needed for my tests. Happy to add the rest at some point.

opsdroid/testing/__init__.py
```diff
@@ -0,0 +1,380 @@
...
+                routes = [web.get(route, self._handler)]
+            elif method.upper() == "POST":
+                routes = [web.post(route, self._handler)]
+            else:
+                raise TypeError(f"Unsupported method {method}")
+
+            self.responses[route] = [(status, response)]
+            self.app.add_routes(routes)
+
+    async def run_test(
```
Cadair, line 142: From a discussion in chat, this could be made a context manager which would lead to nicer tests imo.

jacobtomlinson: Agreed!

Cadair:
Bad stale bot!

cognifloyd (dismissed):
Could you re-run the GHA? I think there's a race condition somewhere with the windows GHA runners. I don't think those errors are caused by this PR.

cognifloyd:
Please add `sphinx-autodoc-typehints` to the `extras_requires.docs` in `setup.cfg`. That should fix the docs build.
https://github.com/opsdroid/opsdroid/blob/7f036c3199a99628c0efb4d6ef142ab8f5d9a3c1/setup.cfg#L128-L133

jacobtomlinson:
Triggered CI

jacobtomlinson:
Thanks so much for picking this up @cognifloyd! Do you want to do the honour of merging once CI passes (don't worry about the coverage drop here).

cognifloyd (approved):

cognifloyd:
OK. In my final review, I realized that I missed updating the mock API server's start method like I did with web.start. So, I added that fix including a unit test for that. Once I got that unit test passing, I also marked the mock broken skill so it gets skipped in code coverage reports.

Now, even codecov.io should be happy. As soon as all tests are green, I'll merge :smiley:

Issue #1553, opened 2020-07-01
Switch Pyup bot to nightlies


jacobtomlinson:
Instead of hard pinning dependencies in order to trigger CI to test new releases we should soft pin and introduce nightly builds to test new dependencies. 

The downside of this is that an unrelated PR could discover the breakage before the nightly runs. Which isn't a great experience for newbies. 

The plus side is installing opsdroid would be more pleasant and it would reduce some maintenance burden of reviewing and merging PyUp bot PRs.

Cadair:
We should write a bot to report scheduled CI fails to the chat so we don't miss them when they break, if only we knew of a good bot framework :thinking: :rofl:

PR #1567, opened 2020-07-07
Add optional dependencies by moving requirements to setup.cfg

Tyagdit:
# Description

Moved most of the params for `setuptools.setup()` to `setup.cfg` so we can start adding optional dependencies for each module in the form of extras. An `[all]` extra is created in setup.py which has deps for all the modules. Moved test deps to `tox.ini` which uses the `[all]` extra

## Status
**UNDER DEVELOPMENT**

## TODO
- [x] Add extras for each module
- [x] Add grouped extras

## Type of change

# Checklist:


tox.ini
```diff
@@ -15,8 +15,23 @@ setenv =
 commands =
      py.test -v --timeout=30 --cov=opsdroid --cov-report term-missing:skip-covered --cov-report=xml:cov.xml {posargs}
 deps =
-     -r{toxinidir}/requirements.txt
-     -r{toxinidir}/requirements_test.txt
+	flake8==3.8.1
```
Cadair, line 18: I think these should be an extra named `test` or similar? (Which isn't installed by `all`).

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+test_suite = tests
+setup_requires = babel
+install_requires =
+	aiohttp==3.6.2
+	aioredis==1.3.1
+	aiosqlite==0.13.0
+	appdirs==1.4.4
+	arrow==0.15.6
+	Babel==2.8.0
+	bleach==3.1.5
```
Cadair, line 95: This is matrix.

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+	certifi==2020.4.5.2
+	click==7.1.2
+	emoji==0.5.4
+	mattermostdriver==7.0.1
+	motor==2.1.0
+	multidict==4.7.6
+	nbconvert==5.6.1
+	nbformat==5.0.6
+	opsdroid-get-image-size==0.2.2
```
Cadair, line 104: this is also matrix.

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+	parse==1.15.0
+	puremagic==1.9
+	pycron==1.0.0
+	pyyaml==5.3.1
+	setuptools==46.2.0
```
Cadair, line 109: Do we actually have a runtime requirement on setuptools?

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+	click==7.1.2
+	emoji==0.5.4
+	mattermostdriver==7.0.1
+	motor==2.1.0
+	multidict==4.7.6
+	nbconvert==5.6.1
+	nbformat==5.0.6
+	opsdroid-get-image-size==0.2.2
+	parse==1.15.0
+	puremagic==1.9
```
Cadair, line 106: I *think* this is also matrix?

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+	certifi==2020.4.5.2
+	click==7.1.2
+	emoji==0.5.4
+	mattermostdriver==7.0.1
+	motor==2.1.0
+	multidict==4.7.6
+	nbconvert==5.6.1
+	nbformat==5.0.6
+	opsdroid-get-image-size==0.2.2
+	parse==1.15.0
```
Cadair, line 105: matcher-parse

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+	aioredis==1.3.1
+	aiosqlite==0.13.0
+	appdirs==1.4.4
+	arrow==0.15.6
+	Babel==2.8.0
+	bleach==3.1.5
+	certifi==2020.4.5.2
+	click==7.1.2
+	emoji==0.5.4
+	mattermostdriver==7.0.1
```
Cadair, line 99: connector-mattermost I assume

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+[options]
+include_package_data = True
+zip_safe = False
+packages = find:
+test_suite = tests
+setup_requires = babel
+install_requires =
+	aiohttp==3.6.2
+	aioredis==1.3.1
+	aiosqlite==0.13.0
```
Cadair, line 91: database-sqlite

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+
+[options]
+include_package_data = True
+zip_safe = False
+packages = find:
+test_suite = tests
+setup_requires = babel
+install_requires =
+	aiohttp==3.6.2
+	aioredis==1.3.1
```
Cadair, line 90: database-redis ??

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+	aiosqlite==0.13.0
+	appdirs==1.4.4
+	arrow==0.15.6
+	Babel==2.8.0
+	bleach==3.1.5
+	certifi==2020.4.5.2
+	click==7.1.2
+	emoji==0.5.4
```
Cadair, line 98: connector-slack?!

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+	multidict==4.7.6
+	nbconvert==5.6.1
+	nbformat==5.0.6
+	opsdroid-get-image-size==0.2.2
+	parse==1.15.0
+	puremagic==1.9
+	pycron==1.0.0
+	pyyaml==5.3.1
+	setuptools==46.2.0
+	slackclient==2.6.2
```
Cadair, line 110: connector-slack

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+	tailer==0.4.1
+	ibm-watson==4.4.1
```
Cadair, line 112: matcher-watson

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+	websockets==8.1
+	webexteamssdk==1.3
```
Cadair, line 114: connector-webexteams ?

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
...
+	voluptuous==0.11.7
+	regex==2020.5.7
```
Cadair, line 116: matcher-regex

setup.cfg
```diff
@@ -39,3 +39,96 @@ output_dir = opsdroid/locale
 [compile_catalog]
 domain = opsdroid
 directory = opsdroid/locale
+
+[metadata]
```
Cadair, line 43: I would suggest we put this at the top

tox.ini
Tyagdit: I thought since all tests are gonna be using tox anyway this would reduce some clutter in setup.py

setup.cfg
Tyagdit: Doesn't look like it, same with websockets i think. The websockets connector uses aiohttp

tox.ini
Cadair: You don't have to run the tests in tox, so it's good to provide an easy way for people to install the requirements outside of tox.

setup.cfg
Cadair: I could well be wrong.

setup.cfg
Cadair: I would be ok with that, be interested to know what the others think.

Cadair: Maybe it would be best to do this with the `common` group @jacobtomlinson mentioned in chat, so it's possible to have a very very light opsdroid (maybe just using the webhook connector for instance) but we make all our default installation documentation be `pip install opsdroid[common]`.

Cadair:
@Tyagdit can you rebase this? We should try and get this merged now.

setup.cfg
Cadair: Removing this should fix the build
```suggestion
```

Cadair:
This PR can easily be made to fix https://github.com/opsdroid/opsdroid/issues/1553 as well. We should remove all the version pins in here and add a cron job: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule

Dockerfile
```diff
@@ -8,7 +8,6 @@ COPY opsdroid opsdroid
 COPY setup.py setup.py
 COPY versioneer.py versioneer.py
 COPY setup.cfg setup.cfg
-COPY requirements.txt requirements.txt
 COPY README.md README.md
 COPY MANIFEST.in MANIFEST.in
```
Cadair, line 13: I assume we want the docker file to install all the deps? If we do line 17 below needs updating to be `.[all]`

jacobtomlinson: Yeah the docker image should have everything. Maybe we should also have a slim image which just installs the core too.

jacobtomlinson:
I also just disabled pyup-bot and set dependabot to check for security updates only.

setup.cfg
```diff
@@ -1,3 +1,130 @@
...
+	pytest-cov>=2.7.1
+	pytest-mock>=3.2.0
+	pytest-timeout>=1.4.0
+	pydocstyle>=5.0.2
+	asynctest>=0.13.0
+	mypy-lang>=0.5.0
+	sphinx>=3.0.4
+	sphinx-click>=2.3.2
+	recommonmark>=0.6.0
+	deadlinks>=0.3.2
```
Cadair, line 125: it would probably be good to move these into a "docs" requirement, and use that over "test" for read the docs. Installing less things on RTD means a faster build.

setup.cfg
```diff
@@ -1,3 +1,130 @@
...
+    Environment :: Console
+    Framework :: AsyncIO
+    Intended Audience :: Developers
+    Intended Audience :: System Administrators
+    Intended Audience :: Information Technology
+    License :: OSI Approved :: Apache Software License
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: 3.6
```
Cadair, line 23:
```suggestion
```

setup.cfg
Tyagdit: so the config should just use these 4?

Cadair: no test should be all the others, and these should be docs.

Tyagdit: yea i meant the RTD CI check

Tyagdit: seems like pytest also needs to be in docs?

Cadair: looks like yeah

Cadair (approved):

docs/installation.md
```diff
@@ -50,6 +50,18 @@ C:\Users\myaccount> opsdroid config gen | out-file "configuration.yaml" -encodin
...
+
+You can individually select modules to install dependencies for by specifying them
+in the square brackets. For instance, to install the redis database and webex connector
+module dependencies in addition to the opsdroid core package:
+```
+pip install opsdroid[database_redis,connector_webex]
+```
+
+`[common]` includes the matrix and slack connectors and sqlite database modules
+You can also use `[all]`, `[all_connectors]`, `[all_databases]`, `[all_parsers]` and `[test]`
```
jacobtomlinson, line 63: Can we link to the setup file here or something so users can discover all the available options?

docs/installation.md
```diff
@@ -7,7 +7,7 @@
 ## Quickstart
 
 ```bash
-$ pip3 install opsdroid
+$ pip3 install opsdroid[common]
```
jacobtomlinson, line 10: Should the quickstart install common? Or just core?

jacobtomlinson (changes requested):
Thanks for this. I've added a couple of small queries.

Also what happens in the case that a user doesn't install a module but configures it. For example if they run `pip install opsdroid[common]` and then tryies to configure the `redis` database. I assume they get an error but I also assume it's not a very helpful one.

If we are making modules optional we should propbably be catching the import errors and logging advice on how to install them.

docs/installation.md
Tyagdit: So add a comment block on top of setup.cfg and link to the file here?

docs/installation.md
Tyagdit: I think since its for getting the users set up and try out opsdroid quickly, common would be a better choice

docs/installation.md
jacobtomlinson: Not necessarily. We just need to list all the options somewhere. The `options.extras_require` section in setup.cfg lists all of the options (other than the meta ones that you already listed in the docs). So it might be useful to link there.

docs/installation.md
jacobtomlinson: But common only gives Slack, Matrix and sqlite right? And the quickstart docs do not mention those at all.

I expect many users will be unfamiliar with extra packages specified this way. Opsdroid it aimed at novice users. I just feel like this is unnecessary complexity at this point.

Perhaps this is a sign that we should rethink the quickstart. What do we actually want to show people here?

Cadair:
Note to future selves, this probably needs calling out in the release notes.

jacobtomlinson:
🎉🎉🎉🎉🎉

PR #1577, opened 2020-07-12
Add a default option to memory.get, port memory tests to pytest and add testing documentation

Cadair:
# Description

This one always catches me out, I expect `get()` to have `dict.get()` like semantics and be able to specify a default.

## Status
**UNDER DEVELOPMENT**

## Type of change

# Checklist:


Cadair:
Ok, this got a little out of hand.

I have added some new docs `maintaining/testing.rst` which uses https://github.com/astropy/sphinx-automodapi/ to automatically generate API docs for the whole `opsdroid.testing` submodule. This currently only works when using an rst file, so an rst file it is. Given there is little in there currently, it's not really needed but I am sure that module will expand a lot soon.

Cadair:
Thanks to the RTD build you can see the built testing docs here: https://opsdroid--1577.org.readthedocs.build/en/1577/maintaining/testing.html

.github/workflows/ci.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        component: [docker, docs, lint]
+        component: [docker, lint]
```
Cadair, line 44: I have removed this from here, because RTD will build and preview the docs now I have modified the webhooks to trigger it.

docs/maintaining/testing.rst
```diff
@@ -0,0 +1,57 @@
+Testing
+=======
+
+When making a contribution to opsdroid, it's important that tests are written to validate that the functionality works, and continues to work in the future.
+
+opsdroid is currently slowly migrating from `unittest <https://docs.python.org/3.8/library/unittest.html>`__ to `pytest <https://docs.pytest.org>`__ as it's testing framework.
```
jacobtomlinson, line 6: Can we link to the issue here and also add todo item to the issue to remove this when done.

opsdroid/testing/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
+Testing helpers for opsdroid.
+
+opsdroid provides a set of pytest fixtures and other helpers for writing tests
+for both opsdroid core and skills.
+"""
+from .helpers import *  # noqa
```
jacobtomlinson, line 7: Why have a separate `helpers` file?

docs/maintaining/testing.rst
```diff
@@ -0,0 +1,57 @@
...
+
+The ``mocker`` fixture is provided by pytest-mock, and provides convenient access to things in the mock library, as well as automatic teardown of patches added with ``mocker.patch``.
+
+
+opsdroid test helpers (`opsdroid.testing`)
+------------------------------------------
+
+.. automodapi:: opsdroid.testing
+   :headings: -#
+   :no-heading:
```
jacobtomlinson, line 57: Is this the reason why you had to use RST here? 

I had a read of the docs for `automodapi` but struggled to see what it does differently to automodule.

I'd also like to understand more about why this wouldn't work inside an ` ```eval_rst``` ` block in markdown?

docs/maintaining/testing.rst
```diff
@@ -0,0 +1,57 @@
+Testing
```
jacobtomlinson, line 1: Opsdroid *currently* uses markdown for all documentation. I'm happy to open up discussions around changing that, but I'd rather not introduce inconsistencies here without a change of policy.

docs/maintaining/index.md
```diff
@@ -8,4 +8,5 @@
    overview
    supported-python-versions
    making-a-release
+   testing
```
jacobtomlinson, line 11: I would argue that this isn't for maintainers, but instead it is for contributors.

Perhaps the contributing docs need breaking down into a few sections like the maintaining ones are?

docs/conf.py
```diff
@@ -35,6 +35,7 @@
     "sphinx.ext.autodoc",
     "sphinx.ext.napoleon",
     "sphinx_click.ext",
+    "sphinx_automodapi.automodapi",
```
jacobtomlinson, line 38: Has this dependency been captured anywhere?

docs/maintaining/testing.rst
Cadair: automodapi generates a complete list of all the functions, classes and variables and generates all the API documentation for all of the objects in the module. `.. automodule` only captures the module docstring.

it appears to fail in an `eval_rst` because of the lack of interlinking support, as the main include works but none of the links to the object documentation work.

docs/maintaining/testing.rst
Cadair: How's a policy on "use markdown, unless a sphinx extension breaks it".

docs/conf.py
Cadair: it's in the requirements.txt ?

opsdroid/testing/__init__.py
Cadair: I was thinking it would be good to have the externalapi helper and other larger blocks of code in a separate file to more core fixtures etc.?

Cadair: Maybe this should be in a `fixtures` file :thinking:

docs/maintaining/testing.rst
Cadair: done.

opsdroid/testing/__init__.py
jacobtomlinson: 👍 on a fixtures file

docs/maintaining/testing.rst
jacobtomlinson: Ah ok. So kinda like the option to automatically include methods when using `autoclass`.

I'm surprised by this. `eval_rst` within a markdown document and a regular RST file should both result in the same AST as they use the same parser. Then the same docs should be generated.

I've not seen issues like this with `autodoc` which I assume would suffer from the same problems.

Cadair: I didn't fall deep down the pit of working it out, I swapped it to rst to see if it fixed it and it did, then I pushed it ;)

docs/maintaining/testing.rst
jacobtomlinson: In this situation it sounds like the problem is with the extension. If an extension has a bug which results in not working correctly with `eval_rst` then I suggest we do not use the extension.

docs/conf.py
jacobtomlinson: Ah I didn't see it in there.

docs/maintaining/testing.rst
Cadair: Looking at the generated source, something must be funky, because the HTML isn't generating a link. So it's not that it's generating a broken link, sphinx isn't generating the link at all...

opsdroid/testing/__init__.py
Cadair: done.

docs/maintaining/testing.rst
Cadair: The bug has been patched in https://github.com/astropy/sphinx-automodapi/pull/112. I have removed automodapi from this for now.

docs/contributing/testing.md
```diff
@@ -0,0 +1,74 @@
...
+
+opsdroid is currently slowly migrating from
+[unittest](https://docs.python.org/3.8/library/unittest.html) to
+[pytest](https://docs.pytest.org) as it\'s testing framework. This work
+is tracked in issue
+[\#1502](https://github.com/opsdroid/opsdroid/issues/1502). As a part of
+this migration tests can be found in one of two places, old unittest
+tests are found in the `tests/` directory in the root of the repository
+and new pytest tests are found in a `tests/` directory in the same
+subfolder as the code, i.e `opsdroid/tests/`.
```
jacobtomlinson, line 15: There appear to be some errant `\`s going on in here.

jacobtomlinson: Also why so many line breaks?

docs/contributing/testing.md
```diff
@@ -0,0 +1,74 @@
...
+actions after a `yield` statement.
+
+The `mocker` fixture is provided by pytest-mock, and provides convenient
+access to things in the mock library, as well as automatic teardown of
+patches added with `mocker.patch`.
+
+## opsdroid test helpers
+
+```eval_rst
+.. automodule:: opsdroid.testing
```
jacobtomlinson, line 68: Should this have `:members:` and `:imported-members:`?

docs/contributing/testing.md
Cadair: oh I just let pandoc loose on it. :confused:

docs/contributing/testing.md
Cadair: I decided to do it manually to be able to add subtitles.

docs/contributing/testing.md
Cadair: fixed.

Cadair:
Yay! Thanks.

jacobtomlinson:
Thanks for bearing with me being stubborn and going down this long rabbit hole 😆

PR #1587, opened 2020-07-22
Add matrix database

Tyagdit:
# Description

Rewrote the [matrix database](https://github.com/SolarDrew/database-matrix) with matrix-nio. Added a delete method. Added encryption support for the database. When encrypting data is sent as encrypted room events in encrypted rooms whose event id is stored as state events. Added migration from opsdroid.database to dev.opsdroid.database event. Needs some more testing with encrypted rooms.

## Status
**UNDER DEVELOPMENT**

## Type of change

# Checklist:


opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+                "Error putting key into matrix room {self.room_id}: {ori_data.message}({ori_data.status_code})"
+            )
+            return
+        ori_data = ori_data.content
+
+        if self._single_state_key:
+            value = {key: value}
+
+        elif not isinstance(value, dict):
+            raise ValueError("When single_state_key is False value must be a dict.")
```
Cadair, line 99: We should probably expand this message to make it clear that `single_state_key` is a config option.

```suggestion
            raise ValueError("When the matrix database is configured with single_state_key=False the value stored in the database must be a dict.")
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+        )
+
+        _LOGGER.debug(f"Putting {key} into matrix room {self.room_id}.")
+
+        ori_data = await self.connector.connection.room_get_state_event(
+            room_id=self.room_id, event_type=self._event_type, state_key=state_key,
+        )
+        if isinstance(ori_data, RoomGetStateEventError):
+            _LOGGER.error(
+                "Error putting key into matrix room {self.room_id}: {ori_data.message}({ori_data.status_code})"
```
Cadair, line 90:
```suggestion
                f"Error putting key into matrix room {self.room_id}: {ori_data.message}({ori_data.status_code})"
```

opsdroid/configuration/example_configuration.yaml
```diff
@@ -237,6 +237,10 @@ connectors:
 #    port:       "12345"       # (optional) default "27017"
 #    database:   "mydatabase"  # (optional) default "opsdroid"
 #
+#  matrix:
+#    default_room: "main"
+#    single_state_key: "dev.opsdroid.database"
```
Cadair, line 242: This should be a bool right?
```suggestion
#    single_state_key: false
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+
+    @property
+    def connector(self):
+        """Define the initiated matrix connector."""
+        return self.opsdroid.get_connector("matrix")
+
+    @property
+    def room_id(self):
+        """Define ID of the room to store data in."""
+        return (
```
Cadair, line 36: Could we use the connector `lookup_target` method here? https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/matrix/connector.py#L294

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+        data = await self.connector.connection.room_get_state(room_id=self.room_id)
+        if isinstance(data, RoomGetStateError):
+            _LOGGER.error(
+                f"Error migrating from opsdroid.database to {self._event_type} in room {self.room_id}: {data.message}({data.status_code})"
+            )
+            return
+
+        if not any(event["type"] == self._event_type for event in data.events):
+            for event in data.events:
+                if event["type"] == "opsdroid.database":
```
Cadair, line 54: We are going to get here at every startup of the database right, just that the content of the state event (once it's been replaced) would be empty. I think you need this to be:

```suggestion
                if event["type"] == "opsdroid.database" and event["content"]:
```

You should also add a test that the new state event and the redaction isn't sent on a second run of this method.

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+                "the value passed must be a dict."
+            )
+
+        sub_key = list(value.keys())[0]
+
+        if list(value.keys())[0] in ori_data:
+            if self._single_state_key:
+                existing_val = await self.get(sub_key)
+            else:
+                existing_val = await self.get({key: sub_key})
```
Cadair, line 110: Haven't we already retrieved the value in `ori_data` above? Given the encryption support I can understand needing to replace the call to `room_get_state_event` with a call to `self.get` to handle the decryption, however, I think that we should only ever need a `self.get(key)` to do that?

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+
+        return True
+
+    async def get(self, key):
+        """Get a value from the database for a given key."""
+
+        if self.should_migrate:
+            await self.migrate_database()
+
+        if not self._single_state_key and not isinstance(key, dict):
```
Cadair, line 143: I don't think this is correct.

If `single_state_key` is False then we use `key` as the `state_key`, if `single_state_key` is True then we do `value = event["content"][key]`?

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+                "When the matrix database is configured with single_state_key=False, key must be a dict."
+            )
+
+        # If the single state key flag is set then use that else use state key.
+        state_key = (
+            "" if self._single_state_key is True else self._single_state_key or key
+        )
+
+        if isinstance(state_key, dict):
+            state_key, key = list(state_key.items())[0]
```
Cadair, line 154: I don't understand how state_key can ever be a dict.

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+        if self._single_state_key:
+            value = {key: value}
+
+        elif not isinstance(value, dict):
+            raise ValueError(
+                "When the matrix database is configured with single_state_key=False, "
+                "the value passed must be a dict."
+            )
+
+        sub_key = list(value.keys())[0]
```
Cadair, line 104: this doesn't make any sense to me, if `single_state_key` is `False` then there can be an arbitrary number of keys in `value`.

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+                existing_val = await self.get({key: sub_key})
+            if existing_val == value[sub_key]:
+                _LOGGER.debug("Not updating matrix state, as content hasn't changed.")
+                return
+
+        if self.should_encrypt:
+            room_event = await self.opsdroid.send(
+                GenericMatrixRoomEvent(content=value, event_type=self._event_type)
+            )
+            data = {**ori_data, sub_key: {"encrypted_val": room_event.event_id}}
```
Cadair, line 119: surely this should be `key` not `sub_key`?

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+                f"Error getting {key} from matrix room {self.room_id}: {data.message}({data.status_code})"
+            )
+            return
+        data = data.content
+
+        _LOGGER.debug(f"Got {data} from state request.")
+
+        for k, v in data.items():
+            if isinstance(v, dict) and "encrypted_val" in v:
+                resp = await self.connector.connection.room_get_event(
```
Cadair, line 174: We probably need to at least log something here if we fail to decrypt this event because of missing keys. (i.e. it was sent a long time ago with a different instance of opsdroid that can no longer decrypt it etc)

tests/test_database_matrix.py
```diff
@@ -0,0 +1,571 @@
...
+    db = DatabaseMatrix({}, opsdroid=opsdroid_matrix)
+    db.should_migrate = False
+    await db.put("twim", {"hello": "world"})
+
+    patched_send.assert_has_calls(
+        [
+            matrix_call(
+                "GET",
+                "/_matrix/client/r0/rooms/%21notaroomid/state/dev.opsdroid.database/",
+            ),
```
Cadair, line 71: I think there is an issue with the tests here, that we should only be making this call once, but from the implementation I have to assume we are making it more than once. (Same goes for all the tests)

tests/test_database_matrix.py
```diff
@@ -0,0 +1,571 @@
...
+                "PUT",
+                "/_matrix/client/r0/rooms/%21notaroomid/state/dev.opsdroid.database/",
+                {"twim": "hello", "pill": "red"},
+            ),
+        ],
+    )
+
+
+@pytest.mark.asyncio
+async def test_put_encrypted(patched_send, opsdroid_matrix, mocker):
```
Cadair, line 317: I think we need a single_state_key variant of this test, with more than one item in the dict. (probably good to have a couple of tests with more than one item in the dict).

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+        if list(value.keys())[0] in ori_data:
+            if self._single_state_key:
+                existing_val = await self.get(sub_key)
+            else:
+                existing_val = await self.get({key: sub_key})
+            if existing_val == value[sub_key]:
+                _LOGGER.debug("Not updating matrix state, as content hasn't changed.")
+                return
+
+        if self.should_encrypt:
```
Cadair, line 115: We should only be doing this if e2ee is enabled in the room and `should_encrypt` is set.
I am pretty sure you can tell if a room is encrypted by the presence of a certain state event.

```suggestion
        if self.room_is_encrypted(self.room) and self.should_encrypt:
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,285 @@
...
+        ori_data = await self.connector.connection.room_get_state_event(
+            room_id=self.room_id, event_type=self._event_type, state_key=state_key,
+        )
+        if isinstance(ori_data, RoomGetStateEventError):
+            _LOGGER.error(
+                f"Error getting the state event of type {self._event_type} with "
+                f"state_key '{state_key}' from matrix room {self.room_id}: "
+                f"{ori_data.message}({ori_data.status_code})"
+            )
+            return
```
Cadair, line 89: Given my [comment](https://github.com/opsdroid/opsdroid/pull/1587/files#r474656706) below, shouldn't this be 
```suggestion
            return {}
```

Cadair: Also, this isn't always an error, if this is being triggered by the first PUT in a room, meaning logging this at error is probably a bad idea under all situations.

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,307 @@
...
+        elif not isinstance(value, dict):
+            raise ValueError(
+                "When the matrix database is configured with single_state_key=False, "
+                "the value passed must be a dict."
+            )
+
+        data = await self._verify_and_get(state_key, value)
+        if data is None:  # dict could be empty
+            _LOGGER.error("Error putting key into matrix room")
+            return
```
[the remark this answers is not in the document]
Cadair: This line: https://github.com/opsdroid/opsdroid/pull/1587/files#diff-c97d0db15ae0cc89347831db7cc93d05R89 causes a `None` to be returned if the state event doesn't exist?

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,307 @@
...
+        )
+
+        if not self._single_state_key:
+            if isinstance(state_key, dict):
+                state_key, key = list(state_key.items())[0]
+            else:
+                key = None
+
+        _LOGGER.debug(
+            f"Getting {key or 'full state'} from matrix room {self.room_id} with state_key={state_key}."
```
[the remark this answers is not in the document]
Cadair:
```suggestion
            f"Getting {key} from matrix room {self.room_id} with state_key={state_key}."
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,285 @@
...
+                        content=event["content"],
+                        state_key=event["state_key"],
+                    )
+                    await self.connector.connection.room_redact(
+                        room_id=self.room_id, event_id=event["event_id"],
+                    )
+
+        self.should_migrate = False
+
+    async def _verify_and_get(self, state_key, value_dict):
```
Cadair, line 77: I think this method could use a little readjustment, both to make it's error handling more applicable and to make it equally useful to `put` and `get`.

I think that restricting this to just getting an existing value out of the store and decrypting it if needed would be better than worrying about deduplicating with a new value, as that behaviour only belongs in `put` not `get`?

I would then call it something like `_get_and_decrypt`?

opsdroid/database/matrix/__init__.py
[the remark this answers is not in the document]
Tyagdit: if the state event doesn't exist, the `room_get_state_event` in `_verify_and_get` would return a default response with the error keys we discussed. So for the first put in a room `data` won't be None

opsdroid/database/matrix/__init__.py
Tyagdit: We won't get a `RoomGetStateEventError` for the first put in the room, that would happen for other problems like connection errors or membership errors etc

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,260 @@
...
+        )
+
+        _LOGGER.debug(
+            f"Getting {key} from matrix room {self.room_id} with state_key={state_key}."
+        )
+
+        data = await self.connector.connection.room_get_state_event(
+            room_id=self.room_id, event_type=self._event_type, state_key=state_key,
+        )
+        if isinstance(data, RoomGetStateEventError):
```
Cadair, line 153: We need to detect the 404 here like we discussed, I think
```python
if isinstance(data, RoomGetStateEventError) or (isinstance(data, RoomGetStateEventResponse) and data.transport_response.status == 404):
```
should catch all the issues.

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,260 @@
...
+            _LOGGER.error(
+                f"Error getting {key} from matrix room {self.room_id}: {data.message}({data.status_code})"
+            )
+            return
+
+        if data.transport_response.status == 404:
+            data.content.clear()
```
Cadair, line 160: The code I suggested above is more future proof than this, because now I have patched nio in the next release the 404 will return the error object.

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,260 @@
...
+                content=data,
+                target=self.room_id,
+                connector=self.connector,
+                state_key=state_key,
+            )
+        )
+
+        return True
+
+    async def get(self, key, check_dict=None):
```
Cadair, line 135: it feels weird to delegate the `check_dict` to `get`. Why not just put that code in `put`?

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,260 @@
...
+
+        if data.transport_response.status == 404:
+            data.content.clear()
+        data = data.content
+
+        _LOGGER.debug(f"Got {data} from state request.")
+
+        check = check_dict or data
+
+        for k, v in data.items():
```
Cadair, line 167: Aren't we now decrypting all events? Shouldn't we need to do something like this:
```
if self._single_state_key:
    data = {key: data[key]}
```
before this loop, so we only decrypt the keys we need if single state key is true?

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,260 @@
...
+            )
+            return
+
+        if data.transport_response.status == 404:
+            data.content.clear()
+        data = data.content
+
+        _LOGGER.debug(f"Got {data} from state request.")
+
+        check = check_dict or data
```
Cadair, line 165: This variable isn't being used anywhere?

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,260 @@
...
+                data[k] = resp.event.source["content"][k]
+
+        try:
+            if self._single_state_key and not check_dict:
+                data = data[key]
+        except KeyError:
+            _LOGGER.debug(f"{key} doesn't exist in database")
+            return None
+
+        return data
```
Cadair, line 187: I think with my suggestion above this can just be replaced with:
```suggestion
        if self._single_state_key:
            return data.get(key)
        return data
```

opsdroid/database/matrix/__init__.py
Cadair: Actually we should detect if the key is not in the return here for single state key
```python
if self._single_state_key:
    if key not in data:
        return None
    data = {key: data[key]}
```

opsdroid/database/matrix/__init__.py
Cadair, line 187: With my above comments this becomes
```python
return data
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,260 @@
...
+            data = {**data, **value}
+
+        if self.is_room_encrypted and self.should_encrypt:
+            for k, v in value.items():
+                room_event = await self.opsdroid.send(
+                    GenericMatrixRoomEvent(
+                        target=self.room_id, content={k: v}, event_type=self._event_type
+                    )
+                )
+                data[k] = {"encrypted_val": room_event.event_id}
```
Cadair, line 119: When I said in matrix that we should only send one event per put what I think I meant is this:
```suggestion
            room_event = await self.opsdroid.send(
                GenericMatrixRoomEvent(
                    target=self.room_id, content=value, event_type=self._event_type
                )
            )
            data[k] = {"encrypted_val": room_event.event_id}
```

docs/databases/matrix.md
```diff
@@ -0,0 +1,50 @@
...
+the dict in it. If single state key is disabled then the key passed should be a
+dict of the form `{state_key: key}`. Pass a list of keys to delete multiple keys.
+
+The database provides a context manager `memory_in_room` to perform some
+operations in a different rooms. use it as:
+```
+db = opsdroid.get_database("matrix")
+with db.memory_in_room(new_room_id):
+	database ops
+```
```
Cadair, line 50:
```suggestion

## Encryption

In encrypted Matrix rooms, state events (used by the database) are not encrypted. This means that anything put into the matrix database in an encrypted room would not be encrypted. To prevent this the matrix database send the **values** you put into your database into the room as regular events, which are encrypted, and then references these events from the database (which is still a state event).

This has two effects you should be aware of:
1) The keys in your database are not encrypted, **do not put sensitive information in the key**.
2) If opsdroid can't decrypt the event, it will be dropped from the database. This means that if you need long term storage in your encrypted rooms you must take steps to save the matrix connectors store (where the decryption keys are kept), so back up your store and keep it safe!
```
```
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+                if isinstance(resp, RoomGetEventError):
+                    _LOGGER.error(
+                        f"Error decrypting {v['encrypted_val']} while getting "
+                        f"{key}: {resp.message}({resp.status_code})"
+                    )
+                    continue
+                data[k] = resp.event.source["content"][k]
+
+        if get_full:
+            return {**ori_data.content, **data}
```
Cadair, line 183: Doing

```
data = ori_data.content
```
above did a copy by reference, which means that assigning to `data[k]` has modified `ori_data.content` so I am not sure if this line is doing anything? (i.e. return data would be the same).

Tyagdit: It's to account for the reassignment in [line 162](https://github.com/opsdroid/opsdroid/pull/1587/files#diff-c97d0db15ae0cc89347831db7cc93d05R162) so that when ssk is true we send the entire dict back to put

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+            )
+        elif ori_data.transport_response.status == 404:
+            _LOGGER.error(
+                f"Error getting {key} from matrix room {self.room_id}: Event not found"
+            )
+            return
+
+        data = ori_data.content
+
+        _LOGGER.debug(f"Got {data} from state request.")
```
Cadair, line 159:
```suggestion
        _LOGGER.debug(f"Got {data} from state in room {self.room_id}.")
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+
+        if self._single_state_key:
+            if key in data:
+                data = {key: data[key]}
+            elif not get_full:
+                _LOGGER.debug(f"{key} doesn't exist in database")
```
Cadair, line 165:
```suggestion
                _LOGGER.debug(f"{key} doesn't exist in state event in room {self.room_id}.")
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+                return
+            else:
+                return data
+
+        for k, v in data.items():
+            if isinstance(v, dict) and "encrypted_val" in v:
```
Cadair, line 171: We could be slightly stricter here and do:
```suggestion
            if isinstance(v, dict) and len(v) == 1 and "encrypted_val" in v:
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+                resp = await self.connector.connection.room_get_event(
+                    room_id=self.room_id, event_id=v["encrypted_val"],
+                )
+                if isinstance(resp, RoomGetEventError):
+                    _LOGGER.error(
+                        f"Error decrypting {v['encrypted_val']} while getting "
```
Cadair, line 177:
```suggestion
                        f"Error decrypting event {v['encrypted_val']} while getting "
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+                    continue
+                data[k] = resp.event.source["content"][k]
+
+        if get_full:
+            return {**ori_data.content, **data}
+
+        return data[key]
+
+    async def delete(self, key):
+        """Delete a key from the database."""
```
Cadair, line 189: A according to the docstring in the base class key should only be a string (not list of strings), so we should call that out here (it needs a parameter blocks).

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+        if isinstance(data, RoomGetStateEventError):
+            _LOGGER.error(
+                f"Error deleting {key} from matrix room {self.room_id}: {data.message}({data.status_code})"
+            )
+            return
+
+        if data.transport_response.status == 404:
+            _LOGGER.error(f"State event with state key '{state_key}' doesn't exist")
+            return
+        data = data.content
```
Cadair, line 211:
```suggestion

        data = data.content
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+            room_id=self.room_id, event_type=self._event_type, state_key=state_key,
+        )
+        if isinstance(data, RoomGetStateEventError):
+            _LOGGER.error(
+                f"Error deleting {key} from matrix room {self.room_id}: {data.message}({data.status_code})"
+            )
+            return
+
+        if data.transport_response.status == 404:
+            _LOGGER.error(f"State event with state key '{state_key}' doesn't exist")
```
Cadair, line 209:
```suggestion
            _LOGGER.error(f"State event {self._event_type} with state key '{state_key}' doesn't exist.")
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+            return
+        data = data.content
+
+        _LOGGER.debug(f"Got {data} from state request.")
```
Cadair, line 213:
```suggestion
        _LOGGER.debug(f"Got {data} from state event in room {self.room_id}.")
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+
+        if not isinstance(key, list):
+            key = [key]
+
+        return_value = []
+        for k in key:  # key can be a list of keys to delete
+            try:
+                return_value.append(data[k])
+                _LOGGER.debug(f"Deleting {k} from {self.room_id}")
```
Cadair, line 222:
```suggestion
                _LOGGER.debug(f"Deleting key '{k}' from database in {self.room_id}.")
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+                del data[k]
+            except KeyError:
+                _LOGGER.debug(f"Not deleting {k}, as it doesn't exist")
```
Cadair, line 225: I think this should be logged at a higher level than debug?
```suggestion
                _LOGGER.warning(f"Unable to delete '{k}' from database in room {self.room_id} as it doesn't exist.")
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+        if not return_value:
+            return None
+        elif len(return_value) == 1:
+            return return_value[0]
+        else:
+            return return_value
+
+    @contextmanager
+    def memory_in_room(self, room):
+        """Switch to a different room for some operations."""
```
Cadair, line 246:
```suggestion
        """Use room state in the given room rather than the default."""
```

tests/test_database_matrix.py
```diff
@@ -0,0 +1,1677 @@
...
+            matrix_call(
+                "PUT",
+                "/_matrix/client/r0/rooms/%21notaroomid/state/dev.opsdroid.database/",
+                {"twim": value},
+            ),
+        ],
+    )
+
+
+# @pytest.mark.asyncio
```
Cadair, line 380: Please remove all these commented out tests.

tests/test_database_matrix.py
```diff
@@ -0,0 +1,1677 @@
...
+                "PUT",
+                "/_matrix/client/r0/rooms/%21notaroomid/state/dev.opsdroid.database/hello",
+                {"hello": "world"},
+            ),
+        ],
+    )
+
+
+@pytest.mark.asyncio
+@pytest.mark.xfail(
```
Cadair, line 242: I think skip might be better than `xfail` if the deps aren't installed?

Cadair (changes requested):
I have tested this out a bunch today, I think it's great! There are a few nits left, mainly the wording of the logger calls.

@SolarDrew You going to have time to get a good review of this in before we merge it?

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+                if event["type"] == "opsdroid.database" and event["content"]:
+                    if not self._single_state_key:
+                        event["content"] = {event["state_key"]: event["content"]}
+                    await self.connector.connection.room_put_state(
+                        room_id=self.room_id,
+                        event_type=self._event_type,
+                        content=event["content"],
+                        state_key=event["state_key"],
+                    )
+                    await self.connector.connection.room_redact(
```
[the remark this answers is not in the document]
Tyagdit: yep even if it fails to redact we won't try to migrate again. The older state event will stay but be ignored

tests/test_database_matrix.py
Tyagdit: We're mocking the methods that encryption would affect anyway so without deps they would still fail correctly. Plus in that case we won't know if they xpass after future changes.

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+_LOGGER = logging.getLogger(__name__)
+CONFIG_SCHEMA = {
+    "default_room": str,
+    "single_state_key": Any(bool, str),
+    "should_encrypt": bool,
+}
+
+
+def memory_in_event_room(func):
+    """Use room state from the received message rather than the default."""
```
Cadair, line 22:
```suggestion
    """Use room state from the room the message was received in rather than the default."""
```

opsdroid/database/matrix/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+    async def connect(self):
+        """Connect to the database."""
+
+        _LOGGER.info("Matrix Database connector initialised.")
+
+    async def put(self, key, value):
+        """Insert or replace a value into the database for a given key."""
+
+        if self.should_migrate:
+            await self.migrate_database()
```
SolarDrew, line 91: Is it worth just doing this on startup rather than checking if it's been done every time `get()` or `put()` is called?

SolarDrew (approved):
I've not had a chance to look properly through the tests but since they're passing and the coverage is fine I'm happy to accept that they're sensible.

Thanks for all your work getting this into core, @Tyagdit , I think it's going to prove to be a really useful feature for opsdroid.

opsdroid/database/matrix/__init__.py
Cadair: you could be in a new room everytime with the decorator.

Cadair (approved):

Issue #1541, opened 2020-06-17
Drop Python 3.6


jacobtomlinson:
We should now drop Python 3.6 as 0.19 is out.

_Originally posted by @Cadair in https://github.com/opsdroid/opsdroid/issues/1534#issuecomment-644825884_

PR #1586, opened 2020-07-21
Removing Python 3.6 from setup.py and also the support table as per #1541

suvhotta:
# Description

Fixes #1541

## Status
**READY** | **UNDER DEVELOPMENT** | **ON HOLD**

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update
- Documentation (fix or adds documentation)

# How Has This Been Tested?

- Test A
- Test B

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

There is a modification in the existing documentation hence no new test cases have been added.

PR #1588, opened 2020-07-22
dropping python 3.6

suvhotta:
# Description

Removed the python 3.6 dependencies from the following locations : 
    setup.py, ci.yml, tox.ini, .readthedocs..yml, opsdroid/cli/utils.py, supported-python-versions.md

Fixes #<issue>
#1541 

## Status
**READY** 

## Type of change

- Updating the existing dependencies and documentations.

# Checklist:
- [x]  I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.
- [x] I have modified the tests so that python3.6 is accordingly verified.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -3,7 +3,7 @@
 version: 2
 
 python:
-  version: 3.6
+  version: 3.7
   install:
     - requirements: requirements.txt
     - requirements: requirements_test.txt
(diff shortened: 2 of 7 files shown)


suvhotta:
@FabioRosado I've modified all the files and dropped python3.6 however still test(ubuntu-latest, 3.6) and test(windows-latest 3.6) are running when I created this PR, can u help me with that?

Cadair:
Something I noticed looking over this is we currently don't specify the supported python version in the setup.py file. i.e. [this](https://setuptools.readthedocs.io/en/latest/setuptools.html?highlight=python_requires#options) which tells PyPI which versions of Python we support. @suvhotta could you add that to the `setup()` function in setup.py?

suvhotta:
@Cadair I guess they're already specified in the setup.py file that we're currently supporting 3.7 and 3.8. Could you please recheck and lemme know if I've to change something else.

Cadair:
The classifiers are not the same as the `python_requires` key. `python_requires` is a formal version spec equivalent to `install_requires` for the version of Python you support, so in this case it would be `python_requires='>=3.7'`.

suvhotta:
@FabioRosado  Can this be merged or need I do any more changes to it?

jacobtomlinson:
This all looks great. I think we just need to do as @Cadair says and add `python_requires='>=3.7'` to te setup.

jacobtomlinson:
Looks like this work has been done in #1586 (and merged in #1595).

PR #1599, opened 2020-07-29
Add xfail to watchdog test, fix deprecation warning for logging

FabioRosado:
# Description

I've tried to implement your suggestion Jacob but on my mac this test keeps failing even after I bumped the timeout to 30seconds. So I'm not too sure about how to fix this problem. Since you mentioned that we could add xfail to this particular test this is what I have done as well as a temporary fix until we can figure out how to fix it properly.

## Status
**READY** 

## Type of change

- Tests

# How Has This Been Tested?

- Tox - Green
- test_watchdog passes due to xfail.

# Checklist:


jacobtomlinson:
Thanks so much for looking at this!

PR #1604, opened 2020-07-30
Experiment with Events as dataclasses

Cadair:
Seems to mostly work. Had to do some hacks to maintain API.

The largest advantage of this that I can see is `__repr__` and `__eq__` now "do the right thing"


Cadair:
I would like to try and use this to improve the documentation generation, i.e. pulling the types (and maybe the descriptions) from the annotations.

jacobtomlinson:
Sure! I generally want to move documentation from the markdown files into docstrings and other automatic generation tools.

If we can get the markdown bare bones enough (mostly directives within `eval_rst` blocks) then I may support switching the whole lot to rst 😉.

Cadair:
I tired to use the typehints autodoc plugin and it all fell apart so might be swapping `events.md` to rst out of necessity :p

Cadair:
bad stale bot

jacobtomlinson:
Actually I think stale bot is right. What is happening here? Is this ready? We still have linting failures.

Cadair:
I need to do some documentation updates to make the type hints nice, and fix the linting.

Issue #1502, opened 2020-05-26
Migrate tests from unittest to pytest


jacobtomlinson:
Our existing test suite has been written with the Python [`unittest`](https://docs.python.org/3/library/unittest.html) framework. However, as the test suite has grown and opsdroid has become more complex we are running into issues with the tests. Mainly around setting up and tearing down tests.

The @opsdroid/maintainers team have decided that we want to migrate all tests to be written with the [`pytest`](https://docs.pytest.org/en/latest/contents.html) framework instead so that we can make better use of fixtures. Fixtures are more reusable and portable and should help reduce complexity all over.

There's a lot to be done but it can be done piece by piece as `pytest` can run tests in either format. So if you wish to help in the effort you can start by searching the codebase for unittest suites. These are classes which are subclassed from `unittest.TestCase` or `asynctest.TestCase`, so searching all files for `unittest.TestCase` and `asynctest.TestCase` should be a good place to start.

For detailed information on running the test suite and contributing to opsdroid [see the docs](https://docs.opsdroid.dev/en/latest/contributing/index.html). But the quickest way to get started us with [`tox`](https://tox.readthedocs.io/en/latest/).

```bash
pip install -U tox  # You only need to install tox once

tox -e py36,lint  # Run the Python 3.6 tests (the lowest version we support) and the linter
```

Once you have found a test suite you wish to convert there are a few steps you need to follow to convert from unittest to pytest:
- Move tests from top level `tests` directory to a nested `tests` directory in opsdroid. Create one in an appropriate place if there isn't already one.
- Remove test from class, pytest tests are just regular functions.
- Change assertions to use regular `assert` or [pytest assertions](https://docs.pytest.org/en/latest/assert.html).
- Mark async tests. In unittest we write async tests by using the `asynctest.TestCase`, but in pytest we decorate our tests with `@pytest.mark.asyncio` instead.
- Move setup operations to fixtures. If a test class contains a `setUp` method anything created here should become a fixture. Check the existing fixtures in `conftest.py` before creating new ones.
- Add docstrings to tests and fixtures to explain what they do. We have been pretty rubbish with this up until now and there are many tests which are not obvious in what they are testing.

Here's an example:

```python
# Before (unittest)
import asynctest
import asynctest.mock as mock

from opsdroid.cli.start import configure_lang
from opsdroid.core import OpsDroid
from opsdroid.events import Message
from opsdroid.matchers import match_regex
from opsdroid import constraints


[...]

PR #1608, opened 2020-08-03
mongo test from unittest to pytest

Waterbyte:
# Description

Fixes< #1502 >

## Status
**READY**

## Type of change

- Test

# How Has This Been Tested?

- Local Pytest check

# Checklist:


opsdroid/database/mongo/tests/test_mongo.py
```diff
@@ -0,0 +1,77 @@
...
+from opsdroid.database.mockmodules.mongo.mongo_database import DatabaseMongoTest
+from opsdroid.database.mongo import DatabaseMongo
+
+
+@pytest.fixture()
+def database(config):
+    return DatabaseMongo(config)
+
+
+@pytest.mark.parametrize("config", [{"example_item": "test"}])
```
FabioRosado, line 12: I am not particular familar with `pytest.mark.parametrize` since I've just recently started using pytest instead of unittest. But I'm assuming this is loading a fake config, correct? 

Perhaps we could change the "example_item": "test" for "database": "mongo"?

opsdroid/database/mongo/tests/test_mongo.py
```diff
@@ -0,0 +1,77 @@
...
+        raise Exception
+    else:
+        pass
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("config", [{}])
+async def test_get(database):
+    database.database = {}
+    database.database["test"] = DatabaseMongoTest({})
```
FabioRosado, line 35:
```suggestion
    database.database["mongo"] = DatabaseMongoTest({})
```

I just think it would make more sense to name it mongo than "test", what do you think?

opsdroid/database/mongo/tests/test_mongo.py
```diff
@@ -0,0 +1,77 @@
...
+    with pytest.raises(TypeError):
+        await database.get("test")
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("config", [{}])
+async def test_put(database):
+    database.database = {}
+    database.database["test"] = DatabaseMongoTest({})
```
FabioRosado, line 44: Same as above

opsdroid/database/mongo/tests/test_mongo.py
Waterbyte: Hi, @FabioRosado , thanks for the reply. The name "test" was in the unittest counterpart, so I kept it same. Happy to change it since it makes more sense.

opsdroid/database/mongo/tests/test_mongo.py
Waterbyte: Hi, yes `pytest.mark.parametrize` loads fake config. Since "mongo" is DBMS name, wouldn't it be better to have something like "database":"test_db"?

opsdroid/database/mongo/tests/test_mongo.py
FabioRosado: Oh I understand, I haven't look at the old tests haha would be good to change it to mongo then 😁

opsdroid/database/mongo/tests/test_mongo.py
FabioRosado: Thank you for the explanation that makes sense 😁  yeah that would make more sense then, we are on the mongo tests anyway and you can call your collection whatever you need 👍

PR #1618, opened 2020-08-11
Add handle for message_replied subtype

### LINKED ISSUE #1617: ERROR opsdroid.connector.slack.events: Username lookup failed for 'user'.
# Description
Every time a message is sent inside a thread in Slack, the RTM API sends a message event with subtype "message_replied" that at the moment is not properly handled.

This is causing Opsdroid to call unnecessarily the [`_get_user_name` ](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/slack/events.py#L195) method in the slack connector, which at the same time ends in an unnecessary call to the Slack API.

Everytime a response is originated inside a thread the logs show the message:
```
ERROR opsdroid.connector.slack.events: Username lookup failed for 'user'.
ERROR opsdroid.connector.slack.events: Username lookup failed for 'user'.
```

## Steps to Reproduce
Use the hello skill to reply inside a thread in Slack

## Expected Functionality
The subtype "message_replied"  has to be properly handled

## Experienced Functionality

Everytime a response is originated inside a thread the logs show the message:
```
ERROR opsdroid.connector.slack.events: Username lookup failed for 'user'.
ERROR opsdroid.connector.slack.events: Username lookup failed for 'user'.
```

## Versions
- **Opsdroid version: v0.19.0+17.g78046f2.dirty
- **Python version: 3.8
- **OS/Docker version:Ubuntu 20.04

## Configuration File
```
logging:
  level: info

welcome-message: true

connectors:
  - name: websocket
  - name: slack
    token: $SLACK_TOKEN
    start_thread: true

skills:
  hello: {}
```

chillipeper:
# Description

This is an attempt to fix #1617 

Tests are not added, as the method being added is not implemented.

Would like to know if this is the correct approach, and any suggestion on tests also welcome!

## Status
 **UNDER DEVELOPMENT**

## Type of change

# How Has This Been Tested?
Tested Manually 

# Checklist:

- [X ] I have performed a self-review of my own code


FabioRosado:
Hello @chillipeper  thank you for both raising the issue and submitting this PR. This seems like something we need to address and support in our codebase. I'm thinking that perhaps a user could talk to opsdroid and then reply to an opsdroid message to do something.

For example, if you say to opsdroid "When was the last time I watered my plants" and opsdroid replies: "Two weeks ago" you might want to reply back with "Okay watered the plants" and opsdroid will update the timer for watering the plants (this is a hypothetical skill)

I'd suggest expanding this PR if you can to implement a solution and you can even add that other message as a [linked event](https://github.com/opsdroid/opsdroid/blob/1f0bdfe00fd79108f3a25fc08f06a9097d1e7352/opsdroid/events.py#L87)

Also you would have to add tests for this so the coverage doesn't drop 😄 👍

PR #1619, opened 2020-08-15
migrated redis unittest to pytest

### LINKED ISSUE #1502: Migrate tests from unittest to pytest

Waterbyte:
# Description

Fixes #1502 

## Status
**READY**

## Type of change

- Test

# How Has This Been Tested?

- pytest in local machine

# Checklist:


opsdroid/database/redis/tests/test_redis.py
```diff
@@ -0,0 +1,127 @@
...
+    database = RedisDatabase({})
+    assert None is database.client
+    assert 0 == database.database
+    assert "localhost" == database.host
+    assert 6379 == database.port
+    assert None is database.password
+
+    log_iterator = iter(caplog.records)
+    first_log = next(log_iterator)
+    assert first_log.levelname == "DEBUG"
```
jacobtomlinson, line 29: Asserting that the first log line is a debug level message isn't really testing much of value here, and it going to be brittle. (I appreciate this is a direct port, but the tests were poor to begin with and it would be good to improve them while we convert).

It would be better to check what the line says and asset that in the whole text.

```python
assert "Loaded Redis database connector." in caplog.text()
```

Tests should check that things functioned correctly (picked up the correct default config) and that the user experience is as expected (logs a debug message to say it was loaded).

opsdroid/database/redis/tests/test_redis.py
```diff
@@ -0,0 +1,127 @@
...
+
+    await database.connect()
+
+    assert mocked_connection.called
+
+    log_iterator = iter(caplog.records)
+    first_log = next(log_iterator)
+    assert first_log.levelname == "DEBUG"
+    second_log = next(log_iterator)
+    assert second_log.levelname == "INFO"
```
jacobtomlinson, line 46: Same as above

opsdroid/database/redis/tests/test_redis.py
```diff
@@ -0,0 +1,127 @@
...
+    with suppress(OSError):
+        await database.connect()
+
+    assert mocked_connection.called
+
+    log_iterator = iter(caplog.records)
+    first_log = next(log_iterator)
+    assert first_log.levelname == "DEBUG"
+    second_log = next(log_iterator)
+    assert second_log.levelname == "WARNING"
```
jacobtomlinson, line 64: Same as above

opsdroid/database/redis/tests/test_redis.py
Waterbyte: @jacobtomlinson Thanks for the review. Although I would like to ask if I implemented your suggestion, wouldn't test fail if someone modifies those log messages? Is that situation ok?

jacobtomlinson: I'm happy with that situation. You could make the assertion simpler and just look for `Loaded Redis` or something to make it more robust.

Waterbyte:
@jacobtomlinson resolved assertion of logs issue

jacobtomlinson:
This is great many thanks!

Have you filled in the [claim form](https://forms.gle/iQZV9DjkaUiSaDXK6) for [opsdroid stickers](https://medium.com/opsdroid/contributor-sticker-packs-738058ceda59) btw? You've contributed a few times and I don't think I see you on the list. We would love to send you some to say thanks for your efforts here.

PR #1620, opened 2020-08-21
Converted the test_database_sqlite.py file to a pytest

rr3khan:
# Description

Converted test_database_sqlite.py  to pytest

Fixes #<#1502>

## Status
 **Done** 

## Type of change

Test

# How Has This Been Tested?

Tests pass locally using Tox

# Checklist:


FabioRosado:
Hello, I did a quick gaze at it but will give you a proper review later.

Have you installed the dev and test dependencies? You can do it with the command `pip install -r requirements_dev.txt` and again the same but for the tests one.

If that doesn't work, try to run tox from within the opsdroid folder and use the `tox` command again.

If you are still having issues let me know and I'll try to give you a hand

tests/test_database_sqlite.py
```diff
@@ -1,115 +1,116 @@
...
 
+    """
+    database = DatabaseSqlite({"path": "sqlite.db"})
+    assert database.client == None
+    assert database.db_file == "sqlite.db"
+    assert database.table == "opsdroid"
+    assert database.conn_args == {"isolation_level": None}
 
-class TestDatabaseSqliteAsync(asynctest.TestCase):
     """A async database test class.
```
FabioRosado, line 32: Since you removed the class from here, I think it's better to just remove this.

tests/test_database_sqlite.py
```diff
@@ -1,115 +1,116 @@
...
 
+import asyncio
+
 from opsdroid.database.sqlite import DatabaseSqlite
 from opsdroid.cli.start import configure_lang
 
-
-class TestDatabaseSqlite(unittest.TestCase):
-    """A database test class.
+"""A database test class.
```
FabioRosado, line 11: Let's remove these comments. My reasoning behind that is that we already have the comment on line 1 saying what the test is about 😄

tests/test_database_sqlite.py
```diff
@@ -1,115 +1,116 @@
...
+@pytest.mark.asyncio
+async def test_connect():
+    """Test database connection.
+
+    This method will test the database connection of sqlite database.
+    As the database is created `opsdroid` table is created first.
+
+    """
+    database = DatabaseSqlite({"path": "sqlite.db"})
+    opsdroid = amock.CoroutineMock()
```
FabioRosado, line 48: We have an `opsdroid` fixture, I believe if you add opsdroid as a parameter of the function you could then remove this line.

So this function will look like: `async def test_connect(opsdroid):`
I'm happy to keep it as it is if you think is best (or if the fixture won't work haha)

If it works and you are happy with it, then we could refactor the tests a bit more and add opsdroid as a parameter to every function on this file.

tests/test_database_sqlite.py
```diff
@@ -1,115 +1,116 @@
...
+    else:
+        assert table == "opsdroid"
+        assert data == {}
+        assert client == "Connection"
+
+
+@pytest.mark.asyncio
+async def test_deprecated_path():
+    database = DatabaseSqlite({"file": "sqlite.db"})
+    assert database.db_file == "sqlite.db"
```
FabioRosado, line 116: I know that this wasn't added to the previous tests, but I think we could add another assertion here to check if the logging was shown or not - this is mostly so if we move things around then this test will fail and we know that we need to fix something.

This test would have to look like this:

```python
@pytest.mark.asyncio
async def test_deprecated_path(caplog):
    database = DatabaseSqlite({"file": "sqlite.db"})
    assert database.db_file == "sqlite.db"
    assert "The option 'file' is deprecated, please use 'path' instead." in caplog.text
```

FabioRosado (changes requested):
Hello @rr3khan  thank you so much for raising this PR. I said that I was going to check it last week, but life got in the way so I apologise for the long wait.

I've added some comments to the PR, the only thing I really think we should remove is the comments, the rest are just comments to know your opinion about the changes proposed.

I also got to say that this is some great work and the docstrings on the test functions are awesome as it will help folks when they come and have to do some changes in the future 😄 👍

tests/test_database_sqlite.py
rr3khan: I tried this but unfortunately could not get it to work, I kept getting an error about the 'opsdroid' fixture not being found.

FabioRosado (approved):
Awesome thank you so much @rr3khan  for the update of the PR! I am happy to merge it now 😄 👍

PR #1621, opened 2020-08-21
Added document options for events with multiple possible users (#1448)

### LINKED ISSUE #1448: Document options for events with multiple possible users
Some chat platforms supported by opsdroid have the possibility of a single event corresponding to multiple users. For example, slack's api call to invite users to a channel takes a list of users as input. Opsdroid, on the other hand, only supports a single user on any given event, and for the sake of flexibility I think this is as it should be.

That being the case, the documentation needs to reflect this disconnect - skill developers should know that they need to create multiple single-user events rather than trying to create one multi-user one, or that their skill can call the api directly rather than emitting many events, if they're concerned about performance. All of this should be described in the documentation for the [base `Event` class](https://docs.opsdroid.dev/en/stable/skills/events.html).

haardikdharma10:
# Description

Added the documentation for issue #1448. As mentioned, included a small note at the end of Events section so that developers know about events with multiple possible users. 

# Status

Ready for review.

# Fixes Issue

Fixes #1448 

## Type of change

# Checklist:

- [x] I have made corresponding changes to the documentation.

### DIFF
--- a/docs/skills/events.md
+++ b/docs/skills/events.md
@@ -14,6 +14,7 @@ The base `Event` class has the following attributes and methods.
 .. autoclass:: opsdroid.events.Event
    :members:
 ```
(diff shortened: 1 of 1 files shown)


FabioRosado (commented):
Hello @haardikdharma10 thank you for working on this, the PR seems good to me. @SolarDrew  since you created the issue, can you give your input on this PR? If you think it does what you mentioned we can merge it then 😁

SolarDrew:
Yep, looks good to me. Thanks again for fixing this issue @haardikdharma10 :D

haardikdharma10:
@FabioRosado, @SolarDrew It's my pleasure. Looking forward to fix more issues :D

PR #1622, opened 2020-08-21
Migrate test_logging.py to pytest

### LINKED ISSUE #1502: Migrate tests from unittest to pytest

tardigrde:
# Description
Migrating test_logging.py to pytest.

Fixes #1502 

## Status
READY

## Type of change
Test

# How Has This Been Tested?
With pytest on local machine.

# Checklist:


tests/test_logging.py
```diff
@@ -134,38 +129,42 @@ def test_configure_extended_logging(self):
...
-        self.config = {}
-        opsdroid.configure_logging(self.config)
 
-    def test_configure_whitelist_and_blacklist(self):
-        opsdroid.configure_logging(self.config)
-        self.assertLogs("_LOGGER", "warning")
+class TestWhiteAndBlackFilter:
+    def test_configure_whitelist_and_blacklist(self, white_and_black_config):
+        opsdroid.configure_logging(white_and_black_config)
+        # self.assertLogs("_LOGGER", "warning")
```
FabioRosado, line 170: In here I think we could replace this assert with a caplog and assert if the warning was logged since whitelist and blacklist can't be used together.

tests/test_logging.py
```diff
@@ -134,38 +129,42 @@ def test_configure_extended_logging(self):
...
-                "filter": {"whitelist": ["opsdroid"], "blacklist": ["opsdroid.core"]},
-            }
+        assert len(rootlogger.handlers), 2
+        # TODO use isinstance
+        assert isinstance(rootlogger.handlers[0], logging.StreamHandler)
+        assert rootlogger.handlers[0].level == logging.INFO
+        assert isinstance(rootlogger.handlers[1], logging.handlers.RotatingFileHandler)
+
+        assert rootlogger.handlers[1].level == logging.INFO
+        # self.assertLogs("_LOGGER", "info")
```
FabioRosado, line 146: This one was checking if `_LOGGER` is using the info level, but now that I look at it, the previous assertion is doing exactly that, so we could just remove it

FabioRosado (commented):
Hello @tardigrde  apologies for the delay in replying. You can use `self.assertLogs` to assert if the log was called. The `self.assertLogs(_LOGGER, None)` bit that you are talking about, is to check that if we have the whole logging method blacklisted, then opsdroid shouldn't log anything so `_LOGGER` should be none.

That test might be a bit too broad but let us test that the blacklist filter is working 😄 

I'd be happy to replace that with `capsys` if it checks that nothing was logged 👍 

I also noticed that we have a lot of:

```python
        # self.assertLogs("_LOGGER", "info")
```

But  I see that we are already checking if the level is the right one with this assertion: `assert rootlogger.handlers[0].level == logging.CRITICAL` so I guess we can just remove most of the `self.assertLogs`

jacobtomlinson:
With pytest we can use `caplog` which is a much more pleasant way to test the logs.

https://docs.pytest.org/en/stable/logging.html#caplog-fixture

Check out some of @Waterbyte's recent PRs for example of using this.

Also GitHub has a Draft PR featurw if you're not ready for things to be reviewed or merged. I'm going to put this PR into draft, please mark it as ready for review when you are done with it.

tests/test_logging.py
tardigrde: Hmm. 
I have tried using `caplog ` as described [here](https://docs.pytest.org/en/stable/logging.html#caplog-fixture). Or as @Waterbyte  used it [here](https://github.com/opsdroid/opsdroid/pull/1619/files). 
To my great surprise `caplog.text` is an empty string, even though this function call indeed creates logs at `Warning` level. All of the other properties of caplog (`messages`, `record_tuples` and `records`)  are empty lists, except `handler`.

See my code: 
```
def test_configure_whitelist_and_blacklist(self, caplog, white_and_black_config):
        caplog.set_level(logging.INFO)

        opsdroid.configure_logging(white_and_black_config)

        expected_log = '''Both whitelist and blacklist filters \
        found in configuration. Only one can be used at a time \
        - only the whitelist filter will be used.'''

        assert expected_log in caplog.text
```

I have tried setting the logger name to something else explicitly, like this:
> caplog.set_level(logging.INFO, logger='opsdroid.logging'),
None of them worked. 

Do you have a tip of guess why is this not working as it should?

tardigrde: Ok, I resolved the problem with [capsys](https://docs.pytest.org/en/stable/capture.html#accessing-captured-output-from-a-test-function). I used `capsys` in the other two methods as well, where the test checks if there is no output at all as there is no logger configured.

tests/test_logging.py
tardigrde: I removed them

tardigrde:
I marked this as ready. Can you please review it?

FabioRosado (changes requested):
Hello @tardigrde  Thank you for working on this again and for updating the PR.
I added a few comments on the code, everything looks good.

I've also tested the white&black filter test on my end with caplog and it seems that when we run the filters, the logging might not have been set up properly yet, that's why we are unable to capture it from caplog.

I've also tried to change the level of that log to see if it would change anything but didn't. I'm assuming caplog can't see this log for some reason or perhaps the `self.assertLogs` is checking for std.err as well 🤷‍♂️

tests/test_logging.py
```diff
@@ -134,38 +134,50 @@ def test_configure_extended_logging(self):
...
-        self.assertEqual(rootlogger.handlers[1].level, logging.INFO)
-        self.assertLogs("_LOGGER", "info")
+        assert len(rootlogger.handlers), 2
+        assert isinstance(rootlogger.handlers[0], logging.StreamHandler)
+        assert rootlogger.handlers[0].level == logging.INFO
+        assert isinstance(rootlogger.handlers[1], logging.handlers.RotatingFileHandler)
+        assert rootlogger.handlers[1].level == logging.INFO
+
+        captured = capsys.readouterr()
+        assert "INFO opsdroid.logging: Started opsdroid" in captured.err
```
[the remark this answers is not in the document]
tardigrde: No, it was not working for me :(

tests/test_logging.py
```diff
@@ -8,91 +7,89 @@
...
 
-class TestLogging(unittest.TestCase):
-    """Test the logging module."""
 
-    def setUp(self):
-        self._tmp_dir = os.path.join(tempfile.gettempdir(), "opsdroid_tests")
-        with contextlib.suppress(FileExistsError):
-            os.makedirs(self._tmp_dir, mode=0o777)
-        configure_lang({})
+@pytest.fixture(scope="module")
```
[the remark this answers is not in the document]
tardigrde: Sure. You don't have this file yet, am I right? I can't seem to find it. 

btw pytest's way is using a file called conftest.py, see https://docs.pytest.org/en/stable/fixture.html#conftest-py-sharing-fixture-functions, so maybe that's what I would do, if it is ok for you.

tests/test_logging.py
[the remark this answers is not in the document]
FabioRosado: That's alright I think that's the issue yeah, since we are setting up logging then caplog doesn't seem to be able to get that. 

I might check the lib to see if there is any issue about that, but yeah I'm happy with this work 😄👍

jacobtomlinson (changes requested):
Can we move the test file to the new location. See #1502 for info.

tests/test_logging.py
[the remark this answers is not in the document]
FabioRosado: We have a fixtures file already, you can find it here: https://github.com/opsdroid/opsdroid/blob/master/opsdroid/testing/fixtures.py 

If you can't see it you might need to update your fork, if you need help doing that let me know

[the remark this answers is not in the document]
tardigrde: Got it! I'll push in a minute.

tardigrde:
> Can we move the test file to the new location. See #1502 for info.

Of course!

tests/test_logging.py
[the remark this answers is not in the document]
tardigrde: So moving the function to fixtures made me import it, but then it does not behave like a pytest fixture, so I can't include it in the function parameter list, but use it in the function from the outer scope. With conftest.py it might have been better.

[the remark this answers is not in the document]
tardigrde: Ahh and it fails because of this.
But if I leave the _tmp_dir fixture in the parameter list of the function, then black won't let me commit, with:
```
opsdroid/tests/test_logging.py:8:1: F401 'opsdroid.testing.fixtures._tmp_dir' imported but unused
```
And several others, like:
```
opsdroid/tests/test_logging.py:33:33: F811 redefinition of unused '_tmp_dir' from line 8
```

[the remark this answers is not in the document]
tardigrde: How can I make black silent? :shushing_face: 
Or do you have a better idea?

opsdroid/testing/fixtures.py
```diff
@@ -15,3 +18,11 @@ def opsdroid():
...
     with OpsDroid(config={}) as opsdroid:
         yield opsdroid
+
+
+@pytest.fixture
+def _tmp_dir():
+    _tmp_dir = os.path.join(tempfile.gettempdir(), "opsdroid_tests")
+    with contextlib.suppress(FileExistsError):
+        os.makedirs(_tmp_dir, mode=0o777)
+    return _tmp_dir
```
jacobtomlinson, line 28: This isn't quite how fixtures work and there are also cleaner ways to do this.

I don't think we should be using a fixture for this. See my other comment.

opsdroid/tests/test_logging.py
```diff
@@ -0,0 +1,175 @@
...
+    config = {
+        "logging": {"path": os.path.join(_tmp_dir, "output.log"), "console": False}
+    }
+    opsdroid.configure_logging(config)
+    rootlogger = logging.getLogger()
+    assert len(rootlogger.handlers), 2
+    assert isinstance(rootlogger.handlers[0], logging.StreamHandler)
+    assert rootlogger.handlers[0].level, logging.CRITICAL
+    assert isinstance(rootlogger.handlers[1], logging.handlers.RotatingFileHandler)
+    assert rootlogger.handlers[1].level == logging.INFO
```
jacobtomlinson, line 43: Instead of using your `_tmp_dir` fixture it would be better to use `tempfile.TemporaryDirectory` in a context manager.

```suggestion
def test_configure_file_logging():
    with tempfile.TemporaryDirectory() as tmpdir:
        config = {
            "logging": {"path": os.path.join(tmpdir, "output.log"), "console": False}
        }
        opsdroid.configure_logging(config)
        rootlogger = logging.getLogger()
        assert len(rootlogger.handlers), 2
        assert isinstance(rootlogger.handlers[0], logging.StreamHandler)
        assert rootlogger.handlers[0].level, logging.CRITICAL
        assert isinstance(rootlogger.handlers[1], logging.handlers.RotatingFileHandler)
        assert rootlogger.handlers[1].level == logging.INFO
```

opsdroid/tests/test_logging.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def test_configure_file_blacklist(capsys):
+    config = {
+        "logging": {
+            "path": os.path.join(_tmp_dir, "output.log"),
```
jacobtomlinson, line 49: Same here see other comment.

opsdroid/tests/test_logging.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def test_configure_file_logging_directory_not_exists(mocker):
+    logmock = mocker.patch("logging.getLogger")
+    mocklogger = mocker.MagicMock()
+    mocklogger.handlers = [True]
+    logmock.return_value = mocklogger
+    config = {
+        "logging": {
+            "path": os.path.join(_tmp_dir, "mynonexistingdirectory", "output.log"),
```
jacobtomlinson, line 73: Same here

jacobtomlinson (changes requested):
Thanks for this.

I don't think your temporary directory fixture is working correctly and also I think it would be better to use a context manager in the tests themselves. Something like this.

```
with tempfile.TemporaryDirectory() as tmpdir:
    # Do some things
```

opsdroid/testing/fixtures.py
tardigrde: true

opsdroid/tests/test_logging.py
tardigrde: thanks for the tip!
pushed my changes accordingly.

opsdroid/tests/test_logging.py
```diff
@@ -0,0 +1,177 @@
...
+import logging
+import os
+import tempfile
+
+import opsdroid.logging as opsdroid
+from opsdroid.cli.start import configure_lang
+
+configure_lang({})
+
+"""Test the logging module."""
```
FabioRosado, line 11: This comment should be on line 1 since this is the module docstring
```suggestion

```

opsdroid/tests/test_logging.py
```diff
@@ -0,0 +1,177 @@
+import pytest
```
FabioRosado, line 1:
```suggestion
"""Test the logging module."""
import pytest
```

FabioRosado (changes requested):
I've checked the PR following your talk with Jacob and everything seems good. The only thing I am suggesting here is to move the comment to the first line instead of line 11. Then I'm happy to merge it 😄 👍 

I've also added that as a suggestion so you can just click the button haha

tardigrde:
Thank you. That is very kind of you :D
I like your project, I'm planning to contribute some more.

FabioRosado (approved):
Awesome thank you so much for working on this. I am going to merge it now 😄 👍

PR #1626, opened 2020-08-25
Removed Pyup badge from README (fixes #1623)

### LINKED ISSUE #1623: Pyup badge in README links to a 404 page.
# Description
The Pyup badge in the README links to a 404 page. Can we consider removing it? It links to https://pyup.io/repos/github/opsdroid/opsdroid/ which is pretty straightforward and correct as other repositories that use this badge have the same format and links correctly for them.

haardikdharma10:
## Description
Removed the Pyup badge from README as it was not used anymore. Also reviewed all other badges, all of them work perfectly. Sorry about the messed up commits :/

Fixes #1623 

## Status
**READY** 

## Type of change


haardikdharma10:
cc: @jacobtomlinson

opsdroid/core.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, config=None, config_path=None):
         self.connector_tasks = []
         self.eventloop = asyncio.get_event_loop()
         if os.name != "nt":
-            for sig in (signal.SIGINT, signal.SIGTERM):
+            for sig in (signal.SIGINT, signal.SIGTERM, signal.SIGHUP):
```
Cadair, line 57: This isn't correct. The `SIGINT` and `SIGTERM` signals stop opsdroid running, where as we want `SIGHUP` to reload opsdroid.

This needs to be a separate block, which adds an event handler like `INT` and `TERM` but for a different method which reloads opsdroid. (I assume that method already exists).

FabioRosado (changes requested):
I think Cadair already did a good job on how to fix the SIGHUP issue. 
Could I just point out that, since these two issues are very different from one another, it's best to raise two PR's - one for each issue.

You can do that by creating a new branch with the command:
```shell
git checkout -b "<branch name>"
```

Then you can work on this particular branch and push to your origin with `git push -u origin <branch name>`.

You can then raise a PR on opsdroid.

Hope this helps and yeah I'd rather have this issue to solves the `pyup` issue - since it's already done and easy to merge and you can have some time to work on the SIGHUP issue by following Cadair suggestions and also creating tests proving that your code is working and keep the coverage the same.

If you need any help with anything let us know. We will be happy to give you a hand.

haardikdharma10:
Hey @FabioRosado, thanks for the review. I understood Cadair's point and will start working on it. Should I just change the core.py file back to how it was before and add a commit. This will result in only one file changed (README.md) and hence the issue of pyup will be sorted and this PR can then be merged. I'll then work on creating a new PR as suggested by you. Can you give me a thumbs up on this so that I can go ahead.

FabioRosado (approved):
Alright, I've just updated your branch and I'm happy to merge this PR.

You might need, to update your version before you start working on the other issue 👍

haardikdharma10:
Thank you Fabio. Also, out of curiosity, why don't my github profile shows in the Contributors section in README? :( Just looks better that way :P Is the contributors graph not updated?

FabioRosado:
Not sure about that. Did the commit shown today? Not sure if there is some sort of delay on GitHub part or not but it should show - unless the GitHub profile shows only the most active contributors? 🤔

haardikdharma10:
It surely doesn't show the most active contributors because there are github profiles which have contributed lesser than me and still there in the graph :/ Also I guess there's some issue as it shows 2 commits and 1 line changed which definitely isn't true in my case. Don't know what's wrong.

FabioRosado:
Unfortunately, I can't really help you with that, since this seems something that is a github thing. We have no control on which users shown on there.

The reason why it shows that you have 2 commits and 1 line changed, is because you added one line on the markdownfile and removed one. I'm assuming here, but github probably counted the addition one only.

On opsdroid, we also squash merge each commit to keep the tree clean, so any commits that you have done on each PR will be squashed into a single one. Hope this helps a bit 👍

opsdroid/core.py
haardikdharma10: Hi @Cadair, I got your point and also understood that the SIGHUP handler should be added to the reload method (which already exists) but ain't quiet getting how to go about it. A little help would be appreciated:)

PR #1627, opened 2020-08-27
Temporarily pin black to version 19.10b0

Tyagdit:
# Description

Black's new release has a [bug](https://github.com/psf/black/issues/1629) that makes it pretty much unusable, so pinning the older one till that gets fixed

## Status
**READY** 

## Type of change

# Checklist:


jacobtomlinson:
Fun! Thanks for the pin.

PR #1632, opened 2020-09-02
Migrate test_matchers.py to pytest

### LINKED ISSUE #1502: Migrate tests from unittest to pytest

tardigrde:
# Description

Migrating test_matchers.py to pytest.

Fixes #1502 

## Status
**READY**

## Type of change

Test

# How Has This Been Tested?

With pytest on local machine.

# Checklist:


tardigrde:
A little bit of help needed.

The test has these mocked async functions, that return another function. One of them is `get_mock_skill` right in the beginning of the file. I have tried to move these under a pytest.fixture, but then it is required to put the 3 input parameters `opsdroid, config, message` in the function calls (e.g. at test_matchers.py:35 -> `opsdroid.skills.append(decorator(await get_mock_skill()))`, the function call would need 3 input parameters). Should they be even fixtures? I don't think so, because they are not results we are asserting to, but utility functions for the tests.

If these mocked async functions are okay the way they are, then this is ready for a code review.

FabioRosado (approved):
Hello @tardigrde  thank you for working on this issue.

I guess a way to do it would be passing the opsdroid fixture to `get_mock_skill` and a config dict for the `config` parameter and then a Message event for the message parameter.

I'm not sure if it will matter much on this case to be honest and seems that what you have done is still working and following the same pattern of the unittests so it's probably fine.

tardigrde:
Okay, then I would mark this as ready for review, if the above mentioned thing is not a problem.

FabioRosado:
Thanks again for working on this it's now merged 😁

PR #1634, opened 2020-09-04
Convert test parser format to pytest

rr3khan:
# Description

Converted test_parser_format.py to a pytest

Fixes #<#1502>

## Status
**READY** 

## Type of change

Test

# How Has This Been Tested?

Tests pass locally with Tox

# Checklist:


rr3khan:
I tried to use 'opsdroid' as a parameter instead of using 'with OpsDroid as opsdroid:' but it did not seem to work for this test, previously it worked for the test_web.py test, so I am not sure if I am just missing an import for it.

opsdroid/parsers/tests/test_parser_format.py
```diff
@@ -0,0 +1,81 @@
...
+
+async def getMockSkill():
+    async def mockedskill(opsdroid, config, message):
+        pass
+
+    mockedskill.config = {}
+    return mockedskill
+
+
+async def test_parse_format_match_condition():
```
FabioRosado, line 22: We have an opsdroid fixture, so we can refactor this text a bit - we can remove the `with` statement and add `opsdroid` as arg for the test.

```suggestion
async def test_parse_format_match_condition(opsdroid):
```

opsdroid/parsers/tests/test_parser_format.py
```diff
@@ -0,0 +1,81 @@
...
+
+        message = Message("Hello world", "user", "default", mock_connector)
+        skills = await parse_format(opsdroid, opsdroid.skills, message)
+        assert len(skills) == 0
+
+        message = Message("Hello", "user", "default", mock_connector)
+        skills = await parse_format(opsdroid, opsdroid.skills, message)
+        assert mock_skill == skills[0]["skill"]
+        assert skills[0]["message"] is message
+        """Test that the original object has had a new attribute added."""
```
FabioRosado, line 38: I think this should be changed to just a comment since we don't need multiple lines.
```suggestion
        # Test that the original object has had a new attribute added.
```

opsdroid/parsers/tests/test_parser_format.py
```diff
@@ -0,0 +1,81 @@
...
+        assert hasattr(message, "parse_result")
+
+
+async def test_parse_format_search_condition():
+    with OpsDroid() as opsdroid:
```
FabioRosado, line 43: Same thing as the previous comment, let's add opsdroid as an arg and remove this with 👍

opsdroid/parsers/tests/test_parser_format.py
```diff
@@ -0,0 +1,81 @@
...
+        skills = await parse_format(opsdroid, opsdroid.skills, message)
+        assert mock_skill == skills[0]["skill"]
+
+        message = Message("Hello world", "user", "default", mock_connector)
+        skills = await parse_format(opsdroid, opsdroid.skills, message)
+        assert mock_skill == skills[0]["skill"]
+
+
+async def test_parse_format_parameters():
+    with OpsDroid() as opsdroid:
```
FabioRosado, line 62: Same thing here

FabioRosado (changes requested):
Hello again @rr3khan it's great to see you tackling so many of these tests! We really appreciate your kind help and time 👍 

I've added a few comments. Mostly, is to replace all the with statements since we are can use the opsdroid fixture 👍

opsdroid/parsers/tests/test_parser_format.py
rr3khan: Hi @FabioRosado I already tried using opsdroid as an argument but I kept getting an error about it not being a fixture. This is strange since it worked as a fixture for the test_web.py. So I am not sure if I am missing an import or if I am just incorrectly using it.

FabioRosado (approved):
Awesome! Thank you so much for working on this and making the changes that I have mentioned.
I am assuming that you fixed the fixture not being used issue right? 

I got the notification yesterday, but was at work and couldn't reply to you. 

Will merge this PR now 😁 👍

rr3khan:
Yes I was able to fix the fixture issue. Thanks.

PR #1635, opened 2020-09-04
Convert rocketchat connector test to pytest

rr3khan:
# Description

Started conversion of test_connector_rocketchat.py to pytest.
Fixes #<#1502>

## Status
**UNDER DEVELOPMENT** 

## Type of change
Test

# How Has This Been Tested?

Locally test fail with tox.

# Checklist:


FabioRosado:
Hello @rr3khan  How are you doing?

Do you need any help with these tests?

rr3khan:
Hi @FabioRosado  Yes, please, I have been running into some issues with await.connector.connect() do you mind reviewing what I have done so far?

opsdroid/connector/tests/test_connector_rocketchat.py
```diff
@@ -0,0 +1,295 @@
...
+def test_init():
+    """Test that the connector is initialised properly."""
+    connector = RocketChat(
+        {
+            "name": "rocket.chat",
+            "token": "test",
+            "user-id": "userID",
+            "update-interval": 0.1,
+        },
+        opsdroid=OpsDroid(),
```
FabioRosado, line 24: We should use the `opsdroid` fixture as an argument and pass it here. The same with all the other cases

opsdroid/connector/tests/test_connector_rocketchat.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+def test_missing_token(caplog):
+    """Test that attempt to connect without info raises an error."""
+
+    RocketChat({})
+    assert "Unable to login: Access token is missing." in caplog.text
+
+
+def setUp():
```
FabioRosado, line 37: Do we need the setup? In unittest we have to use this to make sure that every test uses whatever we add inside setUp, but not sure about pyttest - that's why I'm asking 😄

opsdroid/connector/tests/test_connector_rocketchat.py
```diff
@@ -0,0 +1,295 @@
...
+
+    connector.latest_update = "2018-10-08T12:57:37.126Z"
+
+    with amock.patch("aiohttp.ClientSession") as mocked_session:
+        connector.session = mocked_session
+
+
+@pytest.mark.asyncio
+async def test_connect(capsys):
+    connector = RocketChat({})
```
FabioRosado, line 57: By passing an empty dict into `RocketChat` you are passing and empty config option to the connector. That's why it's complaining about not having a token.

We can pass the config from the `setUp` test

```suggestion
    connector = RocketChat(
    	{
    		"name": "rocket.chat",
    		"token": "test",
    		"user-id": "userID",
    		"default_target": "test
    	}
    )
```

_Note: I'm using the browser to review and used tabs on this suggestion - not sure if committing this suggestion won't send those tabs and break python._

opsdroid/connector/tests/test_connector_rocketchat.py
```diff
@@ -0,0 +1,295 @@
...
+
+        captured = capsys.readouterr()
+        assert "DEBUG" in captured.err
+        assert patched_request.status != 200
+        assert patched_request.called
+
+
+@pytest.mark.asyncio
+async def test_connect_failure(caplog):
+    connector = RocketChat({})
```
FabioRosado, line 91: This config needs to be changed as well

FabioRosado (changes requested):
I had a look at the code and the tests and initially I thought we need to add `self.headers` to the connector, but we have that set up.

```python
>       resp = await self.session.get(self.build_url("me"), headers=self.headers)
E       AttributeError: 'RocketChat' object has no attribute 'headers'
```

If we go up to the `__init__` we can see that we are using a try/excerpt

```python
        try:
            self.user_id = config["user-id"]
            self.token = config["token"]
            self.headers = {"X-User-Id": self.user_id, "X-Auth-Token": self.token}
        except (KeyError, AttributeError):
            _LOGGER.error(
                _(
                    "Unable to login: Access token is missing. Rocket.Chat connector will not be available."
                )
            )
```

So even though the error fails on the self.headers, the issue that is causing this error, is the missing token on the configuration passed to `RocketChat`.

I've added a few comments to nudge you in the right direction, I think we could create a `config` variable and use the test config inside the `setUp` code, then we just need to pass that config to each test and use it on the assignment `connector = RocketChat(config)`.

If pyttest complains about passing the config as the test parameter, we could either make the config a fixture or add the same config to each test.

Hope this helps. Let me know if you need any further help with these tests 😄 👍

rr3khan:
Hi @FabioRosado I have been running into some issues when trying to using connector.session, get. The error states that get is not an attribute. Do you have any advice on this?

FabioRosado:
Hello @rr3khan  Just now that you closed this PR I've realised that I forgot to help you with the issues on the test. Apologies for my lack of response. Are you still interested in getting this merged? I'll try to run this PR on my end and play around with a test to see if I can figure out how to make it pass and then give you a hand.

Again apologies for forgetting to reply to you 😬

rr3khan:
Hi @FabioRosado , sorry about that, I am just a bit busy with midterms, so I might not have time to come back to this for a while. So I closed it, just in case I take too long to come back to it and accidentally forget about.

PR #1636, opened 2020-09-16
Add Language parameter to configuration example

### LINKED ISSUE #1633: Add lang parameter to configuration example
This is an issue that I'm raising due to #1630, after looking at the configuration example, I've noticed that the language parameter isn't set in the example.

We should update the language configuration to include the `lang` parameter - read the [configuration reference documentation on the language parameter](https://docs.opsdroid.dev/en/stable/configuration.html#language)

This parameter should be included in the configuration example, since new installation of opsdroid will use the example as starting point.

anoopmsivadas:
# Description

Just added a lang parameter to configuration example [included](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/configuration/example_configuration.yaml)
Fixes #1633

## Status
**READY** 

## Type of change

- Enhancement

# How Has This Been Tested?

- Tested manually

# Checklist:

### DIFF
--- a/opsdroid/configuration/example_configuration.yaml
+++ b/opsdroid/configuration/example_configuration.yaml
@@ -28,6 +28,9 @@
 ## Set the location for opsdroid to install modules
 # module-path: "."
 
+## Configure the language to use
+#lang: 'en'
+
 ## Show welcome message
 welcome-message: true
 


FabioRosado (approved):
This is great, thank you for working on this @anoopmsivadas !

PR #1638, opened 2020-09-26
flip a coin skill added to docs\examples

### LINKED ISSUE #1198: Add examples to the documentation
We have a new examples section in our documentation. This is a place for small guides demonstrating examples of what you could do with opsdroid.

We need help adding as many examples as we can to the documentation. We also have a bunch of ["official" skills](https://github.com/opsdroid?utf8=%E2%9C%93&q=%22skill-%22&type=&language=) that you can use with opsdroid, they would be a good place to start for creating examples. 

### How to add an example

- Create a new markdown file in [docs/examples](https://github.com/opsdroid/opsdroid/tree/master/docs/examples)
- Write a guide on building a skill with opsdroid roughly following this layout
  - Introduce what the skill will do ([get the weather forecast](https://github.com/opsdroid/skill-weather), [flip a coin](https://github.com/opsdroid/skill-random), [notify on CI builds](https://github.com/opsdroid/skill-travis), etc)
  - Creating the skill directory (bonus point for [using the cookiecutter](https://github.com/opsdroid/cookiecutter-skill))
  - Configuring the skill in opsdroid
  - Writing the skill
  - Demonstrating the skill
- Once you've written you example add it to the [documentation index](https://github.com/opsdroid/opsdroid/blob/master/mkdocs.yml#L13).

cristinasofia:
# Description

Feedback is welcome! Thank you.

Fixes #1198

### DIFF
--- a/docs/examples/flip-a-coin.md
+++ b/docs/examples/flip-a-coin.md
@@ -1,0 +1,42 @@
+# Flip A Coin
+
+A skill for opsdroid to perform flipping a coin.
+
+## Setting Up
+
+Navigating to 'opsdroid/skills/' there is a file called '__init__.py' in which will contain the functions for opsdroid to flip a coin.
+
+## Configure
+
+First, in the `__init__.py` file include the following packages:
+
+```python
+from opsdroid.skill import Skill
+from opsdroid.matchers import match_regex
+
+import random
+```
+
+Second, create a class for the skill:
+
+```python
+class CoinSkill(Skill):
+    @match_regex('flip a coin')
+    async def flip_a_coin(self, message):
+        flip = random.randint(0, 1)
+        if flip:
+            res = "Heads"
+        else:
+            res = "Tails"
+        
+        await message.respond("{}".format(res))
+
+```
+
+Third, open the 'configuration.yaml' file and add the path:
+
+```yaml
+skills:
+    flip-a-coin:
+    path: ~/skill-flipacoin
+```


docs/examples/flip-a-coin.md
```diff
@@ -0,0 +1,42 @@
...
+        if flip:
+            res = "Heads"
+        else:
+            res = "Tails"
+        
+        await message.respond("{}".format(res))
+
+```
+
+Third, open the 'configuration.yaml' file and add the path:
```
FabioRosado, line 36: It would be good to add how to open the configuration file - on opsdroid we have a cli command to do this: `opsdroid edit config`, perhaps we should include it here

docs/examples/flip-a-coin.md
```diff
@@ -0,0 +1,42 @@
...
+
+```yaml
+skills:
+    flip-a-coin:
+    path: ~/skill-flipacoin
+```
```
FabioRosado, line 42: I think we could add another section to explain how to run this skill - my reasoning behind this is that folks reading this example will be new users and they won't know how opsdroid works exactly.

We could do something like:

- Save your configuration, restart opsdroid
- To run the skill send the message "flip a coin" to your connector to see the result.

docs/examples/flip-a-coin.md
```diff
@@ -0,0 +1,42 @@
+# Flip A Coin
+
+A skill for opsdroid to perform flipping a coin.
+
+## Setting Up
+
+Navigating to 'opsdroid/skills/' there is a file called '__init__.py' in which will contain the functions for opsdroid to flip a coin.
```
FabioRosado, line 7: I'm happy if you want to add your skills inside the `opsdroid/skills` folder, but I think we should update this line a bit. An opsdroid skill is just a python function/class that is imported from somewhere.

I'd say that we could change this to something like:

"create a folder with __init__.py that we will use on opsdroid configuration." 

You could then say that perhaps using the path `opsdroid/skills` could be a good place to put your skills, but opsdroid does some file handling and will load the modules into its own directory after they are installed.

FabioRosado (changes requested):
Hello @cristinasofia 
I'd like to thank you for this PR and for wanting to contribute to the project.

This is a good starting point, but I feel we could improve this example a bit more, here's some thoughts:

- Explain how to set up a new skill
- Add links to the docs 
- Explain how to open the `configuration.yaml`

I've added a few comments, let me know if you agree or would like to comment any of them 😄 👍

PR #1641, opened 2020-10-01
Handle KeyError exception when requesting twitch token

FabioRosado:
# Description

Hello,

when running opsdroid from a clean installation, I've noticed that the method `request_oauth` of the Twitch connect was giving me a KeyError on line 191 - `self.token=data["access_token"]`

The reason for this, was that I wrote the redirect URL wrong, but we should really handle any status code != 200

This PR fixes this bug and logs a warning when data doesn't contain the access_token.

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Tox - all green
- Ran opsdroid - works as expected

# Checklist:

### DIFF
--- a/opsdroid/connector/slack/__init__.py
+++ b/opsdroid/connector/slack/__init__.py
@@ -152,7 +152,8 @@ class ConnectorSlack(Connector):
                 data["thread_ts"] = message.linked_event.event_id
 
         await self.slack.api_call(
-            "chat.postMessage", data=data,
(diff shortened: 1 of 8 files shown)


FabioRosado:
So I've tested this with the latest version of black and everything seems fine. What do you think about updating black to the latest version? I'm asking this since otherwise, I won't be able to fix the PR unless I remove black on my side.

PR #1642, opened 2020-10-01
Add a SIGHUP handler to reload opsdroid

### LINKED ISSUE #1631: Add a SIGHUP handler to reload opsdroid
Many linux daemons respond to the [`SIGHUP`](https://en.wikipedia.org/wiki/Signal_(IPC)#SIGHUP) signal by reloading their configurations. Combined with opsdroid's `no_cache: true` this could be used to reload a skill (or other plugin) while developing it.

This should be added in the same manner as the [existing handlers](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/core.py#L56-L60).

MrNaif2018:
# Description

This PR adds ability to reload opsdroid by SIGHUP signal.

Should I also add this to documentation somewhere? In overview page?
Also: from what I see sigint and other signals are not tested, should I add tests for sighup?
Also, glad to see that opsdroid is improving! Lots of changes since my last contributions during previous Hacktoberfest :tada: 

Fixes #1631

## Status
**READY** 

## Type of change

# How Has This Been Tested?

`opsdroid start` in one terminal, in other one `kill -SIGHUP <pid>`
Opsdroid successfully restarted and used new configuration file.
изображение

# Checklist:

### DIFF
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -58,6 +58,9 @@ class OpsDroid:
                 self.eventloop.add_signal_handler(
                     sig, lambda: asyncio.ensure_future(self.handle_signal())
                 )
+            self.eventloop.add_signal_handler(
+                signal.SIGHUP, lambda: asyncio.ensure_future(self.reload())
+            )
         self.eventloop.set_exception_handler(self.handle_async_exception)
         self.skills = []
         self.memory = Memory()
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -85,7 +85,10 @@ class TestCore(unittest.TestCase):
             self.assertTrue(mock_sysexit.called)
 
     def test_signals(self):
-        with OpsDroid() as opsdroid:
+        loop = asyncio.get_event_loop()
+
+        def real_test_signals(opsdroid):
+            asyncio.set_event_loop(loop)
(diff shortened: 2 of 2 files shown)


Cadair:
More tests are never a bad thing!

Thanks for the PR!

MrNaif2018:
Hmm, tox seems to catch sighup signal and make it an error. Is it possible to make it not do that?

FabioRosado:
Hello @MrNaif2018 thank you for raising this PR and happy Hacktoberfest! Glad to see you back, I'm not sure if we can make tox not pick up that signal, perhaps if we mock it or the method it would stop that from happening?

MrNaif2018:
I have fixed the signal test by wrapping all the code in a thread so sending signals to that thread and not main one.
Windows tests fail because SIGHUP is unsupported on windows. Should I add `pytest.skip` inside that test?
Like so:
```
if os.name == "nt":
    pytest.skip("SIGHUP unsupported on windows")
```

FabioRosado:
I think that could be a good way to deal with the failures if SIGHUP isn't supported 👍

MrNaif2018:
Now the signals test is skipped on windows, so test suite passes

FabioRosado (approved):
Thank you so much for updating the PR 👍

PR #1648, opened 2020-10-02
Migrate helper and connector tests to pytest

### LINKED ISSUE #1502: Migrate tests from unittest to pytest

tardigrde:
# Description

Migrating test_helpers.py and test_conenctor.py to pytest.

Fixes #1502 

## Status
**READY**

## Type of change

Test

# How Has This Been Tested?

With pytest on local machine.

# Checklist:


opsdroid/tests/test_connector.py
```diff
@@ -0,0 +1,145 @@
...
+from opsdroid.core import OpsDroid
+from opsdroid.connector import Connector, register_event
+from opsdroid.events import Message, Reaction
+from opsdroid.cli.start import configure_lang
+
+configure_lang({})
+
+
+@pytest.fixture(scope="session")
+def loop():
```
mvoitko, line 16: This fixture should be event_loop
https://github.com/pytest-dev/pytest-asyncio#event_loop
Consider moving this fixture to `conftest.py`

opsdroid/tests/test_connector.py
```diff
@@ -0,0 +1,145 @@
...
+    class MyConnector(Connector):
+        @register_event(Message, include_subclasses=True)
+        def send_my_event(self, event):
+            pass
+
+    c = MyConnector({})
+    assert MyEvent in c.events
+
+
+"""Test the async methods of the opsdroid connector base class."""
```
mvoitko, line 84: Why is the docstring is in the middle of the file?

opsdroid/tests/test_connector.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def test_connect(loop):
+    connector = Connector({}, opsdroid=OpsDroid())
+    with pytest.raises(NotImplementedError):
+        loop.run_until_complete(connector.connect())
+
+
+def test_listen(loop):
+    connector = Connector({}, opsdroid=OpsDroid())
```
mvoitko, line 41: Consider creating a connector fixture to remove code duplication

mvoitko (changes requested):

opsdroid/tests/test_connector.py
tardigrde: yeah, sure. then we don't even need this fixture. its already implemented for us

opsdroid/tests/test_connector.py
tardigrde: sorry, this is WIP. I wanted to create a draft PR first.

opsdroid/tests/test_connector.py
tardigrde: sure.

opsdroid/tests/test_connector.py
tardigrde: then I put the functions back in classes, that way the docstrings make sense

opsdroid/tests/test_connector.py
mvoitko: According to pytest-asyncio README, we should create a new event loop fixture to avoid using the default one.
https://github.com/pytest-dev/pytest-asyncio/blob/86cd9a6fd2/pytest_asyncio/plugin.py#L169-L174
So we probably need it, but only a definition in pytest

opsdroid/tests/test_connector.py
mvoitko: IMHO the classes are needed only for unittest tests style. Functions are more pytest way

tardigrde: okay, but the classes may refer to classes we test. I can remove the classes, but then do what with the docstrings which indicate what we are testing in the next few tests.

tardigrde:
A core test fails, which I have not touched, and have no idea why. I will have time to look in this after the weekend.

opsdroid/tests/test_connector.py
tardigrde: okay will do.

opsdroid/tests/test_connector.py
```diff
@@ -0,0 +1,116 @@
...
+    def test_event_subclasses(self):
+        class MyEvent(Message):
+            pass
+
+        class MyConnector(Connector):
+            @register_event(Message, include_subclasses=True)
+            def send_my_event(self, event):
+                pass
+
+        c = MyConnector({})
```
FabioRosado, line 61: Could we turn this `c` variable into something else like `connector?`

FabioRosado (changes requested):
Hello @tardigrde  thank you for raising this PR, glad to have you back and happy hacktoberfest! 

I've added a little comment on a choice of a variable name, but so far this looks good. I know you have set it as a draft PR so far so I've just glanced at it briefly. 

Not sure why the test is failing on 3.7 🤔

opsdroid/tests/test_connector.py
FabioRosado: I know that functions are more pytest way, but I think we could have classes to do the tests. If I'm not mistaken we have merged some pytest tests within a class 👍

opsdroid/tests/test_connector.py
FabioRosado: This could be a good idea since we are indeed using connector = Connector... although the only issue I see is that we will have to pass a config to the connector sometimes. 

If we decide to go with the fixture route, we probably will need to update all the other pyttests that test a connector - which is out of scope of this PR

opsdroid/tests/test_connector.py
tardigrde: then I leave in the classes. I think they can be useful for interpreting what we are testing.

opsdroid/tests/test_connector.py
tardigrde: I created a fixture for the ones, that does not need config, like this:
```
@pytest.fixture(scope="session")
def no_config_connector():
    return Connector({}, opsdroid=opsdroid)
```
Maybe it is a bit clunky. Other thing could be a higher-order fixture, for which you can pass a config, and it gives you the connector instance.

opsdroid/tests/test_connector.py
tardigrde: Changed it.

opsdroid/tests/test_connector.py
tardigrde: I am not sure, if this is correct. What I did is only copy the same code as they have, conftest now has this:
```
@pytest.yield_fixture
def event_loop():
    """Create an instance of the default event loop for each test case."""
    loop = asyncio.get_event_loop_policy().new_event_loop()
    yield loop
    loop.close()
```
Do I have to create a new custom event loop? If so, do you have any wishes how should it look like?

tardigrde:
Thank you @FabioRosado!

I think the PR is ready for review. I still have a question regarding event_loop, which I asked above.

opsdroid/tests/test_connector.py
mvoitko: You need to create one fixture with no config by default but the possibility to pass a config as a parameter. This can be achieved writing fixture as a decorator.

tardigrde: I changed the fixture. Can you check if you like it, or if a change has to be made.

FabioRosado (approved):
Thank you for the update, I'm happy with the event loop fixture and since the tests are all green I'm also happy to merge this PR 😄 👍

PR #1650, opened 2020-10-02
Add match_catchall matcher

### LINKED ISSUE #1610: Feature request: Please consider implementing a 'catch-all' responder
# Description
I have a skill with many regex matchers, all starting with some flavor of "botname".  ex: 
```
botname get environment
botname get hostinfo
botname get verbose hostinfo
botname get hosthealth
botname get verbose hosthealth
```

etc, with various complexities of regex applied to capture hostnames, etc.  It would be nice to be able to define a 'catchall' matcher that would fire only if it was the only string matched.  For example, if a user typed the invalid string: 
```
botname gte infohost
``` 
that would match nothing in the above list.  If there was a 'catchall' that was set to match strings for "^botname", it could fire a "print help text" message so long as the string didn't match any of the other responders.  

## Versions
- **Opsdroid version:** 0.19.0
- **Python version:** 3.8.4
- **OS/Docker version:**

## Configuration File
n/a

## Additional Details
Any other details you wish to include such as screenshots, console messages, etc.

MrNaif2018:
# Description

This PR adds a catch-all matcher and relevant tests, as in #1610
Implementation is mostly based on `match_all` implementation

Fixes #1610

## Status
**READY**

## Type of change

# How Has This Been Tested?

Tested with telegram bot, with the following skill:
```
from opsdroid.matchers import match_catchall, match_regex


class HelloSkill(Skill):
    @match_regex(r"hi")
    async def hello(self, message):
        await message.respond("Hey")

    @match_catchall
    async def catchall(self, message):
        await message.respond("Catch-all")

```
If sending hi, it replies with Hey, if sending anything else it replies with Catch-all

# Checklist:


FabioRosado (changes requested):
Hello, thank you for working on this issue.
This looks good, there are two things that we still need to do:

- We need to add this matcher to the documentation
- We need to update these tests to use pytest - we are in the process of moving from unittests #1502

Will you be able to do this?

opsdroid/core.py
```diff
@@ -485,6 +486,8 @@ async def get_ranked_skills(self, skills, message):
                 _LOGGER.debug(_("Checking Rasa NLU..."))
                 ranked_skills += await parse_rasanlu(self, skills, message, rasanlu)
 
+        if isinstance(message, events.Message) and not ranked_skills:
```
Cadair, line 489: This question is one for discussion, I am not really sure one way or another, but do we want this "catch all" to catch things that aren't messages? Should it catch all unhandled events, i.e. joins, images, files etc. My feeling is that if it's going to be named `match_catchall` then it probably should, if it stays as messages only it should be renamed to reflect that?

jacobtomlinson: I think we should have both. A catch all and a catch all messages.

MrNaif2018: So I will add a `messages_only` flag to decorator, defaulting to False?

MrNaif2018:
@FabioRosado I have added documentation and converted tests to pytest format.
The only question left is about how to implement all events matching/only message matching. See my comment above. `messages_only=False` flag?

docs/skills/matchers/catchall.md
```diff
@@ -0,0 +1,21 @@
+# Catchall
+
+The `catchall` matcher is a special case which is called only when no other skills were matched.
+
+It is useful for displaying a help message to the user to show available commands bot replies to, for example.
```
FabioRosado, line 5:
```suggestion
It is useful for displaying a help message to the user to show available commands that the bot replies to, for example:
```

FabioRosado (approved):
I've added a suggestion, it's just to improve the sentence. I think that yeah using the flag to set it to messages only is enough - otherwise, we would have to have some repeated code and I'd rather not have that tbh 

You will need to update the docs with the flag description something like:
"This matches catches everything - messages and events, if you want to catch only messages you can set the flag x to true."

MrNaif2018:
To make it work for both regular events and messages, I have put it to tasks. Should I update all the tests expecting `len(tasks) == 2` to 3?

FabioRosado:
Yeah because the matchers are tasks you will need to update them by incrementing task +1 on the tests :D

MrNaif2018:
@FabioRosado It should pass the test suite now

FabioRosado (approved):
Thank you so much for the update, I'm going to merge this PR now 😄 👍

PR #1660, opened 2020-10-25
Add a config option to disable registering a route for / (#1656)

### LINKED ISSUE #1656: Add a config option to disable registering a route for /
In the web server by default a route is registered for the base / path. This should be configurable in case a user wants to register their own.

villelaitila:
# Description

The intent is to let users to register any route for web index handling instead of
forcing the app to use just / address. New config param `disable_web_index_handler_in_root`
is added to let disabling that root route. The default value is false.

Unit tests have been added to verify that the config variable actually is treated correctly.

The default value for this disabling key is False, which means that disabling / registration
is not done, in order to keep things backwards compatible.

Fixes #1656

## Status
**READY**

## Type of change

- This change requires a documentation update:  Not sure, if documentation is still needed?

# How Has This Been Tested?

- The new unit test tests/test_web verifies that the registration is not applied if this new 
config key is set to True.  
- Tested that adding `disable_web_index_handler_in_root` into YAML file with value
true is translated into Python True boolean value. 

# Checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/opsdroid/opsdroid/1660)

### DIFF
--- a/opsdroid/web.py
+++ b/opsdroid/web.py
@@ -25,7 +25,7 @@ class Web:
         self.web_app = web.Application()
         self.runner = web.AppRunner(self.web_app)
         self.site = None
-        if not self.config.get('disable_web_index_handler_in_root', False):
+        if not self.config.get("disable_web_index_handler_in_root", False):
             self.web_app.router.add_get("/", self.web_index_handler)
             self.web_app.router.add_get("", self.web_index_handler)
 
--- a/opsdroid/web.py
+++ b/opsdroid/web.py
@@ -25,8 +25,10 @@ class Web:
         self.web_app = web.Application()
         self.runner = web.AppRunner(self.web_app)
         self.site = None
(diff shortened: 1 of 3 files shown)


opsdroid/web.py
```diff
@@ -25,8 +25,10 @@ def __init__(self, opsdroid):
         self.web_app = web.Application()
         self.runner = web.AppRunner(self.web_app)
         self.site = None
-        self.web_app.router.add_get("/", self.web_index_handler)
-        self.web_app.router.add_get("", self.web_index_handler)
+        if not self.config.get('disable_web_index_handler_in_root', False):
```
villelaitila, line 28: Self-reviewing my code here now. As YAML does inherently treat booleas strings, should we then compare the value against string "true"  instead of a simple not check?

Cadair (approved):

PR #1662, opened 2020-10-28
Move quickstart to new file and expand it

jacobtomlinson:
Much of our documentation is pretty verbose. Getting started with opsdroid can be a little daunting because there are so many options.

To try and make things better for newbies I decided to write a very opinionated quickstart which doesn't show much of opsdroid power, but does get you going very fast. 

Each time there are multiple options I've tried to choose one and link to the docs on all the others. But without overwhelming the user.

Any input on trimming things down even further would be appreciated.


jacobtomlinson:
This is ready if anyone wants to review and merge.

docs/quickstart.md
```diff
@@ -0,0 +1,122 @@
...
+
+Follow these quick steps to learn the basics and get up and running with Opsdroid.
+
+## Terminology
+
+There are a few terms you will need to be familiar with when using Opsdroid. Here is a quick overview of the
+terms used in this document.
+
+- **Connector** - A module to connect to a chat service like Slack.
+- **Skill** - Skills are functions that you write that get called when something specific happens in your chat service.
```
FabioRosado, line 11: Not sure if it's worth mention that they could be also classes?

```suggestion
- **Skill** - Skills are functions/classes that you write that get called when something specific happens in your chat service.
```

docs/quickstart.md
```diff
@@ -0,0 +1,122 @@
...
+You can find the full [configuration reference here](../configuration).
+
+## Start opsdroid
+
+Now we can run Opsdroid for the first time.
+
+```shell
+$ opsdroid start
+INFO opsdroid.logging: ========================================
+WARNING opsdroid: 'welcome-message: true/false' is missing in configuration.yaml
```
FabioRosado, line 52: I think we should remove that warning log in the code base 😬 
```suggestion

```

FabioRosado (approved):
This looks great and seems pretty easy to follow, I'm sure folks will get much value from following this quickstart instead of getting hit by all the possibilities haha.

I've added some small comments with suggestions to make it easier for you if you agree with them 👍

jacobtomlinson:
Thanks for the suggestions @FabioRosado

FabioRosado:
No worries 😄

PR #1665, opened 2020-10-29
Migrate Telegram connector to webhooks

FabioRosado:
# Description

This PR will introduce a breaking change on the Telegram connector. I've decided to ditch the old way to calling the API - mostly because calling the API every x seconds isn't great, also because we couldn't handle chats outside DM's.

With this PR, we add the possibility to get events from direct messages, groups and channels (the bot needs to be added to the group/channel, so subscribing to these events don't cause any issue if the bot isn't inside any of those)

I've also leveraged the existing events and created new events for the connector, this should allow users to deal with more events coming from Telegram.

Since we will move to webhooks, user's will have to have the `base-url` flag set up. 

I still need to add breaking changes warning in the documentation about this.

Finally, I'm not sure why docs are complaining about the emoji 🤔 

## Status

## Type of change

# How Has This Been Tested?

- Tox - All green
- Ran new connector - opsdroid worked as expected

# Checklist:


FabioRosado:
I might have gone a bit too crazy here, as I wrote the docs about breaking changes, I wonder if I should add that listen method (which was a pain) to provide backwards compatibility for version 0.20.0 and then add the breaking change on 0.21.0? 🤔 

@iobreaker  Apologies for the @, just wanted to check with you if you are still using the connector and if these changes will be a headache for you?

jacobtomlinson:
I'm fine with this being breaking. Anything we can do to warn folks would probably be good, having a warning or error come up if `base-url` is missing is probably enough.

FabioRosado:
I've updated the logging message to mention the breaking changes and also a link to the documentation 👍

.readthedocs.yml
```diff
@@ -9,6 +9,7 @@ python:
       path: .
       extra_requirements:
           - docs
+          - connector-telegram
```
Cadair, line 12:
```suggestion
          - connector_telegram
```

my bad

PR #1669, opened 2020-11-02
Add docs on exposing Opsdroid to the internet and update log message with URL

jacobtomlinson:
I noticed in #1665 that @FabioRosado mentions Ngrok.

I thought it would be helpful to create a doc on exposing Opsdroid to the internet with a variety of tunnelling tools (there are better free options that ngrok). That way we can link to it from wherever we need to mention this and just keep one doc up to date.

I also updated the web server logs to output the `web.base-url` when opsdroid starts if it has been configured.

```console
$ opsdroid start
INFO opsdroid.web: Started web server on https://NGROK_URL.ngrok.io
```

### DIFF
--- a/docs/exposing.md
+++ b/docs/exposing.md
@@ -1,0 +1,393 @@
+# Exposing Opsdroid via tunnels
+
+Many chat connectors require your Opsdroid instance to be accessible on the public internet. This is because
+they send events to Opsdroid by calling HTTP endpoints directly.
+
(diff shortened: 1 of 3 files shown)


FabioRosado (approved):
This looks great! Like I said on our channel, I wanted to raise an issue to address this so it's great that you added the docs already haha

It's also great that you added a lot of different ways to expose opsdroid to the internet, I've been using ngrok and didn't know about the other ones. As you said, they seem a better alternative which I want to try!

PR #1679, opened 2020-11-25
Add Teams connector

jacobtomlinson:
@FabioRosado and I started working on a Teams connector today. 

Things are functional, the bot can receive messages and respond.

Teams is a little weird because the bot has to be @ mentioned in every message in order for it to process it. And to send notification messages to a room the user has to have spoken to the bot in that room at least once. Other than that things are pretty standard.

Imgur

- [x] Tests
- [x] Docs


cognifloyd:
Is the `@` only restriction due to using BotFramework?

FabioRosado:
> Is the `@` only restriction due to using BotFramework?

When I tried to implement this connector I've used the webhooks approach and that was still the only way to trigger the webhook.

Even the docs say that for you to interact with a bot you need to @ it so it must be a design choice

cognifloyd (commented):
Very cool. I just went surfing through the docs to see if I could find anything helpful. Here is what I found.

cognifloyd:
Here's the relevant bit about `@` mentioning bots:
> Bots in channel and group chat conversations require the user to @ mention the bot to invoke it in a channel.
> Bots in a one-to-one conversation do not require an @ mention. All messages sent by the user will be routed to your bot.

https://docs.microsoft.com/en-us/microsoftteams/platform/bots/how-to/conversations/conversation-basics

docs/connectors/teams.md
```diff
@@ -0,0 +1,59 @@
...
+
+databases:
+    sqlite: {}  # Teams assumes one database is configured to preserve state
+```
+
+## Usage
+
+Your bot will only respond to messages where it was mentioned. So take this into account when designing your skills.
+
+The bot can also only talk in channels where is has been spoken to at least once.
```
cognifloyd, line 57:
```suggestion
The bot can also only talk in channels where it has been spoken to at least once.
```

cognifloyd:
OOH! look what I found:
https://docs.microsoft.com/en-us/graph/api/subscription-post-subscriptions?view=graph-rest-1.0&tabs=http

So, if we step outside of the BotFramework supported APIs and use the graph API, it looks like the bot COULD subscribe to all messages in a channel by subscribing to `/teams/{id}/channels/{id}/messages`

opsdroid/connector/teams/teams.py
```diff
@@ -0,0 +1,150 @@
...
+        else:
+            return Response(status=415)
+
+        activity = Activity().deserialize(body)
+
+        # Cache service endpoint for channel
+        teams_channel_id = teams_get_channel_id(activity)
+        if teams_channel_id not in self.service_endpoints:
+            self.service_endpoints[
+                teams_get_channel_id(activity)
```
cognifloyd, line 68:
```suggestion
                teams_channel_id
```

jacobtomlinson:
Interesting. It seems strange that you are required to request additional permissions via some web form that is reviewed by a human. Sounds like it could be inconsistent. I wonder what the criteria are for approving these?

I'm not saying we shouldn't do it. But it should probably be in addition to the regular bot APIs.

opsdroid/connector/teams/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""A connector for Slack."""
```
cognifloyd, line 1:
```suggestion
"""A connector for Teams."""
```

cognifloyd:
Found out why the tests are freezing. Pip is endlessly backtracking.

This is the culprit:
https://github.com/microsoft/botbuilder-python/blob/11b08778beb6f3d31c65fb35b54beabb89fb4bdc/libraries/botframework-connector/setup.py#L12

botframework-connector specifies `PyJWT==1.5.3` which conflicts with ibm-watson's (well ibm-cloud-sdk-core's) requirement: `PyJWT>=1.7.1`

I didn't notice this locally because I have an old version of pip installed so I wasn't using the new pip resolver.

jacobtomlinson:
I found a little time to nudge this forwards.

I ended up adding a test which uses the new test machinery to trigger a message event in the Teams connector. However I struggled to make use of our mock API server for the outbound events as overriding the endpoints didn't seem to have any effect in the bot framework config and it also seems like it takes 6 HTTP requests for the bot framework to just to a credential exchange. 

I decided to try out `vcrpy` here instead which records all the outbound HTTP requests and replays them on subsequent runs of pytest. I temporarily switched out the fake credentials in the tests for my own and things seem ok. VCR seems to have sanitised the recording so the credentials aren't in there. This is the first time we've tried using VCR in opsdroid so thought I would push my changes so far for comment.

ping @opsdroid/maintainers for review

cognifloyd:
stale bot: You're wrong.

pintman:
I hope the feature will be integrated.

opsdroid/connector/teams/teams.py
```diff
@@ -0,0 +1,152 @@
+"""A connector for Slack."""
```
pintman, line 1: "A connector for Microsoft Teams." I guess.

opsdroid/connector/teams/teams.py
```diff
@@ -0,0 +1,152 @@
...
+_LOGGER = logging.getLogger(__name__)
+CONFIG_SCHEMA = {
+    "app-id": str,
+    "password": str,
+    "bot-name": str,
+}
+
+
+class TeamsConnector(Connector):
+    """A connector for Slack."""
```
pintman, line 31: "A connector for Microsoft Teams." I guess.

cognifloyd:
I wonder if there are any hooks we can use to control uuid generation in the teams lib. Maybe a fixture that makes `uuid.uuid()` or similar (whatever it is using to generate the uuid) return something consistent?

jacobtomlinson:
> I wonder if there are any hooks we can use to control uuid generation in the teams lib.

Possibly! I'd rather avoid digging too deep into the teams SDK. I was thinking about trying to write a custom VCR handler which stripped the uuid. But patching `uuid.uuid()` might be a better call.

jacobtomlinson:
Found the `uuid` call that is causing me a headache!

https://github.com/microsoft/botbuilder-python/blob/70fee41f4f122143fd27ac22dc17642642d915c3/libraries/botbuilder-core/botbuilder/core/conversation_reference_extension.py#L16

cognifloyd (commented):
Wow. 😥
Looking at the coverage, I think adding a bit to the test will test the rest of the significant bits.

Also, I'm not sure what the docker/pip test failure is about.

jacobtomlinson:
@cognifloyd you've been great at reviewing here. Do you want to give this a last pass and hit the big green button?

cognifloyd (approved):

PR #1682, opened 2020-11-27
Core changes without new tests or machinery

jacobtomlinson:
These are the core changes from #1558.


tests/test_database_matrix.py
```diff
@@ -101,45 +101,45 @@ async def test_default_config(patched_send, opsdroid_matrix):
     )
 
 
-@pytest.mark.asyncio
```
Cadair, line 104: was this causing trouble?

jacobtomlinson: Yeah

jacobtomlinson:
Thanks for the assistance here @cognifloyd but it doesn't seem to have resolved the problem.

tests/test_database_matrix.py
Cadair: what was the issue?

jacobtomlinson: It was intermittently failing for me locally. But it was a while ago now so I can't remember all the details. It may have been a red herring.

cognifloyd: I re-enabled the test. We'll see if it is still flaky now that I've separated the signal tests.

jacobtomlinson:
This is great thanks @cognifloyd (and @Cadair for resolving the conflict).

Let's take the hit on coverage to get this in.

PR #1687, opened 2020-12-16
Fix  NewRooom, RoomName, JoinRoom, UserInvite, RoomDescription Events for Slack Connector

chillipeper:
# Description

The Slack Classic App has 2 type of tokens, the bot token (which we use to connect to the RTM and post messages) and the user token, which is available for other methods.
The following Events : NewRooom, RoomName, JoinRoom, UserInvite, RoomDescription do not work with the bot token, the slack web client has to be initialized with the user token to be able to make use of them.

This PR fixes that.

## Status
**READY** 

## Type of change

# How Has This Been Tested?

Yes, tests have been added and methods have been tested in a Slack Bot.

# Checklist:

- [X ] I have performed a self-review of my own code
- [X ] I have made corresponding changes to the documentation (if applicable)
- [ X] I have added tests that prove my fix is effective or that my feature works
- [ X] New and existing unit tests pass locally with my changes


jacobtomlinson:
Thanks for raising this @chillipeper.

Given that you need a classic app token, and those are going away at some point, do we want to remove this functionality instead?

If it isn't possible to do via the web API then we probably shouldn't do it.

@Cadair do you have any thoughts on this? I expect these events are useful to you.

chillipeper:
@jacobtomlinson I don't think it would be ok to remove those functionalities. I have a specific use case where I am using them. Actually the new slack app has the same methods, and the bot token supports those methods, the problem atm is that we *need* to use the Slack Classic App, because opsdroid connects trough the RTM API. The classic app support this methods only with the _user token_ as it is the only one that accepts granular permissions.

> If it isn't possible to do via the web API then we probably shouldn't do it.
I don't get this completely, this PR is  executing those methods via the Web API, the difference is that is using a different token (the user token instead of the bot token)

I am happy to modify this PR to get this working.

jacobtomlinson:
Thanks for explaining that @chillipeper, I had misunderstood.

We use classic apps today because we use the RTM API, so this implements things for the classic app. We need to move to new apps and only use the webhooks and events API in the future, but that is a concern for another day.

Happy to merge on passing CI.

chillipeper:
Happy to help when moving to the events API, my only concern with the events API is that it's a requirement to have an external endpoint to receive the events, not everyone has that "luxury", but as you said, that is a discussion for the future.

jacobtomlinson:
> thanks and happy to help when moving to the events API

That would be much appreciated.

> requirement to have an external endpoint to receive the events

Many services are going in this direction. I think our Slack connector is now in the minority of connectors that use an RTM style API.

We even now have docs on exposing your Opsdroid ot the internet. https://docs.opsdroid.dev/en/latest/exposing.html

Issue #1676, opened 2020-11-23
E2EE - Matrix Connector


ironoa:
# Description
I would need to be able to use E2EE with the Matrix connector. My skill is fully working in an unencrypted room (read from the chat, write to the chat), but in an encrypted room my skill is only capable of writing to the chat.

Here it's mentioned that the E2EE is supported: https://docs.opsdroid.dev/en/latest/connectors/matrix.html
That documentation appears only if you select version "latest", if you select v0.19.0 it disappears. But in DockerHub the v0.19.0 and the latest tags seems to coincide ( 5 months ago ) https://hub.docker.com/r/opsdroid/opsdroid/tags

## Support Request
1) Is E2EE on the matrix connector supported on version v0.19.0 or not ?
2) Installing the "matrix-nio[e2e]" and the "libolm" libraries is not straightforward in the opsdroid/opsdroid:v0.19.0 docker image. Could you please help me with that ?
3) After having installed the two libraries mentioned above, should I expect the E2EE functionality starts to work out of the box ? If not, could you guide me please ? 

## Expected Functionality
My skills should be capable of read messages even from encrypted matrix rooms, as it is for the unencrypted rooms.

## Experienced Functionality
My skills can only write messages to encrypted rooms. That messages appear with a warning: Unencrypted 
![image](https://user-images.githubusercontent.com/25957738/99947218-ef0e9300-2d77-11eb-9de4-e7b9cc60f91b.png)

## Versions
- **Opsdroid version: v0.19.0**
- **Docker version: version 19.03.8**

## Configuration File
Please include your version of the configuration file below.

```yaml
web:
  host: '0.0.0.0'
  port: 8080
databases:
  sqlite:
    path: "/app/db/finance.db"  
    table: "finance"  
skills:
  - name: eventmanager
    path: /app/skills/skill-eventmanager
connectors:
  matrix:
    mxid: "XXX"
    password: "XXX"
    rooms:
      "main": "myEncryptedRoomId:XXX.XXX"
    homeserver: XXX
    store_path: "/app/store/"
```

```Dockerfile
FROM opsdroid/opsdroid:v0.19.0
RUN adduser -D -g '' opsdroid
COPY skills /app/skills
USER opsdroid
```

Thanks

Cadair:
The e2ee support has not made it to a release, it's only implemented in the master branch. The documentation you linked is "latest" which is built from the master branch.

[...]

PR #1690, opened 2020-12-30
Handling encryption with matrix connection in the Dockerfile build

dud1337:
The matrix connection's encryption handling doesn't work with the current dev docker build due to missing dependencies. `nio.crypto.ENCRYPTION_ENABLED` returns `False`

To solve this, 3 dependencies are installed:
* `libffi-dev` via `apk`
* `python-olm` via `pip`
* `matrix-nio[e2e]` via `pip`

I've tested this in an encrypted room.

It should be noted that this by default **doesn't have persistence on the generated device key** and it can be a tad annoying if you name the device the same thing: Matrix doesn't see the new key. Using a volume for the `matrix_store` solves this.

Relates to #1676

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -82,7 +82,7 @@ console_scripts =
 connector_matrix =
 	bleach>=3.1.5
 	matrix-nio>=0.14.1
-	python-olm
+connector_matrix_e2e =
 	matrix-nio[e2e]
 connector_mattermost =
 	mattermostdriver>=7.0.1
--- a/setup.cfg
+++ b/setup.cfg
@@ -82,6 +82,8 @@ console_scripts =
 connector_matrix =
 	bleach>=3.1.5
 	matrix-nio>=0.14.1
+	python-olm
+	matrix-nio[e2e]
 connector_mattermost =
 	mattermostdriver>=7.0.1
 connector_slack =
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ ARG EXTRAS=.[all]
 COPY . .
 
 RUN apk update \
-    && apk add --no-cache build-base linux-headers musl-dev python3-dev libzmq zeromq-dev git openssh-client olm olm-dev \
+    && apk add --no-cache build-base linux-headers musl-dev python3-dev libzmq zeromq-dev git openssh-client olm olm-dev libffi-dev \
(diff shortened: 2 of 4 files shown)


jacobtomlinson:
cc @Cadair

setup.cfg
```diff
@@ -82,6 +82,8 @@ console_scripts =
 connector_matrix =
 	bleach>=3.1.5
 	matrix-nio>=0.14.1
+	python-olm
+	matrix-nio[e2e]
```
Cadair, line 86: I don't think we want this here like this.

There are two things, I think `python-olm` should be a dependency of `matrix-nio[e2e]` anyway, also I think that we don't want to require e2e for installing the matrix connector, I think it needs to be opt-in as it's difficult to install on non-linux platforms at the moment.

What I propose is this:
```suggestion
connector_matrix_e2e =
	matrix-nio[e2e]
```

dud1337:
>which will add a new extra so you would do pip install opsdroid[connector_matrix, connector_matrix_e2e] to get all the e2e deps.

Sounds good. As long as it's in the docker image for my laziness :]

I'll say that in the long run, from the matrix perspective, having e2e as "optional" feels wrong, but I understand re: installation difficulties on other platforms.

Cadair:
> I'll say that in the long run, from the matrix perspective, having e2e as "optional" feels wrong

I agree, and the packaging situation is improving for the olm library and bindings, but basically until it's on PyPI it's impractical for us to install it by default.

Cadair (approved):

Cadair:
@jacobtomlinson Are the test fails here expected?

Cadair (changes requested):
Well I am going to assume this test fail isn't expected :stuck_out_tongue: 

https://github.com/opsdroid/opsdroid/pull/1690/checks?check_run_id=1645033142#step:5:463

Cadair:
It seems that it's not finding olm when installing python-olm?

Cadair:
I looked into the error and got a little carried away, I hope you don't mind.

I have changed two things:

1. I excluded the new `connector_matrix_e2e` from being included in the `all` extra.
2. I skipped the install step for the docker builds to save a little time.

Cadair (approved):

dud1337:
Thanks; I was dreading fixing the docker builds on windows anyway.

PR #1694, opened 2021-01-05
Test out enabling 3.9

Cadair:
Add 3.9 to tox and GH actions.


FabioRosado (approved):
Got to love those green ticks :D  nice work on updating the badges. The only thing I could add here is that perhaps we should pick up #1702 and mention that 3.9 is now supported 😄

Cadair:
@FabioRosado done.

PR #1698, opened 2021-01-05
better handling for OSError in opsdroid.web.start

cognifloyd:
# Description

Windows tests were apparently broken in #1682 .
This makes sure that site.start() is called safely (ie site needs to be recreated before being started)
And then test to make sure OSError gets raised correctly when the port is in use.

## Status
**DONE**

## Type of change

# How Has This Been Tested?

I don't have windows, so I'm using GHA to test this on windows.
Locally, the new test works for linux as well.

# Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,7 @@ jobs:
         if: runner.os != 'Windows'
         run: |
           if [ $RUNNER_OS = Linux ]; then
+              sudo apt update || true
               sudo apt install ffmpeg -y
           fi
       - name: Install ffmpeg on windows
--- a/opsdroid/tests/test_web.py
+++ b/opsdroid/tests/test_web.py
@@ -1,7 +1,10 @@
 """Test the opsdroid web."""
+import contextlib
+import socket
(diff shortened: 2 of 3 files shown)


cognifloyd:
I don't get it. This failed on master branch, but this commit passes on windows?
https://github.com/cognifloyd/opsdroid/runs/1652773158?check_suite_focus=true

jacobtomlinson:
Race conditions perhaps? I'll kick the CI and see if it comes up the same.

cognifloyd:
I think this is the right fix. Now to come up with a test that hits the OSError case.

A test would probably mean starting something to block the port, so that we are guaranteed to hit the OSError, and make sure the behavior is sane.

cognifloyd:
K. I had to fix the ubuntu tests (an issue in the azure ubuntu repo mirror, nothing to do with this PR), but now all tests pass.
Also, this fixes the bug I found where hitting OSError would call site.start() repeatedly which wasn't safe to do.

jacobtomlinson (approved):
Awesome thanks!

PR #1705, opened 2021-01-15
Improve matrix connector and database documentation

Cadair:


tox.ini
```diff
@@ -45,7 +45,7 @@ commands =
      docker rmi opsdroid-image:tmp
 
 [testenv:docs]
-basepython = python3
+basepython = python3.8
```
Cadair, line 48: I don't know if this is a good idea or not, but the build is broken with 3.9.

docs/databases/matrix.md
```diff
@@ -58,11 +59,29 @@ async def skill_func(opsdroid, config, message):
...
+```eval_rst
+.. autofunction:: opsdroid.database.matrix.memory_in_event_room
+```
+
+## How it Works
 
-In encrypted Matrix rooms, state events (used by the database) are not encrypted. This means that anything put into the matrix database in an encrypted room would not be encrypted. To prevent this the matrix database send the **values** you put into your database into the room as regular events, which are encrypted, and then references these events from the database (which is still a state event).
+### State Events and State Keys
+
+Matrix state events have a type, which for the opsdroid matrix database is `dev.opsdroid.database`, a "state key" which distinguishes state events with the same type, and content which is a json object.
```
SolarDrew, line 70: Honestly I struggle to parse this sentence a little but that may just be because I'm tired.

tox.ini
SolarDrew: We can always change it back if and when 3.9 doesn't break.

SolarDrew (approved):
A few comments but nothing that I think should stop it merging. Thanks for fleshing this out.

docs/databases/matrix.md
```diff
@@ -67,7 +67,17 @@ async def skill_func(opsdroid, config, message):
 
 ### State Events and State Keys
 
-Matrix state events have a type, which for the opsdroid matrix database is `dev.opsdroid.database`, a "state key" which distinguishes state events with the same type, and content which is a json object.
+Matrix state events are used to store metadata about the room, such as it's members, it's name and topic.
```
SolarDrew, line 70:
```suggestion
Matrix state events are used to store metadata about the room, such as its members, its name and topic.
```

SolarDrew (approved):
Yeah this is much clearer :+1:

PR #1706, opened 2021-01-16
Use pytest marks to add responses to the mock_api fixture

Cadair:
# Description

This solves the pytest puzzle of how to use `ExternalAPIMockServer` in the most pytest way possible.

### DIFF
--- a/opsdroid/connector/github/tests/test_connector_github.py
+++ b/opsdroid/connector/github/tests/test_connector_github.py
@@ -11,9 +11,9 @@ from opsdroid.testing import call_endpoint, running_opsdroid
 
 
 @pytest.fixture
-async def connector(opsdroid, mock_api):
+async def connector(opsdroid, mock_api_obj):
     opsdroid.config["connectors"] = {
-        "github": {"token": "abc123", "api_base_url": mock_api.base_url}
+        "github": {"token": "abc123", "api_base_url": mock_api_obj.base_url}
     }
     await opsdroid.load()
     return opsdroid.get_connector("github")
@@ -68,13 +68,11 @@ async def test_connect_failure(connector, mock_api, caplog):
     assert "Bad credentials" in caplog.text
 
 
-@pytest.mark.add_response("/", "GET")
 @pytest.mark.asyncio
 async def test_disconnect(connector):
     assert await connector.disconnect() is None
 
 
-@pytest.mark.add_response("/", "GET")
 @pytest.mark.asyncio
 async def test_listen(connector):
     """Test the listen method.
--- a/opsdroid/connector/github/tests/test_connector_github.py
+++ b/opsdroid/connector/github/tests/test_connector_github.py
@@ -10,6 +10,12 @@ from opsdroid.matchers import match_event
 from opsdroid.testing import call_endpoint, running_opsdroid
 
 
+# these strings are used in several tests
+ORG, REPO, ISSUE = "opsdroid", "opsdroid", "1"
+ISSUE_TARGET = f"{ORG}/{REPO}#{ISSUE}"
+COMMENTS_URI = f"/repos/{ORG}/{REPO}/issues/{ISSUE}/comments"
+
+
 @pytest.fixture
 async def connector(opsdroid, mock_api_obj):
     opsdroid.config["connectors"] = {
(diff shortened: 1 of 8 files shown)


opsdroid/connector/github/tests/test_connector_github.py
```diff
@@ -42,49 +42,40 @@ def test_missing_token(caplog):
...
-    async def test():
-        await connector.connect()
-        assert "Bad credentials" in caplog.text
-
-    await mock_api.run_test(test)
+    await connector.connect()
+    assert "Bad credentials" in caplog.text
 
 
+@pytest.mark.add_response("/", "GET")
```
Cadair, line 72: I decided to add a "you must register a route" check, but then these tests don't actually use a route, but they still start the mock_api server, so maybe we don't want to enforce that?

opsdroid/connector/github/tests/test_connector_github.py
```diff
@@ -97,76 +88,69 @@ async def test_listen(connector):
     assert await connector.listen() is None
 
 
+@pytest.mark.add_response(
+    "/repos/opsdroid/opsdroid/issues/1/comments", "POST", None, status=201
```
Cadair, line 92: The downside of this approach is having to define this explicitly outside the test, before it was defined from variables used in the tests...

opsdroid/testing/external_api.py
```diff
@@ -122,14 +122,22 @@ def base_url(self) -> str:
         return f"http://{self.host}:{self.port}"
 
     def add_response(
-        self, route: str, method: str, response_path: str, status: int = 200
+        self,
+        route: str,
+        method: str,
+        response: Union[dict, PathLike] = None,
```
Cadair, line 128: I figured you might not always want to load it from a file, so just supporting a dict here would be good?

opsdroid/testing/fixtures.py
```diff
@@ -1,10 +1,42 @@
...
+
+    def _get_connector(config={}):
+        return Connector(config, opsdroid=opsdroid)
+
+    return _get_connector
+
+
+@pytest.fixture
+def bound_address(request):
+    """I have no idea what this does but must have a docstring."""
```
Cadair, line 27: Seriously what does this do?

cognifloyd: pushed some documentation for it :)

opsdroid/testing/external_api.py
cognifloyd: :+1:

opsdroid/connector/github/tests/test_connector_github.py
cognifloyd: At least in this file, with the strings reused in so many tests, I think it was cleaner to pull them out into some constants and build them all once.

cognifloyd (changes requested):
I fixed some issues.

Plus I have one question about our approach with the type of response in `add_response`.

opsdroid/connector/github/tests/test_connector_github.py
cognifloyd: Agreed. I see you dropped that.

opsdroid/testing/external_api.py
```diff
@@ -122,14 +122,22 @@ def base_url(self) -> str:
...
+        status: int = 200,
     ) -> None:
         """Push a mocked response onto a route."""
-        if response_path is not None:
-            with open(response_path) as json_file:
+        if isinstance(response, PathLike):
+            with open(response) as json_file:
                 response = json.load(json_file)
+        elif response is None or isinstance(response, dict):
```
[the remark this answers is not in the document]
cognifloyd: Also, if someone needs to create a more complex response, then they should probably be doing that in a fixture.

```python
@pytest.fixture
def funky_response(mock_api):
    mock_api.add_response("/funky", "GET", status=200, response=generate_funky_object())

def test(mock_api, funky_response):
    # test it
```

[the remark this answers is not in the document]
cognifloyd: (PS: If we drop the ValueError, then codecov.io will be happy)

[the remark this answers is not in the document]
Cadair: Makes sense, have committed the change.

opsdroid/testing/utils.py
```diff
@@ -1,15 +1,36 @@
...
+async def running_opsdroid(opsdroid, start_timeout=1):
+    """Context manager to control when opsdroid is running.
+
+    This async context manager will start opsdroid running for its duration.
+    It can be used like this::
+
+        @pytest.mark.asyncio
+        async def test_with_running_opsdroid(opsdroid):
+            async with running_opsdroid(opsdroid):
+                assert opsdroid.is_running()
```
FabioRosado, line 24: Do we need an empty line below this?

FabioRosado (approved):
This looks good to me, just added a tiny commend on a docstring. 
Great work on this PR folks, the docs make it easy to understand what is expected - +1 for the examples 😄 👍

opsdroid/testing/utils.py
Cadair:
```suggestion
                assert opsdroid.is_running()
                
```
Think so.

opsdroid/testing/fixtures.py
```diff
@@ -19,10 +76,74 @@ def opsdroid() -> OpsDroid:
 
 
 @pytest.fixture
-async def mock_api() -> ExternalAPIMockServer:
-    """Fixture for making a mock API server.
+def mock_api_obj():
+    """Initialize instance of :class:`opsdroid.testing.ExternalAPIMockServer`.
 
-    Will give an instance of :class:`opsdroid.testing.ExternalAPIMockServer`.
+    This fixture returns an instance that hasn't been started, allowing you to
```
Cadair, line 82: @cognifloyd I rewrote this docstring with an example I learnt from #1707

cognifloyd (approved):
:+1:

PR #1709, opened 2021-01-17
Use pre-commit in tox for lint

Cadair:
-------

# Description

Move all the lint checks into pre-commit so the CI holds no more surprises

## Status
**READY**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,3 +8,10 @@ repos:
     rev: v2.3.0
     hooks:
       - id: flake8
+        args: ['--config=setup.cfg']
+  - repo: https://github.com/PyCQA/pydocstyle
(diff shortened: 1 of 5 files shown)


FabioRosado (approved):
This looks good to me, just wanted to double check something that might be a silly question, if we move the lint to pre-commit, does that mean that folks will need to use precommit to get the lint passing?

The only reason why I am asking, is because not everybody has black and precommit installed when contributing 🤔

Cadair:
pre-commit is installed by tox so running tox works the same as before, just there is an extra layer to it. If you do use pre-commit hooks now though you are (in theory) sure the lint will pass.

FabioRosado:
Ah gotcha, thank you for the explanation! I'll merge this in now 😄 👍

PR #1710, opened 2021-01-18
Move most of the magic to the mock_api_obj fixture

Cadair:
# Description

This is a few small improvements to the ExternalAPIMockServer fixtures.

* Move all the setup into the non-running fixture 
* Add a context manager to run the server (as an alternative to `run_test`).

## Status
**READY**

## Type of change

# Checklist:

### DIFF
--- a/opsdroid/testing/fixtures.py
+++ b/opsdroid/testing/fixtures.py
@@ -86,9 +86,8 @@ def mock_api_obj(request) -> ExternalAPIMockServer:
 
         * Use the ``mock_api`` fixture in the test *after* this fixture has
           been called to setup a route.
-        * Use :meth:`opsdroid.testing.ExternalAPIMockServer.run_test`
         * Use the
-          :meth:`opsdroid.testing.ExternalAPIMockServer.running_mock_api`
+          :meth:`opsdroid.testing.ExternalAPIMockServer.running`
           context manager.
         * Manually start and stop the server.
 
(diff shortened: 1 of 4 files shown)


opsdroid/testing/fixtures.py
```diff
@@ -76,17 +76,20 @@ def opsdroid() -> OpsDroid:
...
 
     This fixture returns an instance that hasn't been started, allowing you to
     add routes to the server before it is started.
     There are a few options of how to start the server once this fixture has
     been used:
 
-        * Use `:meth:opsdroid.testing.ExternalAPIMockServer.run_test`
         * Use the ``mock_api`` fixture in the test *after* this fixture has
           been called to setup a route.
+        * Use :meth:`opsdroid.testing.ExternalAPIMockServer.run_test`
```
Cadair, line 89: Should we just remove `run_test` at this point, I personally don't think it's a pattern we should support.

jacobtomlinson (approved):
Woo context managers

opsdroid/testing/fixtures.py
FabioRosado: If we have the context manager, we probably don't need this one 🤔

FabioRosado (approved):
Context Managers are the best thing ever haha great work on this! If we remove `run_test` we need to update the docstring on `ExternalAPIMockServer` since the example contains `run_test`

Cadair:
Pushed a new commit removing run_test and updating the name of the context manager to just `running` as I think it reads better `... with mock_api.running():`

Cadair:
I bumped the codecov threshold now we aren't holding onto 100%, I have found with sunpy that the default of 0.1% can trigger just when you remove some code, 0.2% seems a little more stable.

PR #1715, opened 2021-01-18
Support multiple events in a single sync response

### LINKED ISSUE #1712: Handle multiple events in a single sync response in the matrix connector
# Description

Currently, when parsing a sync response, the matrix connector returns out of the method when it encounters the first event in the timeline.

https://github.com/opsdroid/opsdroid/blob/3fe47d89df719e22861352fc3806d4093832c024/opsdroid/connector/matrix/connector.py#L284-L286

This is then subsequently passed to `opsdroid.parse` in the `listen` method:

https://github.com/opsdroid/opsdroid/blob/3fe47d89df719e22861352fc3806d4093832c024/opsdroid/connector/matrix/connector.py#L306-L309

The fix for this is to allow returning a list of `Event`s from the `_parse_sync_response` method and then iterating over them in `listen`.

Cadair:
# Description

Fixes #1712

## Status
**READY**

### DIFF
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -182,7 +182,6 @@ class ConnectorMatrix(Connector):
 
     async def connect(self):
         """Create connection object with chat library."""
-
         if self._allow_encryption:
             _LOGGER.debug(f"Using {self.store_path} for the matrix client store.")
             Path(self.store_path).mkdir(exist_ok=True)
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -299,14 +299,12 @@ class ConnectorMatrix(Connector):
             ][0]
             sender = await self.get_nick(None, invite_event.sender)
 
-            await self.opsdroid.parse(
-                events.UserInvite(
-                    target=roomid,
-                    user_id=invite_event.sender,
-                    user=sender,
-                    connector=self,
-                    raw_event=invite_event,
-                )
+            yield events.UserInvite(
+                target=roomid,
+                user_id=invite_event.sender,
+                user=sender,
+                connector=self,
+                raw_event=invite_event,
             )
 
         for roomid, roomInfo in response.rooms.join.items():
@@ -320,8 +318,8 @@ class ConnectorMatrix(Connector):
                                 event = self.connection.decrypt_event(event)
                             except nio.exceptions.EncryptionError:  # pragma: no cover
                                 _LOGGER.exception(f"Failed to decrypt event {event}")
-                        await self.opsdroid.parse(
-                            await self._event_creator.create_event(event.source, roomid)
+                        yield await self._event_creator.create_event(
+                            event.source, roomid
                         )
 
     async def listen(self):  # pragma: no cover
@@ -342,7 +340,8 @@ class ConnectorMatrix(Connector):
 
             await self.exchange_keys()
 
-            await self._parse_sync_response(response)
+            async for event in self._parse_sync_response(response):
+                await self.opsdroid.parse(event)
 
     def lookup_target(self, room):
         """Convert name or alias of a room to the corresponding room ID."""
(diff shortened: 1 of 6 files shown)


cognifloyd (approved):

cognifloyd (dismissed):
So, this change looks good, but I'm going to switch my review for now.
Once tests can be migrated (however long that takes), I'll re-review.

SolarDrew:
Honestly I don't think I understand how this change fixes the issue, but the code looks fine.

opsdroid/connector/matrix/connector.py
```diff
@@ -281,8 +281,8 @@ async def _parse_sync_response(self, response):
                             _LOGGER.error(
                                 f"Failed to decrypt event {event}"
                             )  # pragma: nocover
-                        return await self._event_creator.create_event(
```
Cadair, line 284: @SolarDrew this return causes the loop to exit on the first event it encounters. By not returning and calling `opsdroid.parse` directly in the loop we don't terminate the loop early.

SolarDrew: Ah ok, that makes sense. And I guess that's what I get for only looking at the diff.

cognifloyd: I don't see why, but if there was some reason to not call parse in this context, you could also year this as a generator and yield the message to parse, returning when the loop exits.

Cadair: That might work quite well.

SolarDrew: Yeah that's an idea. @jacobtomlinson how would that mesh with your views on beginner-friendliness of the codebase?

Cadair: I have implemented this, it's nice.

Cadair:
I decided to not port the tests forward in this as it was pretty trivial to update the existing ones.

Cadair:
Just realised i need to write a test to assert the multiple event behaviour.

opsdroid/connector/matrix/connector.py
```diff
@@ -270,14 +270,12 @@ async def _parse_sync_response(self, response):
...
-                    user=sender,
-                    connector=self,
-                    raw_event=invite_event,
-                )
+            yield events.UserInvite(
+                target=roomid,
+                user_id=invite_event.sender,
+                user=sender,
+                connector=self,
+                raw_event=invite_event,
```
SolarDrew, line 278: So just because I don't really understand `yield` that well: when does this then come back to the function and run the stuff below? Does the function resume from here next time it's called?

tests/test_connector_matrix.py
```diff
@@ -460,7 +460,12 @@ async def test_parse_sync_response(self, connector):
                 nio.ProfileGetDisplayNameResponse("SomeUsersName")
             )
 
-            returned_message = await connector._parse_sync_response(self.sync_return)
+            returned_message = [
+                m async for m in connector._parse_sync_response(self.sync_return)
```
SolarDrew, line 464: I don't think it matters too much in this context but this isn't especially easy to read, for me at least.

SolarDrew: And I don't know what you would do about that anyway. Just an observation really.

tests/test_connector_matrix.py
```diff
@@ -642,7 +644,8 @@ async def _get_message(self, connector):
             patched_nick.return_value = asyncio.Future()
             patched_nick.return_value.set_result("Neo")
 
-            return await connector._parse_sync_response(self.sync_return)
+            async for x in connector._parse_sync_response(self.sync_return):
+                return x
```
SolarDrew, line 648: Will this not just return the first one? Don't you want to `yield` here?

SolarDrew (commented):
A few comments mostly based on my lack of knowledge of how `yield` works. Looks good overall.

opsdroid/connector/matrix/connector.py
Cadair: correct, the function is suspended and resumed from below the yield on the next iteration.

tests/test_connector_matrix.py
Cadair: It will return just the first one, but this test is only expecting a single return anyway so it shouldn't matter. It would be good to refactor all these tests at somepoint anyway.

SolarDrew (approved):
No further objections.

cognifloyd (approved):
🤖

PR #1749, opened 2021-02-22
Drop of RTM API in Slack Connector in favour of events api and socket mode

chillipeper:
# Description

Hello, I have completely refactored the slack connector in order to drop the RTM API.  This way opsdroid will be able to use the new apps from slack with granular permissions, and stop the use of the classic slack app, which will probably be deprecated at some point.

The refactored connector makes use of the Slack Events API.

This is a large change, so I apologize in advance for dropping everything at once. The following is a summary of all the changes that were done:
- Breaking changes to drop of RTM API in favour of the events api.
- Used matrix connector as template to create the modules for slack
- All slack methods from Slack RTM were ported to the Events API
- some attributes, removed from slack connector, as they were not relevant to the new api
-  Drop of unit test in favour of pytest.
-  slack test suite makes use of the testing module from opsdroid (ie: test_external_api, call_endpoint)

changes to the documentation are not done yet.

I am really hoping this could be pushed  to the master branch in the near future.

## Status
READY

## Type of change

# How Has This Been Tested?

- All Tests have done against an instance of opsdroid in Slack.
- Payloads were extracted from an instance of opsdroid in Slack

# Checklist:


jacobtomlinson (commented):
Thanks so much for taking the time to make this port.

We will definitely need some documentation here to help folks get things set up and for me to be able to test things out.

I'm also curious why you completely dropped the RTM API instead of adding this as a second option. While not the latest thing the RTM API is still supported, so I would see a benefit in allowing folks to use it for as long as it exists.

chillipeper:
@jacobtomlinson I have added a draft of the documentation. 

Here are some reasons why I decided _initially_ to completely drop the RTM support:
1.- RTM API still needs Slack Classic Apps which eventually will be deprecated.
2.- Slack Classic Apps lack functionalities, that new apps have such, as granular permissions, and interactive components, we cannot use with RTM
3.- Slack introduced [Socket Mode](https://api.slack.com/apis/connections/socket) to their new Apps, which allow you to connect through web sockets. This would completely replace RTM functionally. I was planning to add that change to my next PR as I am more interested in that part than in the events API, but this is a necessary step.
4.- I saw the Telegram connector was completely refactored, so just went ahead and did the same

On the other side, I do understand your point, and I am not quite sure how to address it. I thought about creating a different connector called (slack_events_api or something similar). But having duplicated code is not something of my liking.

On a side note, if you think adding socket mode support to this same PR is better, I could go ahead with that as well

cognifloyd:
Since OpsDroid + slack currently uses RTM, all OpsDroid + slack setups use a "classic bot" token.
Current OpsDroid users should probably (or might have to) create a new token regular bot token.

Maybe we should use the token type to determine what APIs are available, where the existing `token` is implicitly a "classic bot" token.
https://api.slack.com/authentication/token-types

```yaml
connectors:
  slack:
    # required to use Events api (including in Socket Mode).
    bot-token: "xoxb-abdcefghi-12345"

    # optional (for future possible use)
    app-token: "xox?-xxxxxxxxxxx"

    # This is a "classic bot" token.
    # only required to use RTM apis.
    token: "xoxb-abdcefghi-12345"

    # optional
    user-token: "xoxp-abcdef-123"
```

Perhaps another setting could allow someone to `prefer_api: events`. If `prefer_api` and `bot-token` are not defined but `token` is defined then this would implicitly be `prefer_api: rtm` to match existing behavior. I'm not sold on the name `prefer_api` but that's the general idea (eg something like `prefer_api: events_socket_mode` leaves much to be desired).

jacobtomlinson:
Wow I hadn't seen socket mode! That's super exciting. I would definitely be happier to make a breaking change if we added socket mode in this PR too. 

My primary concern in dropping the RTM API is I expect a bunch of folks to be unhappy about exposing their bot to the internet. And I can already imagine folks saying "but the RTM API still exists, why can't we keep using it?".

By implementing the socket mode it gives us an easy response to those folks, and allows them to keep their bots operating from behind a corporate firewall without exposing it to the internet.

chillipeper:
@cognifloyd  thanks for the suggestions. Since there is no RTM support in this PR, I have only added the `socket-mode` boolean into the configuration, if `socket-mode` is true, then an `app-token` needs to be added

@jacobtomlinson  I have added the socket mode implementation to this PR, and also documentation to it.

Let me know what you think, and again sorry for the huge PR that has to be reviewed

docs/connectors/slack.md
```diff
@@ -19,34 +26,65 @@ connectors:
     # required
     token: "xoxb-abdcefghi-12345"
     # optional
-    user-token: "xoxp-abcdef-123" # default None. ** 
-    bot-name: "mybot" # default "opsdroid"
+    socket-mode: True # defaul false. if true app-token is required
+    app-token: "xapp-abdcfkje-12345" # socket-mode needs to be true
```
Cadair, line 30: Eugh slacks tokens are fun. Thinking about this the other day, I have a proposal. 

Why don't we not reuse the "token" key. This means that any config with `token:` in it would be invalid when used with the Events API and we can then catch this and print out to the user a nice set of instructions on how to update their slack app and how to switch to the events API.

docs/connectors/slack.md
```diff
@@ -19,34 +26,65 @@ connectors:
     # required
     token: "xoxb-abdcefghi-12345"
     # optional
-    user-token: "xoxp-abcdef-123" # default None. ** 
-    bot-name: "mybot" # default "opsdroid"
+    socket-mode: True # defaul false. if true app-token is required
```
Cadair, line 29: Is there any use for the app-token if socket mode isn't being used?

I am of a mind to suggest that we default to `socket-mode: True`, as that's closer to the current behaviour with RTM and `socket-mode: False` needs extra configuration to make the API endpoint publicly accessible.

opsdroid/connector/slack/connector.py
```diff
@@ -0,0 +1,375 @@
...
+            self.bot_id = self.user_info["user"]["profile"]["bot_id"]
+
+            if self.socket_mode_client:
+                self.socket_mode_client.socket_mode_request_listeners.append(
+                    self.socket_event_handler
+                )
+                await self.socket_mode_client.connect()
+            else:
+                self.opsdroid.web_server.web_app.router.add_post(
+                    "/connector/{}".format(self.name),
```
Cadair, line 89:
```suggestion
                    f"/connector/{self.name}",
```

opsdroid/connector/slack/connector.py
```diff
@@ -0,0 +1,375 @@
...
+                "block_actions",
+                "message_action",
+                "view_submission",
+                "view_closed",
+            ):
+                event = await self._event_creator.create_event(payload, None)
+            else:
+                _LOGGER.info(
+                    f"Payload: {payload['type']} is not implemented. Event wont be parsed"
+                )
```
Cadair, line 133: By default the event creator should return None if you give it an event it doesn't understand. This means that you don't have to add support for new events in more than one place.

Also if you don't have an event to parse you can return early saving you a type check later on.
```suggestion
            else:
                event = await self._event_creator.create_event(payload, None)
                if event is None:
                    _LOGGER.info(
                        f"Payload: {payload['type']} is not implemented. Event wont be parsed"
                    )
                    return
```

opsdroid/connector/slack/tests/test_connector.py
```diff
@@ -0,0 +1,341 @@
...
+        response = await connector.send(event)
+        payload = mock_api.get_payload(api_endpoint)
+
+        return payload, response
+
+    return _send_event
+
+
+@pytest.mark.asyncio
+class TestConnectorSlack:
```
Cadair, line 40: Can you drop the class? I much prefer pytest style tests to just be functions.

opsdroid/connector/matrix/connector.py
```diff
@@ -234,8 +234,6 @@ async def connect(self):
                 "Configuration for the matrix connector should specify mxid and password or access_token."
             )  # pragma: no cover
 
-        self.connection.sync_token = None
```
Cadair, line 237: I am not sure why these changes from #1752 are in this diff :confused:

Cadair (commented):
Thanks so much for this @chillipeper it looks really good.

My main concern is making this as easy to adapt to as possible. I am on the side of not maintaining both this and the RTM API, so I think we should have a way of stopping opsdroid when the slack connector loads with an old RTM API configuration and nicely explaining to the user how to reconfigure their slack app and their opsdroid config.

chillipeper:
@Cadair thanks for the very quick response on the review. I  will try to review them this week and fix whatever possible.

opsdroid/connector/matrix/connector.py
chillipeper: I will try to remove this from this PR. I was merging the master branch to this branch, but my fat fingers actually rebased the branch instead of merging.

opsdroid/connector/slack/connector.py
chillipeper: the event starts already as None on line 118. So I don't see the point on calling again the event_creator.

I do agree on returning early to save type checks.

opsdroid/connector/slack/connector.py
```diff
@@ -0,0 +1,376 @@
...
+from slack_sdk.socket_mode.request import SocketModeRequest
+from slack_sdk.socket_mode.response import SocketModeResponse
+from slack_sdk.web.async_client import AsyncWebClient
+from voluptuous import Inclusive, Required
+
+_LOGGER = logging.getLogger(__name__)
+CONFIG_SCHEMA = {
+    Required("token"): str,
+    Inclusive("socket-mode", "backend"): bool,
+    Inclusive("app-token", "backend"): str,
```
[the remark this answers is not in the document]
cognifloyd: Maybe something like this:
```python
_use_app_token_msg=(
    "Please upgrade your config with a new Slack 'app-token' instead of the classic 'token'. "
    "To access the new Slack events APIs, OpsDroid needs new-style Slack app tokens. "
    "See the docs for instructions on creating the new Slack app and configuring the 'app-token'."
    "<url here>"
)
CONFIG_SCHEMA = {
    Exclusive("token", "token", msg=_use_app_token_msg): str,
    Exclusive(Required("app-token"), "token", msg=_use_app_token_msg): str,
    "socket-mode": bool,
```

I'm not sure if Exclusive and Required can be used together like that.

opsdroid/connector/slack/connector.py
```diff
@@ -0,0 +1,376 @@
...
+                "message_action",
+                "view_submission",
+                "view_closed",
+            ):
+                event = await self._event_creator.create_event(payload, None)
+            else:
+                _LOGGER.info(
+                    f"Payload: {payload['type']} is not implemented. Event wont be parsed"
+                )
+                return
```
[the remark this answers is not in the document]
chillipeper: ohh ok, I missunderstood the suggestions as well then, I will take a look at this

docs/connectors/slack.md
```diff
@@ -19,34 +26,65 @@ connectors:
...
+
+If you are unsure which one is the best for you, [Slack Faq](https://api.slack.com/faq#events_api) provide differences between those two.
+
+**Events API**
+
+* You will need an **endpoint** which exposes your Opsdroid to the **internet**. [Exposing Opsdroid via tunnels](https://docs.opsdroid.dev/en/latest/exposing.html) might help out.
+* Go to your [Slack App](https://api.slack.com/apps)
+* On the left column go to "Event Subscriptions" and toogle the "Enable Events"
+* Under "Request URL" add the `/connector/slack` uri to your endpoint: https://slackbot.example.com/connector/slack. Note that you will have to have your Opsdroid instance running so Slack can validate the endpoint.
+For Socket Mode, you will nee
```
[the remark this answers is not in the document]
chillipeper: removed

chillipeper:
Hello,

I have made some changes to the slack connector now.  Socket mode defaults now to true, which requires an app token. I have chosen to use _logger for the message, as this should be an error for users that upgrade from previous versions. Also made some changes to the documentation

@cognifloyd  I see the comments you posted are outdated from it's the current version. We can't mark token as deprecated, as it is still needed for the new apps besides the app-token

cognifloyd:
Ah! I was mistaken on what app-token was for.
Let's rename `token` to `bot-token` so we don't use the same option that was used in old versions of OpsDroid for the classic Slack tokens.

opsdroid/connector/slack/connector.py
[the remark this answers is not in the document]
cognifloyd:
```python
_use_bot_token_msg=(
    "Please upgrade your config with a new Slack 'bot-token' instead of the classic 'token'. "
    "To access the new Slack events APIs, OpsDroid needs new-style Slack app tokens. "
    "See the docs for instructions on creating the new Slack app and configuring the 'app-token'."
    "<url here>"
)
CONFIG_SCHEMA = {
    Exclusive("token", "token", msg=_use_bot_token_msg): str,
    Exclusive(Required("bot-token"), "token", msg=_use_bot_token_msg): str,
    Inclusive("socket-mode", "backend"): bool,
    Inclusive("app-token", "backend"): str,
```

docs/connectors/slack.md
chillipeper: socket-mode now defaults to True

opsdroid/connector/slack/connector.py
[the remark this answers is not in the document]
chillipeper: It's not possible to use `Exclusive`  that way. I have removed the `token` as required and replaced it with `bot-token` this way if `token` exists in the config it will throw an error, asking to update config

opsdroid/connector/matrix/connector.py
chillipeper: This should be removed now on the last commit

chillipeper:
The PR should be ready with all the changes/suggestions requested.

thanks once again to all for reviewing this

cognifloyd (approved):
I've only read through the code, but it looks good to me.

docs/connectors/slack.md
```diff
@@ -2,51 +2,88 @@
...
 
 * A Slack account
-* Create a [new classic Slack App](https://api.slack.com/apps?new_classic_app=1) give it a name and select the workspace you would like it in.
-  * Select `Bots` option inside the `Add features and functionality` tab
-  * Click Add Legacy Bot User and give it a name and a username
-  * Navigate to the "Install App" section and install the app in your workspace.
-  * Take note of the "Bot User OAuth Access Token" as this will be the `token` you need for your configuration (the token will start with `xoxb-`).
+* Create a [new Slack App](https://api.slack.com/apps) give it a name and select the workspace you would like it in.
+* Select "Bots" option inside the "Add features and functionality" tab
+* Click "Review Scopes to Add". "Under Scopes" --> "Bot Token Scopes select" `chat:write` or `chat:write.customize` (Note that you are required to select at least one scope to install the app)
```
jacobtomlinson, line 17: Also needs `users:read`.

docs/connectors/slack.md
```diff
@@ -2,51 +2,88 @@
...
+* Copy your new token add add it to your opsdroid configuration file as your `app-token`. (the app-token will start with xapp-abcdef-1233)
+
+**Events API**
+* Make sure to set `socket-mode` to `false`
+* You will need an **endpoint** which exposes your Opsdroid to the **internet**. [Exposing Opsdroid via tunnels](https://docs.opsdroid.dev/en/latest/exposing.html) might help out.
+* Go to your [Slack App](https://api.slack.com/apps)
+* On the left column go to "Event Subscriptions" and toogle the "Enable Events"
+* Under "Request URL" add the `/connector/slack` uri to your endpoint: https://slackbot.example.com/connector/slack. Note that you will have to have your Opsdroid instance running so Slack can validate the endpoint.
+
+### Subscribe to events
```
jacobtomlinson, line 58: Is this part of the Events API section? 

Could you user headers instead of bold text here so the hierarchy is correct.

jacobtomlinson (changes requested):
I've just taken this PR for a quick spin. I'm afraid I couldn't get it working.

I followed the new instructions and created a new app. I enabled socket mode and configured the two token.

When starting opsdroid I can see an error about the Slack connector not being available, even though a few lines down it seems to say it has connected to the socket.

```
$ opsdroid start      
INFO opsdroid.logging: Started opsdroid v0.21.0+36.g4b78a54.
INFO opsdroid: You can customise your opsdroid by modifying your configuration.yaml.
WARNING opsdroid.loader: No databases in configuration. This will cause skills which store things in memory to lose data when opsdroid is restarted.
ERROR opsdroid.connector.slack.connector: The Slack Connector will not be available.
INFO opsdroid.connector.slack.connector: Connecting to Slack.
INFO slack_sdk.socket_mode.aiohttp: A new session has been established
INFO opsdroid.connector.slack.connector: Connected successfully with socket mode
```

Also if I check Slack the bot is there but appears to be offline and doesn't respond to messages.

What am I missing?

opsdroid/connector/slack/connector.py
```diff
@@ -0,0 +1,388 @@
...
+            else None
+        )
+        self.auth_info = None
+        self.user_info = None
+        self.bot_id = None
+        self.known_users = {}
+
+        self._event_creator = SlackEventCreator(self)
+
+        _LOGGER.error(_("The Slack Connector will not be available."))
```
chillipeper, line 74: @jacobtomlinson  the Error: ```ERROR opsdroid.connector.slack.connector: The Slack Connector will not be available.``` you are getting is because of this line. It's not needed, not quite sure how it slipped there but I will remove it.

docs/connectors/slack.md
chillipeper: I think this might be the problem why you are not getting a response from opsdroid. This section belongs to the events API **and** to the socket mode. Meaning you have to subscribe to the events you want to receive from slack regardless of the backend you are using.

If want opsdroid to receive messages that go into a public channel or a dm to opsdroid you will need to subscribe to these events:

image

If this was your problem, then it means documentation was a bit hard to follow and might need some tweaking.  Suggestions are welcome :)

docs/connectors/slack.md
chillipeper: Fixed

opsdroid/connector/slack/connector.py
chillipeper: Fixed

hkonala:
@jacobtomlinson I have pulled the latest changes and tested the connectivity with the new slack apps and it is working fine. Still figuring out why the methods in AsyncWebClient are not available in SocketModeClient though. Wanted to use chat_postEphemeral method but not able to (maybe don't know how to do) do with SocketModeClient . But worked with AsyncWebClient 
@chillipeper It would be awesome if you could put out a simple example in documentation on how to use methods in AsyncWebClient

chillipeper:
@hkonala, I am afraid you can't use postEphemeral method, because it's not implemented on this PR.  It would be certainly a nice enhancement feature to have. Maybe you could create an issue for it?  I might try to implement it once this PR get's merged

hkonala:
> @hkonala, I am afraid you can't use postEphemeral method, because it's not implemented on this PR. It would be certainly a nice enhancement feature to have. Maybe you could create an issue for it? I might try to implement it once this PR get's merged

Sure thing @chillipeper . Meanwhile could you help me figure out what are the methods I can use with "SocketModeClient"? I may be doing something wrong here but not able to figure out how to use "SocketModeClient" methods. I am able to use "api_call" method with  AsyncWebClient just fine

chillipeper:
@hkonala we can try to figure out what is the problem you are facing. For this can you please create an Issue, with the problem you are experiencing, errors, steps to reproduce, and config you are using? I just don't want to do this over this PR as I don't want to hinder the review. We can reference later any findings

docs/connectors/slack.md
chillipeper: @jacobtomlinson gentle ping, just curious if you were able to test again by adding the events

jacobtomlinson: Sorry I haven't managed to yet. Time is a little tight for me at the moment. Hoping to get to this over the weekend.

jacobtomlinson:
Sorry I took so long getting back here to review. I've tested it again and this time subscribed to the necessary events. Things seem ok!

image

I wonder if we could check which events we are subscribed to and if there are some supported in opsdroid which are not subscribed whether we could log some warnings or something. But we can leave that for a follow up.

Awesome work here @chillipeper thanks for all your efforts.

jacobtomlinson (approved):

chillipeper:
Thanks so much for merging this to the master branch! Will keep in mind on the checks for events subscribed, I am still unsure how can we check against the API for scopes.

PR #1752, opened 2021-02-26
Fix matrix connector sync_token book keeping and log on successful connect

Cadair:
# Description

This makes the matrix connector error out cleaner if it can't connect to the HS

## Status
**READY**

## Type of change


SolarDrew (approved):
:+1:

PR #1754, opened 2021-03-03
Added joined_rooms method

### LINKED ISSUE #1732: Add a method on the matrix connector which lists all rooms the account is joined to
Somewhat related to #1724 but it's also useful for skills to be able to get a list of all joined (not just configured) rooms.

I am thinking a signature like:

```python
def joined_rooms(self, encrypted_only=False):
```
which returns room_ids for the rooms. Richer nio room objects could then be obtained by doing `connector.connection.rooms[room_id]`.

jerinjtitus:
Signed-off-by: Jerin J Titus <72017981+jerinjtitus@users.noreply.github.com>

# Description

Fixes #1732 

## Status
**UNDER DEVELOPMENT** 

## Type of change

# How Has This Been Tested?

It's not tested yet. Need to make a test for it.

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -385,11 +385,16 @@ class ConnectorMatrix(Connector):
             return
         else:
             room_ids = room_ids.rooms
-            if encrypted_only == True:
+            if encrypted_only is True:
                 for room_id in room_ids:
-                    if isinstance(await self.connection.room_get_state_event(room_id, "m.room.encryption"), nio.responses.RoomGetStateError):
+                    if isinstance(
+                        await self.connection.room_get_state_event(
+                            room_id, "m.room.encryption"
+                        ),
+                        nio.responses.RoomGetStateError,
+                    ):
                         room_ids.remove(room_id)
-                        
+
             return room_ids
 
     @staticmethod
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -378,6 +378,25 @@ class ConnectorMatrix(Connector):
 
         return room
 
+    async def joined_rooms(self, encrypted_only=False):
+        """Returns a list of room_id of all joined rooms"""
+
+        room_ids = await self.connection.joined_rooms()
+
+        if isinstance(room_ids, nio.responses.JoinedRoomsError):
+            _LOGGER.error(
+                f"Error while getting list of joined rooms from room state: {room_ids.message} (status code {room_ids.status_code})"
+            )
+            return
+        else:
+            if encrypted_only == True:
+                room_ids = room_ids.rooms
+                for room_id in room_ids:
+                    if isinstance(self.connection.room_get_state_event(room_id, "m.room.encryption"), nio.responses.RoomGetStateError):
+                        room_ids.remove(room_id)
+                        
+            return room_ids
+
     @staticmethod
     def _get_formatted_message_body(message, body=None, msgtype="m.text"):
         """
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -386,14 +386,16 @@ class ConnectorMatrix(Connector):
 
         room_ids = room_ids.rooms
         if encrypted_only is True:
+            encrypted_room_ids = []
             for room_id in room_ids:
                 if isinstance(
                     await self.connection.room_get_state_event(
                         room_id, "m.room.encryption"
                     ),
-                    nio.responses.RoomGetStateError,
+                    nio.responses.RoomGetStateEventResponse,
                 ):
-                    room_ids.remove(room_id)
+                    encrypted_room_ids.append(room_id)
+            room_ids = encrypted_room_ids
 
         return room_ids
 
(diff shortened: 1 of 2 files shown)


opsdroid/connector/matrix/connector.py
```diff
@@ -378,6 +378,25 @@ def get_roomname(self, room):
...
+        room_ids = await self.connection.joined_rooms()
+
+        if isinstance(room_ids, nio.responses.JoinedRoomsError):
+            _LOGGER.error(
+                f"Error while getting list of joined rooms from room state: {room_ids.message} (status code {room_ids.status_code})"
+            )
+            return
+        else:
+            if encrypted_only == True:
+                room_ids = room_ids.rooms
```
Cadair, line 388: Otherwise room_ids isn't defined if `encrypted_only` is False.

```suggestion
            room_ids = room_ids.rooms
            if encrypted_only == True:
```

opsdroid/connector/matrix/connector.py
```diff
@@ -373,6 +373,25 @@ def get_roomname(self, room):
...
+                for room_id in room_ids:
+                    if isinstance(self.connection.room_get_state_event(room_id, "m.room.encryption"), nio.responses.RoomGetStateError):
```
Cadair, line 390:
```suggestion
                    if isinstance(await self.connection.room_get_state_event(room_id, "m.room.encryption"), nio.responses.RoomGetStateError):
```

Cadair:
Looks like you are on the right lines here. I think it will work (given the comments above) now you just need some tests to prove it :grinning:

opsdroid/connector/matrix/connector.py
jerinjtitus: Yes, my bad.

jerinjtitus:
> Looks like you are on the right lines here. I think it will work (given the comments above) now you just need some tests to prove it

Thanks for the help

jerinjtitus:
image
Is this change necessary?

opsdroid/connector/matrix/connector.py
```diff
@@ -373,6 +373,25 @@ def get_roomname(self, room):
...
+        """Returns a list of room_id of all joined rooms"""
+
+        room_ids = await self.connection.joined_rooms()
+
+        if isinstance(room_ids, nio.responses.JoinedRoomsError):
+            _LOGGER.error(
+                f"Error while getting list of joined rooms from room state: {room_ids.message} (status code {room_ids.status_code})"
+            )
+            return
+        else:
```
Cadair, line 386: You also don't need this else: block as this code will not be executed if the early return happens. So you can dedent this section.

jerinjtitus: Yes. Fine I will change it

jerinjtitus:
So I was going through the tests as a template, like @Cadair suggested, and I decided to look up docs for the various things used in the test. 

image
 
About that, I found `call_count` and `get_request` have the same description. The former can be used to find the number of calls for that API route, but the latter seems to be completely different. (I didn't understand completely what "request" meant. I think it's something unique to every call made)
So I just want to make sure if the description for the latter is right or not.

jerinjtitus:
So, I realized there was a problem with the `encrypted_only` logic.
image

I was removing the `room_id` from `room_ids` list while it was looping. It causes some problems as the loop happens with the help of indices and thus I shouldn't change `room_ids` in the loop.
Example:
image

I was making a test for this feature and I am not sure what I should specify for `@pytest.mark.matrix_connector_config`.
> [yeah, without filtering by encrypted the test should only make one API call so it should be easy enough to provide a canned response to the /joined_rooms endpoint.](https://github.com/opsdroid/opsdroid/issues/1732#issuecomment-788946741)

What I am doing is just asserting `mock_api.called("/_matrix/client/r0/joined_rooms", "GET")` after adding the response
```
@pytest.mark.add_response(
    "/_matrix/client/r0/joined_rooms", "GET"
)
```
That's all?

Cadair:
> I am not sure what I should specify for @pytest.mark.matrix_connector_config.

Using one of the config fixtures would be fine: https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/matrix/tests/conftest.py#L30

>  after adding the response

I would actually add the json response (from memory something like: `{'joined_rooms': ['roomid', 'roomid']}`, but check with the spec) and then assert that the method gives you those roomids out at the end.

jerinjtitus:
> So I just want to make sure if the description for the latter is right or not.

What about this?

jerinjtitus:
> I would actually add the json response (from memory something like: {'joined_rooms': ['roomid', 'roomid']}, but check with the spec) and then assert that the method gives you those roomids out at the end.

Thanks for the reply. I looked into it and added the test. But I am receiving an AssertionError. It means either the method is not working or I am still missing out something in the tests.

image

Cadair:
I don't think you need to worry about calling `connect()` or providing the fixtures that it needs (filter, sync and whoami), you do need to call the new joined rooms method though :wink:

jerinjtitus:
image
Something like this?
I got this an error:
TypeError: object MagicMock can't be used in 'await' expression

Cadair:
I think you want to use the [`connector_connected`](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/matrix/tests/conftest.py#L66) fixture instead.

PR #1765, opened 2021-04-10
[WIP] Search Channel id by name

### LINKED ISSUE #1699: Slack Connector, search a channel ID, by its name
# Description
Currently, if we need to send any message to a Slack Channel, we have to know the channel ID. Unfortunately, this is not always possible, and it would be nice to have a way that Opsdroid translates channel names to channel ids.

I have been searching on the Slack API and there is no method that would do this, so the only thing I can think of, is iterating over all the channels. (probably when Opsdroid starts the connector?)  https://api.slack.com/methods/conversations.list
I would like to submit a PR for this, and I would love your thoughts, suggestions on how could this be better implemented.

## Steps to Reproduce
N/A

## Expected Functionality
```
await self.opsdroid.send(Message("Hello", target="test-channel"))
```

## Experienced Functionality
This will throw an error:
```
await self.opsdroid.send(Message("Hello", target="test-channel"))
```
Only works when channel ID is mentioned:
```
await self.opsdroid.send(Message("Hello", target="CF3ZP8KS8"))
```

## Versions
- **Opsdroid version:** All versions
- **Python version:** All Versions 
- **OS/Docker version:** All Versions

chillipeper:
# Description

Fixes #1699 

## Status
**UNDER DEVELOPMENT** |

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:

### DIFF
--- a/docs/connectors/slack.md
+++ b/docs/connectors/slack.md
@@ -33,11 +33,14 @@ connectors:
     icon-emoji: ":smile:" # default ":robot_face:" **
     default-room: "#random" # default "#general"
     start_thread: false # default false. if true, opsdroid will start a thread when replying to a message
+    channel_limit: 100 # default 100. ***
 ```
 
-\* when `socket-mode` is true, you need to set also an `app-token` (more info: [app level tokens](https://api.slack.com/authentication/token-types#app)
+\* when `socket-mode` is true, you need to set also an `app-token` (more info: [app level tokens](https://api.slack.com/authentication/token-types#app))
+
 ** In order for `bot-name` and/or `icon-emoji` to work, the `chat:write.customize` scope will have to be selected
 
+\*** [Maximum channels](https://api.slack.com/methods/conversations.list#arg_limit) to return on a single iteration.
 ### Choose the Backend API
 
 You need to choose between two backends. The [Events API](https://api.slack.com/apis/connections/events-api) or [Socket Mode](https://api.slack.com/apis/connections/socket).
--- a/opsdroid/connector/slack/connector.py
+++ b/opsdroid/connector/slack/connector.py
@@ -7,17 +7,18 @@ import ssl
 
 import aiohttp
 import certifi
-import opsdroid.events
 from emoji import demojize
-from opsdroid.connector import Connector, register_event
-from opsdroid.connector.slack.create_events import SlackEventCreator
-from opsdroid.connector.slack.events import Blocks, EditedBlocks
 from slack_sdk.errors import SlackApiError
 from slack_sdk.socket_mode.aiohttp import SocketModeClient
 from slack_sdk.socket_mode.request import SocketModeRequest
 from slack_sdk.socket_mode.response import SocketModeResponse
 from slack_sdk.web.async_client import AsyncWebClient
 from voluptuous import Required
+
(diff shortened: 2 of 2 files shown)


chillipeper:
@jacobtomlinson I have created this PR to try to tackle #1699 . I have opened it in an early stage because I am looking for some advice on how can I handle a particular problem:

I have created the `get_channels` method, which grabs the Slack Channels from the API, and also created the `find_channel` method which returns the channel information if the channel exists. What I can't figure out is how can we update/cache on a regular basis the channels.  It sounds expensive to query Slack API and retrieve all channels every time a channel is not found in `known_channels`.  (Imagine if there are thousands of channels. )

So I thought of two alternative options:

1.- Retrieve the list of channels, every x amount of time 
2.- Subscribe to an event which will notify the bot everytime a channel is created (this one requires extra config from the user so might not be ideal)

The problem is that I am not quite sure how to implement either of those. Do you have any thoughts?

Any suggestions are much appreciated

FabioRosado:
Is it right to assume that the channel list will mostly stay the same?

If that is the case we could implement some caching method (or use lru_cache) to keep track of this channel list that would prevent us from calling the API all the time 🤔

chillipeper:
thanks for the suggestion @FabioRosado . I would say the first assumption is right. channel list does not changes very often, but it does change from time to time. 

I have never used `lru_cache` in the past, will browse trough the docs to see how can it be applied here.

I have an additional question, does `lru_cache` will be capable of updating the list every X amount of  time? or how could we achieve that?

FabioRosado:
The lru_cache or any other caching function will cache the results of a function a way for you to do this could be:

- Append the cache decorator to the method that queries the channels
- Call the method that updates the channels ever so often (this will make the cache to be used and not query slack's API)
- After a set number of seconds/minutes (up to you or the user) we could clean the cache and then query slack's API

You could try using the `your_cached_method.cache_clear()` if you would rather implement the cache function yourself, [this answer on stackoverflow](https://stackoverflow.com/a/40274320) might be a good way to implement it (then you can remove the entry from the list).

Let me know if that make sense or if you need help

chillipeper:
Thanks @FabioRosado  for the pointers, they do make sense, but need to dig a bit deeper into the docs. Will do that in the following days.

I have an additional question I am not able to understand:
> After a set number of seconds/minutes (up to you or the user) we could clean the cache and then query slack's API

Any pointers or snippets on how can I achieve that?

FabioRosado:
That is a good question, I wonder if we could use the crontab to update this 🤔  @jacobtomlinson  do you think it would be possible?

Otherwise we would have to rely on a while true and some sleep 🤔

jacobtomlinson:
Sorry about not responding to pings here.

You would definitely leverage the crontab timer to trigger an update.

Thinking about the two options presented by @chillipeper you could probably do both right?

Subscribe to a channel created event which adds new entries to the cached list. And then reload the list if one comes up that is missing. The whole list will be grabbed on startup anyway right?

chillipeper:
hey @jacobtomlinson  sorry for the late reply here. I haven't had time to look into this, but I have an idea on how to approach it. Let's keep it opened, I will try to spend some cycles on this

PR #1767, opened 2021-04-13
Add Rich to logging

FabioRosado:
# Description

This PR is a WIP to add rich to our logging, since I was touching the logging file I did some refactor and hopefully this is looking a bit better, the one thing worth to mention is that we will now pass `config["logging"]` to the `configure_logging` function, not sure if that is alright or not, if you would rather just pass the whole configure we can get just the logging ones from within the `configure_logging` function. The reason why I decided to to that was to cutback on most of our keys inside keys deal.

TODO:

- [ ] Add the walrus operator to refactor the logfile logic
- [x] Update documentation to mention rich
- [x] Update documentation to mention timestamp
- [x] Update documentation to mention user formatter
- [x] Update configuration example to include:
  - timestamp option
  - rich option

This is how the new logs will look like

https://user-images.githubusercontent.com/3131401/116004931-d7745880-a5fc-11eb-9007-e2df365ab9be.mp4

## Status

## Type of change

# How Has This Been Tested?

- Tox all green
- Manual testing

# Checklist:


jacobtomlinson (commented):
This looks really awesome.

Couple of otes:
- Is the on by default? (I think it should be)
  - If yes can you update the `example_configuration.yaml` to reflect the default value.
- Can you update the [config schema](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/configuration/validation.py#L8) with new options?

FabioRosado:
Oh yeah that's a good point! rich will be on by default if the logging is  commented out in the configuration, otherwise it will use one or the other, but yeah it makes sense to set the value for rich as true so no weird thing happens when folks choose to configure the logging further 👍 

Yeah forgot about that, thank you for the reminder I'll update the schema 👍

docs/configuration.md
```diff
@@ -16,19 +16,19 @@ For configuration, opsdroid uses a single YAML file named `configuration.yaml`.
                 `C:\Documents and Settings\<User>\Application Data\opsdroid\opsdroid`
 - System `/etc/opsdroid/configuration.yaml` (*nix only)
 
-_Note: If no file named `configuration.yaml` can be found on one of these folders one will be created for you taken from the [example configuration file](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/configuration/example_configuration.yaml)_
+_Note: If no file named `configuration.yaml` can be found on one of these folders, one will be created for you taken from the [example configuration file](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/configuration/example_configuration.yaml)_
 
-If you are using one of the default locations you can run the command `opsdroid config edit` to open the configuration with your favorite editor (taken from the environment variable `EDITOR`) or the default editor vim.
+Suppose you are using one of the default locations. In that case, you can run the command `opsdroid config edit` to open the configuration with your favourite editor (taken from the environment variable `EDITOR`) or the default editor vim.
```
cognifloyd, line 21: LOL. British vs US English, "favourite" vs "favorite" - another battle in the spelling wars.

docs/configuration.md
```diff
@@ -146,37 +146,38 @@ See [module options](#module-options) for installing custom databases.
...
+If set to true, then a welcome message is printed in the log at startup. It defaults to true.
 
 ```yaml
 welcome-message: true
 ```
 
 ### Logging
 
-Configure logging in opsdroid.
+Configure logging in opsdroid. If you don't have logging settings configured, opsdroid will use rich logging.
```
cognifloyd, line 157: > If you don't have logging settings configured

Which settings are "logging settings"? Anything under `logging.*`? For example, will configuring a `logging.path` disable `rich` logging? What about the `whitelist` or `blacklist`? Or are you only talking about `logging.formatter` (which you've documented below)? Will configuring `logging.formatter` disable `rich` or will the setting be ignored?

opsdroid/logging.py
```diff
@@ -60,71 +61,87 @@ def filter(self, record):
...
-        with contextlib.suppress(KeyError):
-            file_handler.addFilter(ParsingFilter(config, config["logging"]["filter"]))
+    if not config or config.get("rich"):
+        handler = RichHandler(
+            rich_tracebacks=True,
+            show_time=config.get("timestamp", True),
+            show_path=config.get("extended", True),
+        )
+
+    if config.get("console") or config.get("console") is False:
```
cognifloyd, line 132: This is an odd condition. Are you looking for a true or a false but not None?
what about something like:
```suggestion
    if config.get("console") is not None:
```

But shouldn't `console: false` be defined when using rich? I'm confused. lol

opsdroid/tests/test_logging.py
```diff
@@ -174,26 +164,40 @@ def test_configure_default_logging(capsys):
     opsdroid.configure_logging(config)
     rootlogger = logging.getLogger()
     assert len(rootlogger.handlers), 2
-    assert isinstance(rootlogger.handlers[0], logging.StreamHandler)
+    assert isinstance(rootlogger.handlers[0], logging.handlers.RotatingFileHandler)
     assert rootlogger.handlers[0].level == logging.INFO
-    assert isinstance(rootlogger.handlers[1], logging.handlers.RotatingFileHandler)
+    assert isinstance(rootlogger.handlers[1], RichHandler)
```
cognifloyd, line 169: wait. what? if `console: true` then we end up with a `RichHandler`?

cognifloyd (commented):
I'm a bit confused about how the logging config is supposed to work. Could you help me understand better?

docs/configuration.md
FabioRosado: If you install opsdroid, the logging in the configuration file is commented out. If you don't specify any logging you get rich logging.

Similarly before the PR if you don't specify any logging you get the console log.

opsdroid/logging.py
FabioRosado: No, because you might want to use both - console logging is also used to log to file. Before this PR, you can set console log to None to just log to file and nothing in the console. Maybe the way we are doing it is a bit confusing

opsdroid/tests/test_logging.py
FabioRosado: You can have both on, the Rich logging will take precedence by default, but you might want to have console logging to log to a file for example

FabioRosado:
Our logging might be a bit confusing. Console log is enabled by default, then you can overwrite the default or extend it with the `logging` configuration parameter. For example you can get the full path where the log  was called.

In this PR I didn't change how the logging works, just added Rich logging and selected that one as default. Rich logging has some things baked in mostly to present logs in a nicer way. You can activate both logs if you want, for example you might want to use the console log to overwrite some things or use the filtering option which Rich doesn't seem to accept.

cognifloyd:
How can console and rich logging work at the same time?
Here's where the RichHandler is created:
https://github.com/opsdroid/opsdroid/blob/6068071ce23f514c4861d282c1950618a5264682/opsdroid/logging.py#L125-L126
And then the handler gets replaced here:
https://github.com/opsdroid/opsdroid/blob/6068071ce23f514c4861d282c1950618a5264682/opsdroid/logging.py#L132-L133

Only the last one will get a filter:
https://github.com/opsdroid/opsdroid/blob/6068071ce23f514c4861d282c1950618a5264682/opsdroid/logging.py#L139-L140
Only the last one will be added to the `rootlogger`:
https://github.com/opsdroid/opsdroid/blob/6068071ce23f514c4861d282c1950618a5264682/opsdroid/logging.py#L141-L143

If they're supposed to work together, then perhaps they should be called `rich_handler` and `console_handler`?
And then add the filter to both of them, and add both handlers to `rootlogger`?

jacobtomlinson:
It looks like @FabioRosado has some review feedback from @cognifloyd to address here. But I'm keen to get this merged.

cognifloyd:
I'm still a bit confused by some of the code, but I'm looking forward to getting this in as well. I've really liked another project's use of rich.

FabioRosado:
Finally picked this up again. Looking at the code, I realised that the two handlers that should work together are either:

- console and filehandler
- rich and filehandler

I've made changes to the code and to some tests. We are also setting Rich by default unless the user specifies `rich: False` in their configuration.  If the user sets rich to false and doesn't specify `console` then we will keep the same behaviour as before - logging only critical level logs to the console.

opsdroid/logging.py
```diff
@@ -60,71 +61,92 @@ def filter(self, record):
...
+        file_handler = RotatingFileHandler(
+            logfile_path, maxBytes=config.get("file-size", 50e6)
+        )
         file_handler.setLevel(log_level)
         file_handler.setFormatter(formatter)
+        rootlogger.addHandler(file_handler)
 
-        with contextlib.suppress(KeyError):
-            file_handler.addFilter(ParsingFilter(config, config["logging"]["filter"]))
+    if config.get("console"):
```
cognifloyd, line 132: So, given these configs will rich logging be enabled?
```
{"path": False, "console": True, "rich": True}
```
```
{"path": False, "console": True}
```

opsdroid/tests/test_logging.py
```diff
@@ -21,43 +22,41 @@ def test_set_logging_level():
...
+        config = {"path": os.path.join(tmp_dir, "output.log"), "console": False}
+
     opsdroid.configure_logging(config)
     rootlogger = logging.getLogger()
     assert len(rootlogger.handlers), 2
     assert isinstance(rootlogger.handlers[0], logging.StreamHandler)
-    assert rootlogger.handlers[0].level, logging.CRITICAL
-    assert isinstance(rootlogger.handlers[1], logging.handlers.RotatingFileHandler)
-    assert rootlogger.handlers[1].level == logging.INFO
+    assert isinstance(rootlogger.handlers[0], logging.handlers.RotatingFileHandler)
```
cognifloyd, line 42: Hmm. How can `rootlogger.handlers[0]` be both a `logging.StreamHandler` (line 41) and `logging.handlers.RotatingFileHandler` (new line 42)?

cognifloyd (commented):
That improves things. thanks. I have a couple more questions.

opsdroid/logging.py
FabioRosado: We are using a single handler. The default is rich, but if you set both console and rich to true then you get console. Since both Rich and Console are logging to the terminal I don't see a point in setting both to True 

The second example, you will get console

cognifloyd: Ok. Can we add a warning if both are enabled explaining that console takes precedence?

opsdroid/tests/test_logging.py
FabioRosado: Apparently because the RotatingFileHanlder is a StreamHandler. Either way, I changed that to check if the second item is the rich handler.

opsdroid/tests/test_logging.py
```diff
@@ -21,43 +22,52 @@ def test_set_logging_level():
...
-            }
+            "path": os.path.join(tmp_dir, "output.log"),
+            "console": False,
+            "rich": False,
+            "filter": {"blacklist": "opsdroid.logging"},
         }
     opsdroid.configure_logging(config)
     rootlogger = logging.getLogger()
     assert len(rootlogger.handlers) == 2
     assert isinstance(rootlogger.handlers[0], logging.StreamHandler)
```
cognifloyd, line 68:
```suggestion
    assert isinstance(rootlogger.handlers[1], RichHandler)
```

opsdroid/tests/test_logging.py
```diff
@@ -21,43 +22,52 @@ def test_set_logging_level():
...
 def test_configure_file_logging():
     with tempfile.TemporaryDirectory() as tmp_dir:
-        config = {
-            "logging": {"path": os.path.join(tmp_dir, "output.log"), "console": False}
-        }
+        config = {"path": os.path.join(tmp_dir, "output.log"), "console": False}
+
     opsdroid.configure_logging(config)
     rootlogger = logging.getLogger()
     assert len(rootlogger.handlers), 2
```
cognifloyd, line 40:
```suggestion
    assert len(rootlogger.handlers) == 2
```

opsdroid/tests/test_logging.py
```diff
@@ -21,43 +22,52 @@ def test_set_logging_level():
...
 
 
-def test_configure_file_blacklist(capsys):
+def test_configure_file_logging_only():
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        config = {"path": os.path.join(tmp_dir, "output.log"), "rich": False}
+
+    opsdroid.configure_logging(config)
+    rootlogger = logging.getLogger()
+    assert len(rootlogger.handlers), 1
```
cognifloyd, line 52:
```suggestion
    assert len(rootlogger.handlers) == 1
```

opsdroid/tests/test_logging.py
```diff
@@ -174,26 +175,40 @@ def test_configure_default_logging(capsys):
     opsdroid.configure_logging(config)
     rootlogger = logging.getLogger()
     assert len(rootlogger.handlers), 2
```
cognifloyd, line 177:
```suggestion
    assert len(rootlogger.handlers) == 2
```

opsdroid/logging.py
```diff
@@ -60,71 +61,92 @@ def filter(self, record):
...
+    if not handler:
+        handler = logging.StreamHandler()
+        handler.setFormatter(formatter)
+        log_level = get_logging_level("critical")
+
+    if config.get("filter") and handler:
+        handler.addFilter(ParsingFilter(config, config["filter"]))
+    if handler:
+        handler.setLevel(log_level)
+        rootlogger.addHandler(handler)
```
cognifloyd, line 148:
```suggestion
        rootlogger.addHandler(handler)
        if config.get("console") and config.get("rich"):
            _LOGGER.warning("Both console and rich logging are enabled. Using console logging because console takes precedence over rich logging.")

```

cognifloyd (changes requested):

opsdroid/tests/test_logging.py
FabioRosado: That suggestion doesn't make sense, I am setting both console and rich to false and I am testing if we are logging to the file. 

Also after spending time looking into this, the FileHandler creates a streamHandler which uses the file as a stream output so this test seems good

opsdroid/logging.py
FabioRosado: I've tried to add a test for this and even though I see the log, when the test is run this is never shown both with caplog or capsys and I'd rather not block this PR with a failure of 100% coverage for this warning since logging would still work just not as pretty

cognifloyd (approved):
LGTM

PR #1768, opened 2021-04-18
Added codescanalert handling

### LINKED ISSUE #1764: Add code scanning alert event to github connector
Once #1753 gets merged, we should really add the `code_scanning_alert` event so opsdroid can alert us when an alert is raised. You can see the [github documentation](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#code_scanning_alert) on this event and the payload example to get an idea of what could be useful including in the event.

Ni3dzwi3dz:
# Description

Added features for handling Codescan alert. Changes include:

events.py:
-added class CodeScanAlertOpened, that handles created, reopened_by_user,  appeared_in_branch, or reopened actions
-added class CodeScanAlertClosed, that handles closed_by_user or fixed actions

__init__.py:
-added method handle_codescan_alert 

test_connector_github.py:
-added test_codescan_alert method

Added example payload

Added

Fixes #1764 

## Status
**READY** 

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:

- [x ] I have performed a self-review of my own code
- [x ] I have added tests that prove my fix is effective or that my feature works


FabioRosado (changes requested):
Hello  @Ni3dzwi3dz  thank you for working on this issue and creating the PR, you will need to add some logic to the `github_message_handler` method, since this is the method that is handling all the events. The reason why we have the other methods to handle specific events is to keep the huge if/else statements at a minimum (at least on the message handler method)

https://github.com/opsdroid/opsdroid/blob/5cf444246107d541392c8c4b7e3f6216729eb2eb/opsdroid/connector/github/__init__.py#L211

I believe the CI failures are related to the fact that opsdroid doesn't know how to parse this event since its not being handled my the message handler.

I hope this isn't too much to ask, but could you reformat the docstrings? we tend to use double-quotes everywhere, we use black to automate the lint fixes on git commit (you can check our [contributing documentation](https://docs.opsdroid.dev/en/stable/contributing/workflow.html#automatic-linting-with-black) that talks about this)

PR #1788, opened 2021-06-27
Add Slack BlockAction match_event for action_id and block_id.

ajitdsa:
# Description

In [this block actions example](https://docs.opsdroid.dev/en/stable/connectors/slack.html#block-actions) you can match a BlockAction by its value. However, you cannot match more generically by `action_id` or `block_id`.

So I added entities to the `block_action` which would allow this:

```
@match_event(BlockActions, action_id="some-action-id")
@match_event(BlockActions, action_id="some-action-id", value="some-value")
@match_event(BlockActions, block_id="some-block-id")
@match_event(BlockActions, block_id="some-block-id", value="some-value")
```

Fixes (no issue previously created for this. Referenced in this [Matrix conversation](https://matrix.to/#/!CVBVDaPiHphjTxEKuI:matrix.org/$1624830304132708KinFF:matrix.org?via=matrix.org&via=chat.mountainview.theworkpc.com&via=web3.foundation).)

## Status
**READY**

## Type of change

# How Has This Been Tested?

I created a Slack app and then added a block with a select menu from this [Slack API example](https://api.slack.com/reference/block-kit/block-elements#select) and added it to my channel.

Then I implemented the following matchers in my skill at match the example:
```
@match_event(BlockActions, action_id="text1234")
@match_event(BlockActions, action_id="text1234", value="value-0")
@match_event(BlockActions, block_id="section678")
@match_event(BlockActions, block_id="section678", value="value-0")
```
and made sure the proper selections were matched as I expected.

# Checklist:

(I'm sorry, I'm not sure what documentation changes or automated tests would be applicable here, but would welcome any suggestions, or additions to my PR.)

### DIFF
--- a/opsdroid/connector/slack/create_events.py
+++ b/opsdroid/connector/slack/create_events.py
@@ -211,7 +211,9 @@ class SlackEventCreator(events.EventCreator):
                 target=event["channel"]["id"],
                 connector=self.connector,
             )
-            action_value = None
+            action_value =  None
+            action_id = action.get("action_id", None)
+            block_id = action.get("block_id", None)
 
             if action["type"] == "button":
                 action_value = action["value"]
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Thanks for this change. Examples of using features like this are always welcome in the docs if you fee like submitting another pull request.

PR #1793, opened 2021-07-05
Add ephemeral response_type to the Slack connector InteractiveAction respond method

ajitdsa:
# Description

This is not a complete fix. It seems to work only in the case where we are responding to an `InteractiveAction` for the Slack connector where the message being sent is a string. It would be much better if we were able to specify an ephemeral `response_type`, whether a string or Blocks or anything else. 👈  I'm not sure how to accomplish this and would appreciate some help. Once we can get this working for a response, then I think we can do it easily for an original message as well.

The ability to send ephemeral messages is a key part of building an interactive Slack app as it allows a single user to have great interactivity prior to taking an action that would result in posting a message to a channel. This allows you to write entire UI interfaces in Slack, which is quite an expansion from just a chatbot.

Here's what I've been able to accomplish so far.

- Added in `kwargs` to the `respond` method
- Set `response_type` variable based on `kwarg` with same name
- Set `response_type` based on value passed in for the case where `response_event` is a `str`

## Status
**UNDER DEVELOPMENT**

## Type of change

# How Has This Been Tested?

I will need help writing tests for this.

# Checklist:

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,6 +8,7 @@ python:
       extra_requirements:
         - docs
         - connector_telegram
+        - connector_slack
         - database_matrix
         - connector_matrix
 
--- a/docs/connectors/mattermost.md
+++ b/docs/connectors/mattermost.md
@@ -44,3 +44,5 @@ hi
 opsdroid APP [9:06 PM]
 Hi daniccan
 ```
+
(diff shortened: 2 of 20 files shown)


ajitdsa:
I'm going to close this. I want to work on it again later in an improved format 👍

Issue #950, opened 2019-05-16
Versioneer not working in docker container


jacobtomlinson:
Due to the way we are constructing the docker container (not adding the `.git` directory) versioneer is unable to calculate the correct version (`0+unknown`).

We should figure out a way, maybe with multi-stage builds, to correctly retain the version info.

PR #1794, opened 2021-07-09
Implement multi-stage Docker build

rgreinho:
# Description

Updates the Dockerfile to leverage multi-stage build and ensure that the
runtime contains only the dependencies required to run OpsDroid.

The final image was updated to run as an unprivileged user.

The concepts used to build this new version of the Docker image mostly
come from Bitnami's recommendations:
https://engineering.bitnami.com/articles/best-practices-writing-a-dockerfile.html

The wheel `unversal` flag was removed since the project does not have C
extensions, nor supports Python 2 (see
https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels
for reference).

## Status
**READY**

## Type of change

# How Has This Been Tested?

Building and running the image locally.

```
$ docker run --rm -it opsdroid/opsdroid:snapshot
INFO opsdroid.logging: Started opsdroid 0+unknown.
INFO opsdroid: You can customise your opsdroid by modifying your configuration.yaml.
WARNING opsdroid.loader: No databases in configuration. This will cause skills which store things in memory to lose data when opsdroid is restarted.
INFO opsdroid.loader: Cloning dance from remote repository.
INFO opsdroid.loader: Cloning hello from remote repository.
INFO opsdroid.loader: Cloning loudnoises from remote repository.
INFO opsdroid.loader: Cloning seen from remote repository.
^CINFO opsdroid.core: Received stop signal, exiting.
INFO opsdroid.core: Stopping connector websocket...
INFO opsdroid.core: Stopped connector websocket.
INFO opsdroid.core: Stopping database inmem...
INFO opsdroid.core: Stopped database inmem.
INFO opsdroid.core: Stopping web server...
INFO opsdroid.core: Stopped web server.
INFO opsdroid.core: Stopping pending tasks...
INFO opsdroid.core: Stopped pending tasks.
INFO opsdroid.core: Bye!
INFO opsdroid.core: Exiting application with return code 0.
```

# Checklist:

- [ X] I have performed a self-review of my own code

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,64 @@
-FROM python:3.8-alpine
+FROM python:3.9.6-alpine3.14 as builder
 LABEL maintainer="Jacob Tomlinson <jacob@tomlinson.email>"
+LABEL maintainer="Rémy Greinhofer <remy.greinhofer@gmail.com>"
 
 WORKDIR /usr/src/app
-ARG EXTRAS=.[all,connector_matrix_e2e]
+
+ARG EXTRAS=[all,connector_matrix_e2e]
+ENV DEPS_DIR=/usr/src/app/deps
 
 # Copy source
 COPY . .
 
+# Install build tools and libraries to build OpsDroid and its dependencies.
 RUN apk update \
-    && apk add --no-cache build-base linux-headers musl-dev python3-dev libzmq zeromq-dev git openssh-client olm olm-dev libffi-dev \
-    && pip3 install --upgrade pip \
-    && pip3 install --no-cache-dir $EXTRAS \
-    && apk del build-base linux-headers musl-dev python3-dev zeromq-dev
+  && apk add \
+  build-base \
+  gcc \
+  g++ \
+  libffi-dev \
+  linux-headers \
+  musl-dev \
+  olm-dev \
+  openssh-client \
+  openssl-dev \
+  python3-dev \
+  zeromq-dev \
+  && pip install --upgrade \
+  build \
+  pip \
+  setuptools \
+  setuptools-scm \
+  wheel \
+  && mkdir -p "${DEPS_DIR}" \
+  && pip download --prefer-binary -d ${DEPS_DIR} .${EXTRAS} \
+  && pip wheel -w ${DEPS_DIR} ${DEPS_DIR}/*.tar.gz \
+  && python -m build --wheel --outdir ${DEPS_DIR}
+
+FROM python:3.9.6-alpine3.14 as runtime
+LABEL maintainer="Jacob Tomlinson <jacob@tomlinson.email>"
+LABEL maintainer="Rémy Greinhofer <remy.greinhofer@gmail.com>"
+
+WORKDIR /usr/src/app
+
+ARG EXTRAS=[all,connector_matrix_e2e]
+ENV DEPS_DIR=/usr/src/app/deps
+
+# Copy the pre-built dependencies.
+COPY --from=builder ${DEPS_DIR}/*.whl ${DEPS_DIR}/
+
+# Install Opsdroid using only pre-built dependencies.
+RUN apk add --no-cache \
+  git \
+  libzmq \
+  && pip install --no-cache-dir --no-index -f ${DEPS_DIR} \
+  ${DEPS_DIR}/opsdroid-0+unknown-py3-none-any.whl${EXTRAS} \
+  && rm -rf /tmp/* /var/tmp/* ${DEPS_DIR}/* \
+  && adduser -u 1001 -S -G root opsdroid
 
 EXPOSE 8080
 
-CMD ["opsdroid", "start"]
+# Ensure the service runs as an unprivileged user.
+USER opsdroid
+ENTRYPOINT ["opsdroid"]
+CMD ["start"]
--- a/setup.cfg
+++ b/setup.cfg
@@ -135,7 +135,7 @@ docs =
   docutils==0.16
 
 [wheel]
-universal = 1
+universal = 0
 
 [tool:pytest]
 testpaths = opsdroid tests


Dockerfile
```diff
@@ -1,18 +1,62 @@
-FROM python:3.8-alpine
-LABEL maintainer="Jacob Tomlinson <jacob@tomlinson.email>"
+FROM python:3.9.6-alpine3.14 as builder
+LABEL maintainer="Rémy Greinhofer <remy.greinhofer@gmail.com>"
```
jacobtomlinson, line 2: I'd prefer to keep me as the labelled maintainer of these images.

Dockerfile
```diff
@@ -1,18 +1,62 @@
...
+COPY --from=builder ${DEPS_DIR}/*.whl ${DEPS_DIR}/
+
+# Install Opsdroid using only pre-built dependencies.
+RUN apk add --no-cache \
+  git \
+  libzmq \
+  && pip install --no-cache-dir --no-index -f ${DEPS_DIR} \
+  ${DEPS_DIR}/opsdroid-0+unknown-py3-none-any.whl${EXTRAS} \
+  && rm -rf /tmp/* /var/tmp/* ${DEPS_DIR}/* \
+  && adduser -u 1001 -S -G root nonroot
```
jacobtomlinson, line 55: Can we call this user `opsdroid`?

Dockerfile
```diff
@@ -1,18 +1,62 @@
...
+  setuptools \
+  setuptools-scm \
+  wheel \
+  && mkdir -p "${DEPS_DIR}" \
+  && pip download --prefer-binary -d ${DEPS_DIR} .${EXTRAS} \
+  && pip wheel -w ${DEPS_DIR} ${DEPS_DIR}/*.tar.gz \
+  && python -m build --wheel --outdir ${DEPS_DIR}
+
+FROM python:3.9.6-alpine3.14 as runtime
+LABEL maintainer="Rémy Greinhofer <remy.greinhofer@gmail.com>"
```
jacobtomlinson, line 38: Same here

Dockerfile
```diff
@@ -1,18 +1,62 @@
...
+
+# Copy the pre-built dependencies.
+COPY --from=builder ${DEPS_DIR}/*.whl ${DEPS_DIR}/
+
+# Install Opsdroid using only pre-built dependencies.
+RUN apk add --no-cache \
+  git \
+  libzmq \
+  && pip install --no-cache-dir --no-index -f ${DEPS_DIR} \
+  ${DEPS_DIR}/opsdroid-0+unknown-py3-none-any.whl${EXTRAS} \
```
jacobtomlinson, line 53: This version number looks funny. Is versioneer not working properly still?

jacobtomlinson (commented):
I'm generally in favour of this, thanks for the efforts @rgreinho.

I would be keen to have a review from @Cadair given the Matrix dependencies are some of the most complex here.

Dockerfile
rgreinho: Addressed ✅

Dockerfile
rgreinho: Addressed ✅

Dockerfile
rgreinho: I think it is just because there is no git repository within the container. 

If I run it locally I get something like`opsdroid-0.23.0+1.g6d81c23-py3-none-any.whl`.

jacobtomlinson: Sure, but we want to fix this (see #950).

Switching to multi-stage builds here is a nice opportunity to resolve this problem because you can pull in the whole repo into the build stage and get the correct version in the package.

rgreinho: This is not really relevant for the docker build though, as the package name as no impact on what is built. What would the benefit be? 

#950 seems to be a non-issue and should be closed.

jacobtomlinson: I disagree. The `opsdroid.__version__` variable is also set by versioneer at build time and is used in multiple places. #950 mainly covers ensuring that variable is correct.

In the docker images this gets set incorrectly currently, hence my request to fix it in this PR as part of the awesome work you're doing here.

Dissonant-Tech (changes requested):
@rgreinho Try this out. Should fix #950

Dockerfile
```diff
@@ -1,18 +1,64 @@
...
+
+ARG EXTRAS=[all,connector_matrix_e2e]
+ENV DEPS_DIR=/usr/src/app/deps
+
+# Copy the pre-built dependencies.
+COPY --from=builder ${DEPS_DIR}/*.whl ${DEPS_DIR}/
+
+# Install Opsdroid using only pre-built dependencies.
+RUN apk add --no-cache \
+  git \
```
Cadair, line 52: I think we will still need `olm-dev` here as it's a runtime dep for the e2ee matrix stuff as well as a compile time one.

Cadair: Although presumably we might only need `olm` not `olm-dev`?

setup.cfg
```diff
@@ -135,7 +135,7 @@ docs =
   docutils==0.16
 
 [wheel]
-universal = 1
+universal = 0
```
Cadair, line 138: Let's just delete this whole block?

Dockerfile
```diff
@@ -1,18 +1,64 @@
...
+  && pip install --no-cache-dir --no-index -f ${DEPS_DIR} \
+  ${DEPS_DIR}/opsdroid-0+unknown-py3-none-any.whl${EXTRAS} \
+  && rm -rf /tmp/* /var/tmp/* ${DEPS_DIR}/* \
+  && adduser -u 1001 -S -G root opsdroid
 
 EXPOSE 8080
 
-CMD ["opsdroid", "start"]
+# Ensure the service runs as an unprivileged user.
+USER opsdroid
```
Cadair, line 62: This doesn't interfere with running the container with `--user` right?

jacobtomlinson: Nah it just changes the default user.

Dockerfile
```diff
@@ -1,18 +1,64 @@
...
+
+# Copy the pre-built dependencies.
+COPY --from=builder ${DEPS_DIR}/*.whl ${DEPS_DIR}/
+
+# Install Opsdroid using only pre-built dependencies.
+RUN apk add --no-cache \
+  git \
+  libzmq \
+  && pip install --no-cache-dir --no-index -f ${DEPS_DIR} \
+  ${DEPS_DIR}/opsdroid-0+unknown-py3-none-any.whl${EXTRAS} \
```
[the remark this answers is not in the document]
rgreinho: Oh this could simply be achieved using a wildcard directly:

```python
${DEPS_DIR}/opsdroid-*-py3-none-any.whl${EXTRAS}
```

However that does not fix the bug in #950 because `versioneer` is still unable to determine the version number automatically.

jacobtomlinson:
Superseded by #1820

Issue #1796, opened 2021-07-15
Would like to complete the get, put, and delete functions for mongodb


ajitdsa:
# Description
There is an inherent error with the mongodb `put` function. The way the code is written [here](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/database/mongo/__init__.py#L67-L68) if you supply `data` with an `_id` key, then it will only attempt an update of an existing record within the collection specified by the `key` parameter. However, if one doesn't exist, then you can't create it with that `data` key.

Also currently it appears the way the opsdroid mongodb database API is designed, the user is relegated to creating a new collection with a single record in it. I feel the user should be able to easily put, get, and delete multiple records into a single collection.

There also appears to be a problem with the [delete function](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/database/mongo/__init__.py#L92). I have not been able to get it working as written, but I was able to make a change to the code to get it working.

Finally, there is a small issue with the [memory example](https://docs.opsdroid.dev/en/latest/skills/memory.html#example) in that each of the methods should have a different name, but also it might be helpful to have a mongodb-specific example, as it differs slightly from the general example (although I appreciate the attempt to keep them the same.)

## Steps to Reproduce
1. Add a [mongodb configuration](https://docs.opsdroid.dev/en/latest/databases/mongo.html#configuration) to your configuration file.
2. Add the [remember example skill](https://docs.opsdroid.dev/en/latest/skills/memory.html#example) to your opsdroid. (Remember to change the method names.)
3. I setup [opsdroid-desktop](https://github.com/opsdroid/opsdroid-desktop) for testing and added a [websocket connector](https://docs.opsdroid.dev/en/latest/connectors/websocket.html) to support that.
4. I tried invoking the `remember`, `remind me`, and `forget it` commands from opsdroid desktop and all gave errors for mongodb. For `remember` and `remind me`, small adjustments are needed to the example skill. For `forget it` a change is required in the actual mongodb code.
5. Also, if you change the `remember` skill to swap in this line:
`await self.opsdroid.memory.put("remember_me", {"_id": "1", "value": remember})`
your record will not be added to the `remember_me` collection as you would expect due to the bug I explained above.

## Expected Functionality
Should be able to copy/paste example code (step 4 above) and have it work out of the gate.
Should be able to put multiple records into a collection specifying the `_id` and then retrieve and delete those records by the `_id` field... or anything you want, really.

## Experienced Functionality
[...]

PR #1797, opened 2021-07-15
Add new funtionality to the mongo database module plus a couple small drivebys

ajitdsa:
# Description

Fixes https://github.com/opsdroid/opsdroid/issues/1796

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Tested manually using opsdroid-desktop and the remember skill example.
- Added tests to the mongo tests directory.

# Checklist:


.gitignore
```diff
@@ -90,6 +90,7 @@ celerybeat-schedule
 .env
 
 # virtualenv
+.venv/
```
ajitdsa, line 93: driveby

docs/databases/mongo.md
```diff
@@ -20,3 +20,29 @@ databases:
...
+```python
+from opsdroid.skill import Skill
+from opsdroid.matchers import match_regex
+
+
+class RememberSkill(Skill):
+    @match_regex(r'remember (.*)')
+    async def remember(self, message):
+        remember = message.regex.group(1)
+        await self.opsdroid.memory.put("remember_this", {"value": remember})
```
ajitdsa, line 36: converted `remember` to `{"value": remember}` for mongo example.

docs/databases/mongo.md
```diff
@@ -20,3 +20,29 @@ databases:
...
+        await message.respond("OK I'll remember that")
+
+    @match_regex(r'remind me')
+    async def remind_me(self, message):
+        information = await self.opsdroid.memory.get("remember_this")
+        await message.respond(information["value"])
```
ajitdsa, line 42: added `information["value"]` key for mongo example.

docs/skills/memory.md
```diff
@@ -29,17 +29,17 @@ class RememberSkill(Skill):
     @match_regex(r'remember (.*)')
     async def remember(self, message):
         remember = message.regex.group(1)
-        await self.opsdroid.memory.put("remember", remember)
+        await self.opsdroid.memory.put("remember_this", remember)
```
ajitdsa, line 32: changed to `remember_this` just to differentiate from the variable and function name.

docs/skills/memory.md
```diff
@@ -29,17 +29,17 @@ class RememberSkill(Skill):
...
         await message.respond("OK I'll remember that")
 
     @match_regex(r'remind me')
-    async def remember(self, message):
-        information = await self.opsdroid.memory.get("remember")
+    async def remind_me(self, message):
```
ajitdsa, line 36: changed the function names to make them distinct

opsdroid/database/mongo/__init__.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, config, opsdroid=None):
         self.config = config
         self.client = None
         self.database = None
+        self.query_data = {}
```
ajitdsa, line 40: This was the mechanism I added in order to not modify the method signatures.

opsdroid/database/mongo/__init__.py
```diff
@@ -65,9 +66,13 @@ async def put(self, key, data):
...
-            await self.database[key].update_one({"_id": data["_id"]}, {"$set": data})
-        else:
-            await self.database[key].insert_one(data)
+            response = await self.database[key].update_one(
+                {"_id": data["_id"]}, {"$set": data}
+            )
+            if response.raw_result["updatedExisting"]:
+                return
+
+        await self.database[key].insert_one(data)
```
ajitdsa, line 75: Wrote some code so make it so that `put` either updates the `_id` field if it exists or adds a new record to the collection.

opsdroid/database/mongo/__init__.py
```diff
@@ -77,9 +82,14 @@ async def get(self, key):
...
-            {"$query": {}, "$orderby": {"$natural": -1}}
-        )
+        if self.query_data:
+            return await self.database[key].find_one(
+                {"$query": self.query_data, "$orderby": {"$natural": -1}}
+            )
+        else:
+            return await self.database[key].find_one(
+                {"$query": {}, "$orderby": {"$natural": -1}}
+            )
```
ajitdsa, line 92: If the `query_data` was set, then use that for querying, otherwise just default to the previous way.

opsdroid/database/mongo/__init__.py
```diff
@@ -89,4 +99,7 @@ async def delete(self, key):
 
         """
         _LOGGER.debug("Deleting %s from MongoDB.", key)
-        return await self.database[key].delete_one({"$query": {}})
+        if self.query_data:
+            return await self.database[key].delete_one(self.query_data)
+        else:
+            return await self.database.drop_collection(key)
```
ajitdsa, line 105: if `query_data` exists on the object, then use it for the delete, otherwise delete the whole collection (which includes my fix, since the previous code didn't work.)

opsdroid/memory.py
```diff
@@ -20,6 +20,7 @@ class Memory:
     def __init__(self):
         """Create object with minimum properties."""
         self.databases = []
+        self.query_data = {}
```
ajitdsa, line 23: Also needed to add this in the general memory object. I think it could prove handy for other database implementations as well in similar fashion.

ajitdsa (commented):
Comments to help a reviewer.

.gitignore
jacobtomlinson: Typically this kind of stuff should be in your [global gitignore](http://egorsmirnov.me/2015/05/04/global-gitignore-file.html). But I'm not against adding it here.

jacobtomlinson:
Failures look valid here. 

```
FAILED opsdroid/database/mongo/tests/test_mongo.py::test_delete[config0]
 E           AttributeError: 'NoneType' object has no attribute 'drop_collection'
```

Can you please check that test.

Cadair:
I am pretty sceptical about the addition of `query_data` as an API, can you explain what you are trying to solve with it for me? I don't really know much about mongo.

ajitdsa:
@jacobtomlinson I'm on it.

@Cadair yes absolutely. So it's not really making any changes to the API. I wanted an "alternate" way to query the database for gets and deletes. Since you're not so familiar with mongodb, let me explain a little.

So mongo has the following structure:
databases -> collections -> documents
(a database can have many collections, and a collection can have many documents.)

The way things are set up right now, the `key` provided to these methods corresponds to the name of a collection. In the case of the `put` operation, you are specifying what you would like to store in a document in the collection. Using this method, it is possible to store multiple documents in a collection, as long as the `data` passed in is different for each document (in the current implementation, it means [they should not have the same `_id` key](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/database/mongo/__init__.py#L67-L68).)

However, when retrieving the documents via the `get` operation, the only option you have is to specify a `key` (i.e. the collection), and then per the code written, it will just retrieve the [last document from within the collection](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/database/mongo/__init__.py#L80-L82). There is no way to query for a specific document within the collection.

This works if you use mongodb in a very specific way... you create a bunch of collections and then store one document per collection. But so many existing implementations of mongodb (including mine) are set up with multiple documents per collection, and we would like to query for them.

Same thing with the `delete` operation, BTW. Since the only thing you can specify is the `key` (i.e. the collection name), then the only real option is to delete the entire collection (or the last document within the collection... which if there is only one, will remove the collection per default mongodb behavior.)

So now we finally get to what I have done. I basically just added a `query_data` dict as a property on the Database object such that if you set it prior to calling `put` or `delete`, it will add that into the query so that your query will contain that data. So let's look at an example:

In my collection `albums`, I have the following documents:
```
{
    "_id": "1",
    "title": "Dark Side of the Moon",
    "artist": "Pink Floyd"
}
```
```
{
    "_id": "2",
    "title": "Homework",
    "artist": "Daft Punk"    
}
```
```
{
    "_id": "3",
    "title": "Rumours",
    "artist": "Fleetwood Mac"    
}
```
Currently using the methods available, if I want to query a record from this collection, I can only retrieve the last album. But the way I'm proposing it, you can do the following:
```
self.opsdroid.memory.query_data = {"_id": "2"}
self.opsdroid.memory.get("albums") # will return "Homework" album
```
or
```
self.opsdroid.memory.query_data = {"artist": "Pink Floyd"}
self.opsdroid.memory.get("albums") # will return "Dark Side of the Moon" album
```
If you don't set that variable, the current functionality is retained/maintained. The query is only affected if you set `query_data` on the object.

Hopefully that makes sense. Sorry for such a long explanation.

jacobtomlinson:
Thanks so much for the explanation. It really helps to see what you are trying to achieve here. 

Let me add a little Opsdroid context. 

The opsdroid memory module is intended to be a very simple key value store for users to store data. There are multiple backends which can persist that data, but the idea is for them to be totally abstracted. It shouldn't matter which backend you choose, they all work the same way. 

Specifically the opsdroid memory module is not intended as a way to access existing data in a database. 

All of the memory modules make the configured instance of the underlying library available as s property if you want to dig under the hood in your skills. But once you go down that road you're on your own. 

However the overall preference would be that if you feel constrained by the opsdroid memory module then you need to handle database storage yourself within your skill. 

Therefore I'm hesitant to expand the memory API to include something like this.

jacobtomlinson:
For example from your skill `self.opsdroid.get_database("mongo").database` is an instance of [AsyncIOMotorDatabase](https://motor.readthedocs.io/en/stable/api-asyncio/asyncio_motor_database.html#motor.motor_asyncio.AsyncIOMotorDatabase) and you can use it to make queries yourself.

That gives us a clean separation in opsdroid where you can do advanced things without us having to explicitly support it.

ajitdsa:
Okay, I'll go down this route and see where I land opinion-wise. I appreciate the feedback/discussion. Do you mind if we keep this PR open in some sort of "do not merge" state until I can give you feedback on this?

.gitignore
ajitdsa: Oh is there a non-global .gitignore?!

jacobtomlinson: The `.gitignore` is per repo. You can also create a global file which applies to all projects on your computer. Follow the link I shared and check it out.

Anything which is a personal preference like your editor (`.vscode`, `.idea`, etc), environment manager, etc should probably go in your global gitignore file.

docs/databases/mongo.md
jacobtomlinson: I don't think this should be a dict. It should just be the value you want to store `remember`.

.gitignore
ajitdsa: Oh nice, I didn't even realize that was a thing. But it totally makes sense. (I was reading on my phone before and didn't even notice the link.)

ajitdsa:
After giving it some more thought, I do think that providing functionality like this is a correct path for mongodb. I don't feel like this is a special use case, but rather the most common way document-based databases are used.

Thinking of it a slightly different way, if you consider a document database like a file system that has folders and documents, even the simplest use case would be to ignore the folders and just create all the documents in a single folder and then look up the documents by some id/information of the document itself. The unfortunate thing is that mongodb's convention doesn't allow you to just create documents in the "root folder" equivalent, you kinda have to create folders.

The use case that is currently supported is the equivalent of a folder being the key and then generally storing one document per folder. I'm just advocating for the case where we support creating one folder and then storing/retrieving documents from that one folder. I'm still only advocating simple key/value lookups and not some advanced querying option.

I do see a problem with my implementation though. I made a change to the `Memory` object and then had that pass the `query_data` to the database in the handoff functions. But since mongodb is the only case where this applies, I think I should remove it from there and make sure that I can still do everything I need by manipulating the specific database object instance from my skill.

Let me know your thoughts. And as always, thanks. :)

jacobtomlinson:
I've marked this as a draft (the GitHub way of doing "do not merge").

_Responding to some messages sent to me outside of GitHub by @ajitdsa_

> I do think you should either reconsider merging my PR, or work with me to make adjustments to it such that it is more palatable to you

Absolutely! I would love to continue discussing this PR and find something that works for us.

It would also be helpful to understand what you are trying to achieve here. Do you have a specific use case in mind? Or are you simply trying to make the mongo integration more standardised?

> This is literally the most common way nosql databases are used. And we would just be supporting simple retrieval of a key:value pair, but the key would just be a key within a document, not an entire collection.

I am happy to defer to your knowledge of nosql and mongo, I am not an expert by any stretch.

The primary goal here is for users to be able to store simple objects (strings, ints, bools, lists, dicts, anything JSON serlaisable really) with `memory.put(key)`, `memory.get(key)` and `memory.delete(key)`. They can back that storage with whatever they like, MySQL, Mongodb, a CSV file, I really don't mind and the user shouldn't have to care either beyond setting it up.

I don't want the API to grow beyond that, so my primary objection here is to adding `query_data` and insisting that users store something like `{"value": "foo"}` because the whole dict seems redundant.

If the current integration is not using mongo correctly and we should change to storing everything in one document instead of doing one document per key then I'm totally happy with that. We can change the underlying implementation to be the best that it can be. I just don't want to change what the user sees.

>  I really am enjoying opsdroid, and working on opsdroid. Thanks for all the work you've put into it

We are really happy to have you here! Thanks for the help.

---

The other thing I want to reiterate is that when we open up a database in the plugin we generally authenticate with the database and create some kind of object which is used to connect with the database. 

This object should also be accessible to the user so that they can just grab it and interact with the database natively themselves. We could definitely improve naming consistency and documentation around this.

They would need to leave the `opsdroid` table alone because they could break things. But if they want to do more advanced operations they can create their own tables and do whatever they like. My hope is that they will benefit from not having to authenticate or store credentials in other places.

Opsdroid is designed as a framework to help you build certain things. But it is just a design pattern and library that you use within your code, so you are welcome to break free of the guide rails we provide and write whatever code you want, we just won't provide support for that.

jacobtomlinson:
> I don't feel like this is a special use case, but rather the most common way document-based databases are used.

To respond to your comment here too, I am very happy to change the implementation of the mongo module. If we are misusing mongo then we should fix that, and your guidance is much appreciated here.

My only concern is changing the user facing API.

ajitdsa:
Okay! This actually makes sense, and I appreciate that feedback. Yes, in my opinion the project is... "misusing" is a strong word... but yes, misusing mongo. So now that you provided me with some more very helpful context, this is the better answer:

When setting up the mongodb, it would be better for the user to specify a database and a collection. That way the user is just specifying "a single location (or folder) to store serialized json data", and then the "key" that they are passing in for get/put/delete would just be stored in the `_id` column and the `data` passed in via the `put` could just be stored within a `data` key. This would make for the most consistent storage/retrieval mechanism possible, in my opinion.

ajitdsa:
☝️ I'd be happy to start over and write a PR that does it this way instead.

jacobtomlinson:
> This would make for the most consistent storage/retrieval mechanism possible, in my opinion.

Sounds great!

ajitdsa:
Going to close this in favor of a completely different approach.

PR #689, opened 2018-10-03
[WIP] Move MongoDB database module into core #669

limberger:
# Description

Move module MongoDB database into core

Fixes #669

## Status

**UNDER DEVELOPMENT**

## Type of change

- New feature (non-breaking change which adds functionality)
- Documentation (fix or adds documentation)

# How Has This Been Tested?

_Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration._

- Test A
- Test B

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

# Steps

    Make a new submodule directory in opsdroid.database and copy the database code over. [OK]
    Update the requirements.txt with any dependencies from the database if necessary. [OK]
    Write tests for the database. (See the Slack connector tests for inspiration). [WIP]
    Copy the relevant information from the database README.md into a new documentation page. [OK]
    Add the new page to the mkdocs.yml. [OK]
    Add to the list of databases. 
    Add a deprecation notice to the old database module. (See the slack connector)

PR #1799, opened 2021-07-17
Add collection to opsdroid mongo database module

ajitdsa:
# Description

The current implementation of the mongodb module creates multiple mongodb collections and stores one document per collection. This proposed change creates one collection and writes multiple documents to that collection and stores/retrieves them using an `_id` field as specified by a `key`.

- Write all mongo documents to a single collection and retrieve using _id.
- Fix documentation for opsdroid memory skill example.
- Adjust tests to account for collection.

Fixes https://github.com/opsdroid/opsdroid/issues/1796

## Status
**UNDER DEVELOPMENT**

## Type of change

# How Has This Been Tested?

I adjusted the existing tests in `opsdroid/database/mongo/tests/test_mongo.py`, but my Python skills aren't that great, so there is still one broken test and I'm not sure how to fix it. Any help would be appreciated.

# Checklist:


docs/skills/memory.md
```diff
@@ -29,17 +29,17 @@ class RememberSkill(Skill):
...
+        information = await self.opsdroid.memory.get("remember_this")
         await message.respond(information)
         
     @match_regex(r'forget it')
-    async def remember(self, message):
-        await self.opsdroid.memory.delete("remember")
+    async def forget_it(self, message):
+        await self.opsdroid.memory.delete("remember_this")
         await message.respond("Ok I'll forget it")
 ```
```
ajitdsa, line 44: Made the method names in this example unique, and also changed the key name to make it different from the method name for slightly improved clarity.

opsdroid/database/mongo/__init__.py
```diff
@@ -64,10 +65,18 @@ async def put(self, key, data):
...
         """
         _LOGGER.debug("Putting %s into MongoDB.", key)
-        if "_id" in data:
-            await self.database[key].update_one({"_id": data["_id"]}, {"$set": data})
-        else:
-            await self.database[key].insert_one(data)
+        if isinstance(data, str):
+            data = {"value": data}
+        if "_id" not in data:
+            data["_id"] = key
```
ajitdsa, line 71: Did a little data manipulation for simplicity and consistency prior to writing to db. This will make the API as simple and correct as possible.

opsdroid/database/mongo/__init__.py
```diff
@@ -77,9 +86,12 @@ async def get(self, key):
 
         """
         _LOGGER.debug("Getting %s from MongoDB.", key)
-        return await self.database[key].find_one(
-            {"$query": {}, "$orderby": {"$natural": -1}}
+        response = await self.database[self.collection].find_one(
+            {"$query": {"_id": key}, "$orderby": {"$natural": -1}}
         )
+        if response.keys() == {"_id", "value"}:
+            response = response["value"]
```
ajitdsa, line 93: Did this data manipulation on retrieval in order to make the [memory example](https://docs.opsdroid.dev/en/latest/skills/memory.html#example) work without manipulation/adjustment.

opsdroid/database/mongo/tests/test_mongo.py
```diff
@@ -35,24 +43,22 @@ async def test_connect(database):
...
-    database.database["test_db"] = DatabaseMongoTest({})
+    database.database = {"test_collection": DatabaseMongoTest({})}
     with pytest.raises(TypeError):
         await database.get("test_db")
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize("config", [{}])
+@pytest.mark.parametrize("config", [{"collection": "test_collection"}])
 async def test_put(database):
```
ajitdsa, line 55: To be honest, I'm not sure what I'm doing wrong on this test. Any help would be appreciated.

ajitdsa (commented):
Some comments to help the reviewer.

cognifloyd (commented):
This is a much better way to use mongo. Cool!

Frankly, I find these tests a bit confusing too. lol

For the tests, we'll probably need to adjust the `database` fixture (top of the `test_mongo.py`) and `opsdroid/database/mockmodules/mongo/mongo_database.py` (this mocks an `AsyncIOMotorClient`).

I would probably rename `DatabaseMongoTest` to `DatabaseMongoCollectionMock` here:
https://github.com/opsdroid/opsdroid/blob/730b460235fc217ef5d5819742ca3ce6e67c2e5f/opsdroid/database/mockmodules/mongo/mongo_database.py#L4
And instead of `self.dummy_db`, I think it should have `self.dummy_doc`:
https://github.com/opsdroid/opsdroid/blob/730b460235fc217ef5d5819742ca3ce6e67c2e5f/opsdroid/database/mockmodules/mongo/mongo_database.py#L10

Now, I don't understand why the mock methods are doing this (why would anyone `await` a dictionary? `await {}` makes no sense:
https://github.com/opsdroid/opsdroid/blob/730b460235fc217ef5d5819742ca3ce6e67c2e5f/opsdroid/database/mockmodules/mongo/mongo_database.py#L17

I would also move creating `DatabaseMongoTest()` (or `DatabaseMongoCollectionMock()` if we rename it) out of the test functions. I would add another fixture like this:
```python
@pytest.fixture()
def mocked_database(database):
    config = database.config
    collection = config.get("collection", "test_collection")
    database.database = {collection: DatabaseMongoCollectionMock({})}
    return database
```

As far as the tests go, I'm not sure why there's all of the `TypeError` and other error checking. I really don't see how the `test_get`, `test_put`, `test_put2`, and `test_delete` tests are doing much of anything useful.

cognifloyd:
@limberger , you added the tests for the mongodb bits in #689 - can you help me understand what the tests are doing? Why is the mock doing `return await self.dummy_db` (essentially `return await {}`) and then why are the tests are checking for raised `TypeError`? Can you help me understand the thinking behind building the tests this way?

cognifloyd:
Hmm. An older version of the Mock class is a bit closer to what I would create.
https://github.com/opsdroid/opsdroid/pull/689/commits/edffbb0029b7852b9688678e87ee709d9938c3c6

opsdroid/database/mongo/__init__.py
```diff
@@ -64,10 +65,18 @@ async def put(self, key, data):
...
-        else:
-            await self.database[key].insert_one(data)
+        if isinstance(data, str):
+            data = {"value": data}
+        if "_id" not in data:
+            data["_id"] = key
+
+        response = await self.database[self.collection].update_one(
+            {"_id": data["_id"]}, {"$set": data}
+        )
```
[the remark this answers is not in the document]
ajitdsa: Oh that’s a good call!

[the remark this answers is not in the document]
ajitdsa: Also `dummy_doc` makes much more sense 👍

jacobtomlinson (commented):
I don't have much to add over what @cognifloyd has said. But this is looking great overall.

ajitdsa:
Awesome, well I think if I can get some rest help either from @limberger or someone else, we can put a pin in this. I have availability next week if anyone would want to pair on it.

opsdroid/database/mongo/__init__.py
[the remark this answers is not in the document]
ajitdsa: I made the `upsert` addition, although the syntax was slightly different 👍

ajitdsa:
Okay, I made some improvements to this. I made it so that we can set the `collection` via a `contextmanager` now like this:
```python
        remember = "I will be remembered!"
        db = self.opsdroid.get_database("mongo")
        with db.memory_in_collection("mycollection"):
            await self.opsdroid.memory.put("remember_this", remember)
```
obviously, if you don't use the context manager, the behavior will just be the default.

Just want to make sure that this is what we are going for. Looks like I still need to write a test or tests using the contextmanager, and then there is still a broken test that I don't understand why it's broken ☹️

opsdroid/database/mongo/__init__.py
```diff
@@ -100,5 +100,14 @@ async def delete(self, key):
...
 
         """
-        _LOGGER.debug("Deleting %s from MongoDB.", key)
-        return await self.database[self.collection].delete_one({"_id": key})
+        _LOGGER.debug("Deleting %s from MongoDB collection %s.", key, self.collection)
+
+        return await self.database[self.collection].delete_one({"key": key})
+
+    @contextmanager
+    def memory_in_collection(self, collection):
```
cognifloyd, line 108: If two different async tasks retrieve the db, and they each use a different collection, will they interfere with each other?
What if this returned a mutated copy of the db object instead of changing the setting like this?

opsdroid/database/mongo/__init__.py
```diff
@@ -13,6 +14,7 @@
     "database": str,
     "user": str,
     "password": str,
+    "collection": str,
```
cognifloyd, line 17: isn't this: `default_collection_name` or `default_table_name`? Or is that somewhere else?

docs/databases/mongo.md
```diff
@@ -4,19 +4,30 @@ A database module for [opsdroid](https://github.com/opsdroid/opsdroid) to persis
...
   mongo:
-    host:       "my_host"     # (optional) default "localhost"
-    port:       "12345"       # (optional) default "27017"
-    database:   "my_database" # (optional) default "opsdroid"
-    user:       "my_user"     # (optional)
-    password:   "pwd123!"     # (optional)
+    host:                       "my_host"       # (optional) default "localhost"
+    port:                       "12345"         # (optional) default "27017"
+    database:                   "my_database"   # (optional) default "opsdroid"
+    default_collection_name:    "my_collection" # (optional) default "opsdroid"
```
cognifloyd, line 17: Have you adjusted the code to use `default_collection_name` instead of `collection`? It looks like it is still looking for `collection`.

cognifloyd (commented):
Maybe we could make the `contextmanager` be something like this:
```
        remember = "I will be remembered!"
        db = self.opsdroid.get_database("mongo")
        with db.memory_in_collection("mycollection") as memory:
            await memory.put("remember_this", remember)
```
Which is equivalent to:
```
        remember = "I will be remembered!"
        await self.opsdroid.memory.put("remember_this", remember, collection="mycollection")
```
This way it is clear that `self.opsdroid.memory.*` interacts with the default collection unless you explicitly change it. And the context manager is handy for setting that collection var for several memory calls:
```
        with db.memory_in_collection("mycollection") as memory:
            last_remembered = await memory.get("remember_this")
            await memory.put("last_remembered", last_remembered)
            await memory.put("remember_this", remember)
```

docs/databases/mongo.md
ajitdsa: Oh whoops, didn't change the docs. So `default_collection_name` was originally proposed by me, but I reconsidered once we decided to go the `contextmanager` route. I don't think it's necessary to be so verbose anymore. I'll change these docs to reflect.

opsdroid/database/mongo/__init__.py
ajitdsa: Good catch. The `default_table_name` was proposed in a postgres related PR, but whatever we choose we should be consistent.

opsdroid/database/mongo/__init__.py
ajitdsa: That actually sounds like a great idea to me.

ajitdsa:
@cognifloyd in regards to your main comment about how we might better implement the contextmanager, I will say that I was just trying to stay consistent with the matrix database implementation here: https://github.com/opsdroid/opsdroid/blob/master/opsdroid/database/matrix/__init__.py#L288-L294. Would you recommend changing that one to be consistent? Or is there a justifiable reason for the inconsistency?

opsdroid/database/mongo/__init__.py
ajitdsa: Hey @cognifloyd, so I made a modification to the `contextmanager` as requested. It works great, but needed to be changed to an `asynccontextmanager` as a result. I changed the docs to reflect usage, but it's basically as you suggested above. Thanks so much for the help. I'll see if I can make heads-or-tails of the tests, as they currently make no sense to me. Cheers.

cognifloyd: Nice that looks awesome. Maybe try commenting out the tests and slowly reimplements each one so that it makes sense?

ajitdsa:
@cognifloyd I have gone through and fixed the old broken tests, but I may need some help writing a couple new ones to cover these lines of code: https://github.com/opsdroid/opsdroid/pull/1799/checks?check_run_id=3576627681 ... if you're able to help, please feel free, otherwise feel free to suggest some things to try and I'll see what I can do. Cheers.

opsdroid/database/mongo/tests/test_mongo.py
```diff
@@ -35,49 +52,40 @@ async def test_connect(database):
...
 @pytest.mark.asyncio
-@pytest.mark.parametrize("config", [{}])
-async def test_get(database):
-    database.database = {}
-    database.database["test_db"] = DatabaseMongoTest({})
-    with pytest.raises(TypeError):
-        await database.get("test_db")
+@pytest.mark.parametrize("config", [{"collection": "test_collection"}])
+async def test_get(mocked_database):
+    mocked_database.database = {"test_collection": DatabaseMongoCollectionMock({})}
```
cognifloyd, line 57: You don't need this line because it is handled in the `mocked_database` fixture.
```suggestion
```

opsdroid/database/mongo/tests/test_mongo.py
```diff
@@ -35,49 +52,40 @@ async def test_connect(database):
...
-            await database.put("test_db", {})
-        except NotImplementedError:
-            raise Exception
-        else:
-            pass
-    else:
-        raise Exception
+@pytest.mark.parametrize("config", [{"collection": "test_collection"}])
+async def test_get2(database):
+    database.database = {"test_collection": DatabaseMongoCollectionMock({})}
```
cognifloyd, line 64:
```suggestion
```

opsdroid/database/mongo/tests/test_mongo.py
```diff
@@ -35,49 +52,40 @@ async def test_connect(database):
...
-async def test_put2(database):
-    try:
-        await database.put("test_db", {})
-    except TypeError:
-        pass
-    else:
-        raise Exception
+@pytest.mark.parametrize("config", [{"collection": "test_collection"}])
+async def test_put(mocked_database):
+    mocked_database.database = {"test_collection": DatabaseMongoCollectionMock({})}
```
cognifloyd, line 72:
```suggestion
```

cognifloyd (commented):
Cool. That makes the tests so much clearer!
I left comments about a couple unnecessary lines.
Now we need to figure out what assertions to make to validate that the code works as designed.
I don't have time to do anything atm, but I'll think about what that might be.

opsdroid/database/mongo/tests/test_mongo.py
ajitdsa: You right, you right. Updated.
Ok, my friend @eheiberg helped me create a test for the context manager. We think we're at a pretty good place.

ajitdsa:
Okay! I think I actually learned enough that I'm happy with these tests and all code is covered! If you guys are okay with it, from my standpoint this is ready to be merged. :) If we think more changes are needed, let me know. Either way, thanks so much for your help, @cognifloyd

opsdroid/database/mongo/tests/test_mongo.py
ajitdsa: The more I look at it, I'm not sure there is anything to assert in the general case. Since we are mocking the mongo database/collection, I think we are mostly looking to make sure nobody adds any code that would generate an exception with the test as defined. Let me know if you see something different.

jacobtomlinson:
Thanks for all the effort here @ajitdsa. And thanks for the review @cognifloyd.

PR #1808, opened 2021-08-03
Fix GitHub event creation

jacobtomlinson:
When parsing payloads from the GitHub webhook there are a few series of if statements. In the case of a PR Merged event the if was passing and creating the event, but then the next if was also passing and replacing it with a PR Closed event.

This PR updates it to use `elif` instead which fixes this issue.

I've also expanded the test to ensure that the skill gets called, cc @ajitdsa.

### DIFF
--- a/opsdroid/connector/github/__init__.py
+++ b/opsdroid/connector/github/__init__.py
@@ -113,7 +113,7 @@ class ConnectorGitHub(Connector):
                 sender=user,
             )
 
-        if payload["action"] == "completed":
+        elif payload["action"] == "completed":
             if payload["check_run"]["conclusion"] == "success":
                 event = github_events.CheckPassed(
                     action=payload["action"],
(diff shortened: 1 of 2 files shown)


opsdroid/connector/github/__init__.py
```diff
@@ -113,7 +113,7 @@ async def handle_check_event(self, payload: dict, user: str) -> github_events:
                 sender=user,
             )
 
-        if payload["action"] == "completed":
+        elif payload["action"] == "completed":
```
ajitdsa, line 116: Good catch. I did think it was weird that they were all indepedent `if` statements but didn't think about that might cause problems.

ajitdsa (approved):
Approved!

jacobtomlinson:
Thanks for the review @ajitdsa

Issue #1786, opened 2021-06-27
Instructions for Slack connector in `Socket Mode` may need adjusting


ajitdsa:
# Description
When I went to create a new opsdroid and connect it to Slack in `Socket Mode`, I could see the app receiving messages, but couldn't get opsdroid to respond in the channel. During the course of trial and error I did get it to work in `Events API` mode, and then when I switched back to `Socket Mode`, that started working. That's when I realized that `Events API` is doing a URL verification step with a challenge parameter (see [here](https://api.slack.com/apis/connections/events-api#subscriptions) in the `URL verification handshake` section) and that needs to be set up before you can switch over to socket mode. I've since reproduced this issue in order to make sure it was correct. This tripped me up for quite a while, so I thought I would report it so that adjustments could be made to help future droiders.

## Steps to Reproduce

1. Create and setup a new Slack app.
2. Setup opsdroid using the Slack connector and add the `hello` skill.
3. Follow the `Socket Mode` instructions [here](https://docs.opsdroid.dev/en/stable/connectors/slack.html#choose-the-backend-api) in order to enable socket mode.
4. Trigger the `hello` skill by going into the channel and typing `hello`.

There is no response in the channel.

## Expected Functionality
There should be a response from opsdroid in the channel. Here's what you need to do in order to get things to work properly:

1. After following the above steps, go back and change the settings to use `Events API`, including setting `socket-mode` to `false` in opsdroid and restarting. (There is a typo in these instructions: the word `toggle` is spelled `toogle`.)
2. Give it a quick test to make sure you set everything up correctly. You should see a response to the `hello` skill in the channel.
3. Now go back to Slack and turn `Socket Mode` back on for your app.
4. Disconnect your tunnel.
5. Change the `socket-mode` setting in opsdroid back to `true`.

Now when you go back to Slack and trigger the hello skill again, you will get a response from opsdroid using Socket Mode.

## Experienced Functionality
So the lesson here is that using the Socket Mode instructions from the Slack connector page, I was not able to establish a proper Socket Mode connection. However, if I set up Events API first, and then changed it to Socket Mode, I was able to establish a proper Socket Mode connection.

## Versions
- **Opsdroid version:** 0.22.0
- **Python version:** 3.9.5
- **OS/Docker version:** 20.10.5, build 55c4c88

## Configuration File
Please include your version of the configuration file below.

```yaml
logging:
  level: debug
  path: opsdroid.log
  console: true
  timestamp: true
welcome-message: true
[...]

PR #1823, opened 2021-08-15
Adjust Slack instructions to account for necessary verification step

ajitdsa:
# Description

The instructions for setting up Slack in Socket Mode actually require the user to set up the Events API first. This is due to an inconsistency in Slack and doesn't appear to be anything we can fix via code. Therefore, the next best solution was to adjust the instructions for the connector. I went back and forth on how to do this, and ended up here.

Fixes https://github.com/opsdroid/opsdroid/issues/1786

## Status
**READY**

## Type of change

# How Has This Been Tested?

n/a

# Checklist:


jacobtomlinson:
Boo to Slack, but thanks for making this change here.

PR #1834, opened 2021-09-30
Log errors instead of printing them

jacobtomlinson:
Looks like some errors were being printed instead of logged. This is causing them to be truncated.

### DIFF
--- a/opsdroid/core.py
+++ b/opsdroid/core.py
@@ -142,14 +142,14 @@ class OpsDroid:
             context (String): Describes the exception encountered.
 
         """
-        print("ERROR: Unhandled exception in opsdroid, exiting...")
+        _LOGGER.error("Unhandled exception in opsdroid")
         if "future" in context:
             try:  # pragma: nocover
                 context["future"].result()
             # pylint: disable=broad-except
             except Exception:  # pragma: nocover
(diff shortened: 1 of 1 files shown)


Cadair (changes requested):
Wait no.

I made these print on purpose. During the teardown of the interpreter the logger can vanish leading to these lines throwing even more errors.

Let me see if I can dredge up that change.

Cadair:
https://github.com/opsdroid/opsdroid/pull/1069/files#r327158672

jacobtomlinson:
Hrm that's a shame. It looks like the print statements here are truncating the errors. Do you have any thoughts on how we can fix that?

Cadair:
Do you mean that it's not printing the whole traceback?

jacobtomlinson:
Yeah

xloem:
I opened #1983 before noticing this.

It looks like it would make sense to wrap logging of the exceptions in `sys.is_finalizing` https://github.com/python/cpython/issues/70976#issuecomment-1093712202 note also python 3.10 is much more stable for this case.

I'd think ideally these exceptions would be passed to user code.

jacobtomlinson:
Closing as super stale

PR #1841, opened 2021-10-07
Bugfix KeyError when sending data into a new room before a sync happened

### LINKED ISSUE #1838: Interacting with a new or newly joined room doesn't work
# Description
If in a single event triggered by opsdroid a skill creates or joins a room. Then attempts to interact with it for example send a message in that "new" room this error occurs `nio.exceptions.LocalProtocolError: No such room with id !roomID:server.com found.`

It doesn't matter how long you wait between creating/joining and sending the message or other event the same error will occur.

The only "workaround" for this is to create/join a room and use a different matcher to trigger a separate event.

## Steps to Reproduce
put this into a skill configured to use the matrix connector and run.
```python
  @match_regex(r'!run')
  async def run(self, opsdroid, config, message):
    test = await self.opsdroid.send(NewRoom(name='name'))
    await self.opsdroid.send(Message('string', target=test))
```

### LINKED ISSUE #1839: Add support to the matrix client for waiting for a sync to occur
There are certain operations, such as joining a new room (or creating one) where nio requires a sync call to return to refresh its internal state.

For opsdroid this means we must call sync and then process it's return (and trigger any events) before we return from any coroutines which join a new room, otherwise we end up with issues like #1838 where things behave in weird ways.

Without looking at the code, we will need to factor out the code from the `listen` method into a separate coroutine which can be called from the relevant methods when a state refresh is needed.

see https://github.com/poljar/matrix-nio/issues/279 for more info.

Sleuth56:
# Description
A key error would occur when sending a message or other state into a newly joined or created room. 

What's happening is when encryption is enabled, matrix-nio holds a local list of all rooms. It uses this list to make sure that all devices have it's sender key before sending encrypted messages. Opsdroid doesn't tell nio to resync with the server and refresh the internal list after a call is made that would alter that list. This addresses that by adding a call to force a sync to happen and adding that to functions that alter the available rooms list. Thus resolving the KeyError.

Fixes #1838
Fixes #1839

## Status
**NEEDS REVIEW**

## Type of change

# How Has This Been Tested?
This is a duplicate of my issue above. More context can be found there. #1838
```python
  @match_regex(r'!run')
  async def run(self, opsdroid, config, message):
    test = await self.opsdroid.send(NewRoom(name='name'))
    await self.opsdroid.send(Message('string', target=test))
```

Running locally on Linux against my own skills and Synapse instance.

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -324,27 +324,30 @@ class ConnectorMatrix(Connector):
                             event.source, roomid
                         )
 
-    async def listen(self):  # pragma: no cover
-        """Listen for new messages from the chat service."""
-        while True:  # pylint: disable=R1702
-            response = await self.connection.sync(
+    async def sync(self):
+        response = await self.connection.sync(
                 timeout=int(60 * 1e3),  # 1m in ms
                 sync_filter=self.filter_id,
                 since=self.connection.sync_token,
             )
-            if isinstance(response, nio.SyncError):
-                _LOGGER.error(
-                    f"Error during sync: {response.message} (status code {response.status_code})"
-                )
-                continue
+        if isinstance(response, nio.SyncError):
+            _LOGGER.error(
+                f"Error during sync: {response.message} (status code {response.status_code})"
+            )
+            return
 
-            _LOGGER.debug(_("Matrix sync request returned."))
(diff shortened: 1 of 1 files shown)


Sleuth56:
@Cadair I'm very confused why adding sync calls would fail tests. I don't understand pytest at all though.

PR #1845, opened 2021-10-22
Get connector name from configuration,

### LINKED ISSUE #1844: Allow users to specify connector name
I wanted to use two slack connectors, but in a lot of connectors, we are hardcoding the name in the `__init__` method, if we need to create a path in the API for webhooks, we can only use one name and one connector since both would be called slack and when we turn our configuration into a dict, the two connectors get merged into one.

Ideally, we should allow users to rename the connectors to whatever they want within the configuration.

FabioRosado:
# Description

Fixes #1844

This PR replaces all the hardcoded names with `self.config.get("name", <connector name>)` to allow users to specify the name of the connector. It also adds a section in our connectors page telling users how to use two connectors in the same opsdroid instance since I had to search for a while about how to do this until I remembered that Jacob had mentioned something like that in one of our channels 😆 

I'm also using the import organize imports setting in VSCode so some imports were automatically changed. Let me know if you would rather me revert these changes.

## Status

## Type of change

# How Has This Been Tested?

- Created two slack connectors, both worked

# Checklist:

### DIFF
--- a/opsdroid/connector/facebook/__init__.py
+++ b/opsdroid/connector/facebook/__init__.py
@@ -3,9 +3,11 @@ import json
 import logging
 
 import aiohttp
+
+from voluptuous import Required
+
 from opsdroid.connector import Connector, register_event
 from opsdroid.events import Message
-from voluptuous import Required
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


jacobtomlinson:
We use black and isort in Dask. Seem to play nicely together these days.

PR #1849, opened 2021-10-24
Update example config and slack docs

FabioRosado:
# Description

While working on a skill for slack I've noticed that our example configuration was out of date and I had to look into the docs to figure out what was missing from the configuration. I've also noticed a small inconsistency on getting slack to work since I had to set `socket-mode` to false before I was able to start anything since slack was failing with an unhelpful message (it only mentioned that the configuration had changed).

This PR updates the example configuration so new users will have the right configuration params and moves the `Event mode` first in the slack docs and adds a note saying: HEY do this first!

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:

### DIFF
--- a/docs/connectors/gitter.md
+++ b/docs/connectors/gitter.md
@@ -32,5 +32,9 @@ To use the Gitter connector you will need a user for the bot to use and generate
 ```yaml
 connectors:
   gitter:
-    room-id: "to be added" #required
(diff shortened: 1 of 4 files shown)


jacobtomlinson (approved):
Thanks for this @FabioRosado

PR #1855, opened 2021-10-31
Allow skip training rasa model on startup

### LINKED ISSUE #1853: Use pre-trained Rasa model.
# Description
There is no option to take a pre-trained Rasa model (perhaps trained on a more powerful machine) and use that for predictions instead of training a new model each time opsdroid is started.

## Expected Functionality
There should be a configuration option under something such as `parsers->rasanlu->no-train` which by default is false, but can be set to true in order to prevent opsdroid sending the request to re-train each time.

## Experienced Functionality
No such option exists.

## Versions
- **Opsdroid version:**0.23.0
- **Python version:**3.9.2
- **OS/Docker version:**Debian 11

FabioRosado:
# Description

Fixes #1853

This issue seemed like an easy win so I added a flag to allow setting the training on/off for the Rasa NLU parser. 

## Status

## Type of change

# How Has This Been Tested?

- Added breakpoint to confirm that we skip the training on startup
- Tox all green

# Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -366,6 +366,7 @@ parsers:
     project: opsdroid
     token: 85769fjoso084jd
     min-score: 0.8
+    train: True
 ```
 
 Some parsers will allow you to specify a min-score to tell opsdroid to ignore any matches which score less than a given number between 0 and 1. You just need to add the required min-score under a parser in the configuration.yaml file.
--- a/docs/skills/matchers/rasanlu.md
+++ b/docs/skills/matchers/rasanlu.md
@@ -10,6 +10,9 @@ You can also set a `min-score` option to tell opsdroid to ignore any matches whi
 
 `models-path` is used to load the model in Rasa. It defaults to `models`. If your Rasa instance writes trained models somewhere else, you can set this option to whatever path you need.
 
+`train` is used to make opsdroid train your model on each start. Setting this configuration flag to `False` allows you to use a previously trained
+model.
+
 ```yaml
 
 parsers:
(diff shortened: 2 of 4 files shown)


Cadair:
Seems fine @oleg-fiksel rewrote the rasa parser iirc, so wonder if they have any comments.

jacobtomlinson (approved):
This looks great thanks!

oleg-fiksel:
I think we should make the check (load or not to load the model) not in core.py but in rasanlu.py (func `train_rasanlu`).
Because if we completely ignore `train_rasanlu` we ignore the `_check_rasanlu_compatibility` and therefore will get an error if the Rasa API isn't compatible with our implementation.

We should probably move the call of `_check_rasanlu_compatibility` some place before ether way because if no intens are found (https://github.com/FabioRosado/opsdroid/blob/fr/rasa/opsdroid/parsers/rasanlu.py#L157) the compatibility check is also not called.

jacobtomlinson:
Sounds good @oleg-fiksel. Do you feel like raising a follow up PR to this one?

oleg-fiksel:
Yes. I plan to work on Rasa Matcher on Christmas holidays. Can this wait until then?

jacobtomlinson:
Sure

bparbhu:
So in this fix was this done with the use of a container that was spun up or a rasa server which already trained a model and the config options just point you to the file path and model name? The reason I'm asking is I'm trying to do the same in Docker-compose and I am spinning up my rasa instance at the same time. So would this have worked in a situation where there was already a rasa instance with a trained model?

oleg-fiksel:
> I think we should make the check (load or not to load the model) not in core.py but in rasanlu.py (func `train_rasanlu`). Because if we completely ignore `train_rasanlu` we ignore the `_check_rasanlu_compatibility` and therefore will get an error if the Rasa API isn't compatible with our implementation.
> 
> We should probably move the call of `_check_rasanlu_compatibility` some place before ether way because if no intens are found (https://github.com/FabioRosado/opsdroid/blob/fr/rasa/opsdroid/parsers/rasanlu.py#L157) the compatibility check is also not called.

Here's the promised PR: #1886 🙂 
Sorry for quite a big delay.

PR #1860, opened 2021-11-05
Add support for mongodb+srv:// urls in MongoDB backend

### LINKED ISSUE #1859: User configurable connection for mongo-based databases
So the pymongo client has a multitude of ways for connecting to different mongo services

So for MongoDB Atlas users the connection string is given as such 
for python connections to the mongo db atlas 

`mongodb+srv://<username>:<password>@<cluster-name>.mongodb.net/myFirstDatabase`

In making the mongo connection to be user configurable we can specify different types of mongo services versus
just asking for the basic connection arguments like port, user name,  pass, and also we can give users an easier way to connect versus making assumptions about the type of mongodb the kinds of credentials they might have. 

As long as the pymongo client accepts the connection and connects the user to the database and the collection they want I think this would be great!

Thanks again guys!

jacobtomlinson:
# Description

Added a new `protocol` configuration option to the MongoDB backend to support using `mongodb+srv://` URLs.

As that protocol doesn't support port numbers I also shuffled things a little so that if the default `27017` port is used it is omitted from the URL altogether. This protocol also requires the `dnspython` package so added that to the mongo extras install.

Fixes #1859

## Status
**READY** 

# Checklist:

### DIFF
--- a/docs/databases/mongo.md
+++ b/docs/databases/mongo.md
@@ -11,12 +11,13 @@ An accessible MongoDB server with the database that you provide already created.
 ```yaml
 databases:
   mongo:
-    host:                       "my_host"       # (optional) default "localhost"
-    port:                       "12345"         # (optional) default "27017"
-    database:                   "my_database"   # (optional) default "opsdroid"
-    collection:                 "my_collection" # (optional) default "opsdroid"
-    user:                       "my_user"       # (optional)
-    password:                   "pwd123!"       # (optional)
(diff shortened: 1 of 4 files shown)


bparbhu:
Awesome, thanks @jacobtomlinson , much appreciated work here!

FabioRosado (approved):
Nice!

Do we need to update any docs?

jacobtomlinson:
This PR updates `docs/databases/mongo.md` :)

PR #1875, opened 2021-12-13
Implement command center backend

### LINKED ISSUE #1872: Idea: Add a command centre for opsdroid
**This issue is currently a draft/brain dump for this idea - I will do some clean up in the future**

## Goal

Allow users to test opsdroid from the browser and enable/disable modules from a running bot.

## Problem

You need to edit your config when opsdroid isn't running to add connectors/skills. You also need to add a connector to interact with opsdroid since the default connector is the WebSocket one.

## Context

Often, when opsdroid is running, and you have many skills/connectors, it can be hard to figure out what modules are running without looking into the config, having logging set to debug mode, or running one of the `list config` CLI commands. Users also need to install a connector to interact with the bot since opsdroid comes only pre-installed with the WebSocket connector.

### Extending the opsdroid API

We need to extend the existing API to allow us to list all modules by type (connectors, parsers, skills) and update modules to toggle them on/off. We could as well allow users to include modules.

I have a [POC branch](https://github.com/FabioRosado/opsdroid/tree/fr/command-center) where I am hacking some of the API. It does mostly:

-  Add get route to list connectors, skills, parsers, config (
  -  sensitive configuration params like tokens are removed from the config that we return
- Add patch  route to update a module
- Add a new payload dataclass with a bunch of validation so we can handle user input better
- Add `command-center` parameter to the `web` section to enable the new API endpoints, and in the future, the UI
  - Added a token parameter under `command-center` that will be used for a poor-man authorization
    - In the future, I want to generate a new token on each start of opsdroid (if the user didn't provide one) and push this into an envvar so the UI can use it 

The configuration looks like this atm:

```yaml
web:
  command-center:
    enabled: True
    token: <my-token>
```

We should also create a new param under `command-center` to allow users to toggle the UI on/off as well letting them specify a port in which the UI runs.

#### Background info and current status

Jacob created an electron app ([opsdroid desktop](https://github.com/opsdroid/opsdroid-desktop) that allowed users to download the app, start opsdroid, and you could interact with opsdroid from within the electron app. A user (sorry don't recall who) suggested that we split the electron app and the web component to run opsdroid within the browser.

In response, opsdroid web was created. But we haven't touched it in a very, very, very long time, and at least on Node 16, you can't even start opsdroid web because many modules are deprecated and a lot of things fail.  I've started updating opsdroid web dependencies, but I still have issues with some things. Once the API is finished, I will dedicate more time to working on the opsdroid web.

Once opsdroid desktop is usable, we could make opsdroid desktop install opsdroid web from npm and be a wrapper for it (this was the original idea for us to split the app, but I don't think we implemented this?)

#### Suggestions

Since we already have opsdroid web created, we can use this as our command centre UI and extend it since opsdroid web is just a chat window.

TBD -> Opsdroid web mockup

Here's a rough overview of the things that I would like to add to opsdroid dev:

- Keep the chat windows always on the right
- Add a sidebar that shows status by calling the `/stats` endpoint from time to time
- Add a tab that contains the config (with sensitive parameters removed)
  - Section with modules that users can enable/disable
  - Allow users to toggle core skills even if they haven't been running
- Use an environment var to toggle the sidebar for the command-centre on/off

FabioRosado:
# Description

Fixes #1872

This is the WIP PR implementing the backend for opsdroid command center/dashboard.  The code is finished and i stared working on some tests and docs, but still need to bump coverage to 100% and finish off writing the docs for the command center page.

Opening the PR early on for feedback on my approach 😄 

## Status

## Type of change

# How Has This Been Tested?

- Tox - All green
- Ran opsdroid and tried `GET` on connectors, databases, parsers, skills endpoint
- Ran opsdroid and ran `PATCH` request to activate/deactivate modules

# Checklist:


docs/command-center.md
```diff
@@ -0,0 +1,33 @@
+# Command Center
+
+Opsdroid Command Center is the dashboard that allows you to interact with Opsdroid without the need to have any connector installed (other than the default websocket one). The dashboard allows you to enable/disable modules as well see your active modules and installed configuration from within your browser.
+
+Opsdroid Command Center is powered by the [opsdroid web](https://github.com/opsdroid/opsdroid-web) package, you need to first install the package using `npm`.
```
jacobtomlinson, line 5: We should handle this so that the web interface gets installed via Python. Maybe as part of our build process here or adding some packaging to PyPI in opsdroid web.

docs/command-center.md
```diff
@@ -0,0 +1,33 @@
...
+    dashboard-port: 8081
+    token: <Your chosen secret token>
+    excluded-keys:
+      - module
+      - path
+```
+
+Let's have a look at each of these configuration arguments.
+
+- `enabled` - Toggles Command Centter on/off, if enabled new routes will be added to Opsdroid's [RESTful api](rest-api.md)
```
jacobtomlinson, line 25: I think it would be fine to enable everything by default and maybe not even have a disable.

opsdroid/web.py
```diff
@@ -29,10 +79,34 @@ def __init__(self, opsdroid):
...
             self.web_app.router.add_get("/", self.web_index_handler)
             self.web_app.router.add_get("", self.web_index_handler)
 
+        self.excluded_keys = EXCLUDED_CONFIG_KEYS + self.command_center.get(
+            "excluded-keys", []
+        )
+        self.expected_header = self.command_center.get("token")
+        if self.command_center:
+            self.web_app.router.add_get("/connectors", self.connectors_handler)
+            self.web_app.router.add_get("/connectors/", self.connectors_handler)
```
jacobtomlinson, line 93: I wouldn't bother with the routes with the training slash. This only matters for the root route as browsers may append a slash.

jacobtomlinson (commented):
This looks like a great start! Seems like a great approach.

docs/command-center.md
FabioRosado: I removed this docs for now since I need to fix opsdroid web first, I tried running it and it seems its very broken on the latest node version

docs/command-center.md
FabioRosado: I think we should keep it disabled and only enable it if users want to. the reason for that is  because anyone with the URL would be able to deactivate/activate modules if they know opsdroid's url 🤔

opsdroid/web.py
```diff
@@ -198,6 +274,10 @@ async def web_index_handler(self, request):
         """
         return self.build_response(200, {"message": "Welcome to the opsdroid API"})
 
+    async def check_request(self, request):
+        if self.expected_header and request.headers.get("AUTHORIZATION") is None:
```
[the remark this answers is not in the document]
FabioRosado: Sounds good, you think we shouldn't allow users to use this feature without specifying an authorization token?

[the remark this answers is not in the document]
oleg-fiksel: IMO, yes. Details: https://github.com/opsdroid/opsdroid/pull/1875#discussion_r837910313

docs/rest-api.md
```diff
@@ -53,3 +53,258 @@ The response includes information on whether a skill was successfully triggered
...
+```python
+import requests
+
+ requests.patch(
+   "http://localhost:8080/skills", 
+   json={
+      "module_type": "skills",
+      "module_name": "seen",
+      "config": {
+          "enabled": False
```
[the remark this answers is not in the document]
FabioRosado: Yeah that's my bad, it needs to be in json format 😅  updated the docs thanks!

docs/rest-api.md
```diff
@@ -91,7 +91,7 @@ The GET method returns a list of all the connectors loaded in opsdroid and their
 }
 ```
 
-The PATCH method allows you to update a connector configuration, you can also turn it off/on my setting `enabled` to `False`. Note that in order for the connector to be updated, Opsdroid will disconnect the connector and load all the configuration again.
+The PATCH method allows you to update a connector configuration, you can also turn it off/on my setting `enabled` to `false`. Note that in order for the connector to be updated, Opsdroid will disconnect the connector and load all the configuration again.
```
oleg-fiksel, line 94: `my` -> `by`

oleg-fiksel (changes requested):
Seem that the command centre is still enabled then config contains:
```
web:
  command-center:
    enabled: False
    token: blah
```

opsdroid/web.py
```diff
@@ -198,6 +278,15 @@ async def web_index_handler(self, request):
         """
         return self.build_response(200, {"message": "Welcome to the opsdroid API"})
 
+    async def check_request(self, request):
+        client_token = request.headers.get("Authorization")
+        if (
+            client_token is None
+            or client_token != f"Basic {self.authorization_token}"
```
oleg-fiksel, line 285: If we want to stick to the Basic Auth standard then it needs a `username:password` base64 encoded. Otherwise I'd just remove the `Basic ` not to confuse the user, that we are using Basic Auth.

oleg-fiksel (approved):

opsdroid/web.py
```diff
@@ -100,12 +174,18 @@ def base_url(self):
...
             return self.config.get("base_url")
+        elif self.config.get("base-url"):
+            return self.config.get("base-url")
         protocol = "http" if self.get_ssl_context is None else "https"
         return f"{protocol}://{self.get_host}:{self.get_port}"
 
     async def start(self):
         """Start web servers."""
-        _LOGGER.info(_(f"Started web server on {self.base_url}"))
+        if self.command_center and not self.authorization_token:
```
oleg-fiksel, line 184: Seem that this is also triggered when the command centre is explicitly disabled and token is not set. This could confuse users.

Config to reproduce:
```:YAML
web:
  command-center:
    enabled: False
```

jacobtomlinson:
Awesome work here @FabioRosado! Let's get this in and iterate from here.

PR #1898, opened 2022-03-19
Docker: opsdroid should not be a member of root

dahrens:
There is no need that the user opsdroid is member
of the group root. It is also not an system account since
it has an uid > 999. Instead set no password.

# Description

Changes the usergroup of the opsdroid user.

## Status

**READY**

## Type of change

For people using the container this might be a breaking change depending on how they use volumes/binds for the container.
If they rely on the current group it will break.

# How Has This Been Tested?

Running tox.

# Checklist:


FabioRosado (approved):
Hello @dahrens  apologies for the delay in getting this in. I'm merging it now 😄 👍

PR #1907, opened 2022-04-30
Remove python 3.7 from Ci and add python 3.10

FabioRosado:
# Description

Remove python 3.7 from Ci and dds python 3.10, not sure if this is a great idea or what might break but hey, might be a good idea to test?

## Status

## Type of change

# How Has This Been Tested?

- Test A
- Test B

# Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,8 +20,8 @@ classifiers =
   Programming Language :: Python
   Programming Language :: Python :: 3
   Programming Language :: Python :: 3 :: Only
-  Programming Language :: Python :: 3.7
   Programming Language :: Python :: 3.8
+  Programming Language :: Python :: 3.9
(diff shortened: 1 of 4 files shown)


Cadair (approved):

setup.cfg
```diff
@@ -20,8 +20,8 @@ classifiers =
   Programming Language :: Python
   Programming Language :: Python :: 3
   Programming Language :: Python :: 3 :: Only
```
Cadair, line 22: Shouldn't we have a python version specifier in setup.cfg?

Cadair: Like this: https://github.com/sunpy/sunpy/blob/main/setup.cfg#L31

FabioRosado: Good point I'll add it here

PR #1911, opened 2022-05-05
Remove error log and replace with debug log

### LINKED ISSUE #1910: Slack Connector is logging an error which is not really an error
# Description
Every time opsdroid responds in a thread for the slack connector, the log is throwing an error in the following form:

```code
ERROR    Payload: {...} is not  implemented. Event wont be parsed
```

This is not really an error, but just a misplaced log message that should be a debug message.

The [create_event](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/slack/connector.py#L190) method from the slack connector is returning an empty event.  The [event_handler](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/slack/connector.py#L195) checks that the event is not empty and throws the error, which should not be. If the event returns empty (which is valid) the event handler should not parse the event. 

Since it's labeled as an error, every time there is a response in a thread from opsdroid, the ERROR is logged, which creates a lot of false entries in a production log.

## Steps to Reproduce

Create a skill where opsdroid replies inside a thread

## Expected Functionality
The error log should not appear as it's not really an error

## Experienced Functionality
The logger is longing an ERROR message which is not really an error

## Versions
- **0.26:**

chillipeper:
# Description

Fixes #1910 

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Tests passing 
-  locally tested in sample instance

# Checklist:

### DIFF
--- a/opsdroid/connector/slack/connector.py
+++ b/opsdroid/connector/slack/connector.py
@@ -8,19 +8,17 @@ import ssl
 
 import aiohttp
 import certifi
+import opsdroid.events
 from emoji import demojize
-
+from opsdroid.connector import Connector, register_event
+from opsdroid.connector.slack.create_events import SlackEventCreator
+from opsdroid.connector.slack.events import Blocks, EditedBlocks
 from slack_sdk.errors import SlackApiError
 from slack_sdk.socket_mode.aiohttp import SocketModeClient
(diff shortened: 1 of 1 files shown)


FabioRosado (approved):
Thank you for the fix!

PR #1913, opened 2022-05-11
fix: Remove obsolete in-tree-build are from Dockerfile

jinnko:
As of pip 21.3[^1] the in-tree-build feature has been the default.

We're now getting pip 22 and it seems in-tree-build is no longer being handled and causes builds to fail.  Removing it solves the issue.

[^1]: https://pip.pypa.io/en/stable/news/#v21-3

# Description

When building the docker container we're getting pip 22.1 and fail the build with  the following:

```
#0 61.35 option --use-feature: invalid choice: 'in-tree-build' (choose from '2020-resolver', 'fast-deps')
```

## Status

READY

## Type of change

# How Has This Been Tested?

- Try a docker build before the change and see it fail.
- Try a docker build after the change and see it pass.

# Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN apk update \
     setuptools-scm \
     wheel \
     && mkdir -p "${DEPS_DIR}" \
-    && pip download --use-feature=in-tree-build --prefer-binary -d ${DEPS_DIR} .${EXTRAS} \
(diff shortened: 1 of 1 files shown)


jacobtomlinson (approved):

jacobtomlinson:
Many thanks!

PR #1931, opened 2022-06-15
Small improvements to the websocket connector

### LINKED ISSUE #1930: Add token to websockets connector
Currently, anyone that knows opdroid url and endpoint will be able to request a socket to initialize a websocket connection.

Ideally, we should allow users to select a token in the configuration settings. When opsdroid gets a request, if the token doesn't exist then we just reject the request.

This will also a nice feature to implement along side opsdroid-web v2

FabioRosado:
# Description

Fixes #1930

This PR adds the possibility to specify a token with the websocket connector. If the token is provided, opsdroid will check the token against the request header when a user requests a new socket. If the token doesn't match, then we return a 403 response.

This PR also adds a new dataclass `WebsocketMessage` that we use when we receive a text message through the websocket connection. This allow users to pass a stringified json payload and supply extra bits of information such as `user` and `socket` which will be used when opsdroid parses the message.

The motivation behind the dataclass, was that in opsdroid-web I want to be able to pass a username and let opsdroid use that username when replying to messages such as with the hello skill.

## Status
**READY** | ~~**UNDER DEVELOPMENT**~~ |~~**ON HOLD**~~

## Type of change

# How Has This Been Tested?

- Ran existing tests - all green
- Testing manually with opsdroid-web all seem good

# Checklist:

### DIFF
--- a/opsdroid/connector/websocket/__init__.py
+++ b/opsdroid/connector/websocket/__init__.py
@@ -164,7 +164,7 @@ class ConnectorWebsocket(Connector):
         if self.authorization_token and (
             client_token is None or client_token != self.authorization_token
         ):
-            raise aiohttp.web.HTTPForbidden()
+            raise aiohttp.web.HTTPUnauthorized()
         return True
 
     async def listen(self):
--- a/opsdroid/connector/websocket/__init__.py
+++ b/opsdroid/connector/websocket/__init__.py
@@ -9,10 +9,47 @@ import aiohttp.web
 from aiohttp import WSCloseCode
 from opsdroid.connector import Connector, register_event
 from opsdroid.events import Message
+import dataclasses
+from typing import Optional
 
 _LOGGER = logging.getLogger(__name__)
 HEADERS = {"Access-Control-Allow-Origin": "*"}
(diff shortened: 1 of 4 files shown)


opsdroid/connector/websocket/__init__.py
```diff
@@ -108,6 +153,20 @@ async def websocket_handler(self, request):
...
+
+        If the token does not exist in the header, but exists in the configuration,
+        then we will simply return a Forbidden error.
+
+        """
+        client_token = request.headers.get("Authorization")
+        if self.authorization_token and (
+            client_token is None or client_token != self.authorization_token
+        ):
+            raise aiohttp.web.HTTPForbidden()
```
oleg-fiksel, line 167: In the case if the authorization failed I'd suggest to return HTTP 401 (Unauthorized) instead of 403 (Forbidden). The later means the authorization is successful but the user has no rights to access the resource.
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors

FabioRosado: That makes sense 😄  Thank you for the review updated the error with the Unauthorized one 👍

PR #1935, opened 2022-06-27
Add CLI option to print config file path

jacobtomlinson:
Sometimes I want to know where my config file is but not open it like with `opsdroid config edit`. This PR adds `opsdroid config path` which prints out the path to the config file.

```console
$ opsdroid config path
/home/jacob/.config/opsdroid/configuration.yaml
```


FabioRosado (approved):
I thought we had something like this haha

PR #1936, opened 2022-06-28
If env HTTPS_PROXY/HTTP_PROXY is configured, slack event resp now respects proxy config

rodrigo-albuquerque:
# Description

Fairly simple. I was testing opsdroid behind a proxy with HTTP_PROXY/HTTPS_PROXY configured and it was timing out.

Added trust_env=True to aiohttp.ClientSession and it solved the problem.

Reference: https://docs.aiohttp.org/en/stable/client_reference.html

### DIFF
--- a/opsdroid/cli/start.py
+++ b/opsdroid/cli/start.py
@@ -6,7 +6,6 @@ import logging
 
 import click
 from opsdroid.cli.utils import (
-    check_dependencies,
     configure_lang,
     path_option,
     welcome_message,
@@ -30,7 +29,6 @@ def start(path):
     configuration.
 
     """
-    check_dependencies()
 
     config_path = [path] if path else DEFAULT_CONFIG_LOCATIONS
     config = load_config_file(config_path)
--- a/opsdroid/cli/utils.py
+++ b/opsdroid/cli/utils.py
@@ -5,7 +5,6 @@ import gettext
 import logging
 import os
 import subprocess
-import sys
 import time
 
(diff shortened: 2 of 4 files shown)


jacobtomlinson (commented):
Thanks this seems great. Is there a way we can construct a simple test to ensure that this continues to work in the future?

rodrigo-albuquerque:
I'm not sure if I understand your question. You mean trust_env?

If so, `aiohttp.helpers` can confirm whether `trust_env` picked up the proxy or not:

```python
>>> import aiohttp
>>> helpers = aiohttp.helpers.proxies_from_env()
>>> helpers
{'https': ProxyInfo(proxy=URL('http://myproxy:8080'), proxy_auth=None), 'http': ProxyInfo(proxy=URL('http://myproxy:8080'), proxy_auth=None)}
>>>
```

if no proxy is picked up, helpers should be an empty dict {}.

jacobtomlinson:
I mean can we test that opsdroid works with a proxy? What I want to avoid is the `trust_env` accidentally getting removed in the future during a refactor and this functionality breaking again. We should add a test to avoid regressions here.

This would probably involve a small fixture that starts a proxy in a new process and temporarily configures `HTTP(S)_PROXY`. Then asserts that requests were successfully sent via the proxy.

rodrigo-albuquerque:
Oh, I think I've got, like a test_slack_connector_proxy.py right?

PR #1942, opened 2022-07-28
Add auto mode for pytest

chillipeper:
# Description

```
Starting from pytest-asyncio>=0.17, three modes are provided: auto,
strict, and legacy. Starting from pytest-asyncio>=0.19 the strict
mode is the default.
```
This is causing tests to fail when the latest version pytest-asyncio is installed

https://github.com/pytest-dev/pytest-asyncio/releases/tag/v0.19.0

Also removing the duplicate entry of dialogflow which is also making the tests fail

## Status
**READY**

# How Has This Been Tested?

Yes tests pass now

# Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -121,7 +121,6 @@ database_matrix =
 test =
   pre-commit
   coveralls>=2.0.0
-  dialogflow>=0.8.0
   astroid>=2.4.1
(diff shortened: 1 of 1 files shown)


FabioRosado (approved):
Thanks! I've been thinking about doing this for a long time haha

PR #1943, opened 2022-07-28
Fix channel count when retrieving channels from slack

chillipeper:
# Description

Currently the `_get_channels` method from the slack connector, is not updating all the channels that it has retrieved.

This is because the logic is not accounting for the first instance

This PR fixes this and updates the correct number of channels 

## Status
**READY** 

## Type of change

# How Has This Been Tested?

Yes, tests are passing, and bug was discovered on a running instance

# Checklist:

- [ x] I have performed a self-review of my own code
- [x ] I have made corresponding changes to the documentation (if applicable)
- [ x] I have added tests that prove my fix is effective or that my feature works
- [ x] New and existing unit tests pass locally with my changes

### DIFF
--- a/opsdroid/connector/slack/connector.py
+++ b/opsdroid/connector/slack/connector.py
@@ -194,30 +194,25 @@ class ConnectorSlack(Connector):
 
         return data
 
-    async def _retrieve_channels_from_slack(self, **kwargs):
-        """Retrieve channels from the slack API
-        args:
-            **kwargs any argument taken by https://api.slack.com/methods/conversations.list
-
-        returns:
-            string if a next cursor exists, otherwise None
-        """
-        channels = await self.slack_web_client.conversations_list(**kwargs)
-        self.known_channels.update({c["name"]: c for c in channels["channels"]})
-
-        return channels["response_metadata"].get("next_cursor")
-
     async def _get_channels(self):
-        """Grab all the channels from the Slack API"""
+        """Grab all the channels from the Slack API. This method runs while opsdroid
+        is running at every refresh_interval.
+        """
 
         while self.opsdroid.eventloop.is_running():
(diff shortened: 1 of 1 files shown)


opsdroid/connector/slack/connector.py
```diff
@@ -197,23 +194,30 @@ def _generate_base_data(self, event: opsdroid.events.Event) -> dict:
 
         return data
 
+    async def _retrieve_channels_from_slack(self, **kwargs):
```
jacobtomlinson, line 197: I'm curious why this is a new function, it seems very similar to `_get_channels`, even the docstrings are almost the same.

`Retrieve channels from the slack API` vs `Grab all the channels from the Slack API`

chillipeper: The 3 instructions under `_retrieve_channels_from_slack` had to be placed twice under `_get_channels`, so thought of shortening them into one method, if this doesn't makes sense, I can move all back to `_get_channels`

jacobtomlinson: I think the added indirection here makes it harder to read.

Perhaps instead of calling `_retrieve_channels_from_slack` twice you could refactor so the `conversations_list` only happens once inside the loop.

```python
cursor = None

while True:
    channels = await self.slack_web_client.conversations_list(cursor=cursor, limit=self.channel_limit)
    self.known_channels.update({c["name"]: c for c in channels["channels"]})
    cursor = channels["response_metadata"].get("next_cursor")

    if not cursor:
        break
```

chillipeper: Fair enough, I was calling it twice because I wasn't sure if `conversations_list` would take a `None` argument.  Will test it and modify accordingly

chillipeper: I have simplified `_get_channels` method

jacobtomlinson (approved):

PR #1947, opened 2022-10-09
Fix CI and potential bug in loader

FabioRosado:
# Description

CI has been broken for a while, I remember that one day it just stopped passing. Not sure if it was a change in the importlib or the PurePosixPath, but now when you try to pass an empty string to `importlib.util.find_spec(namespace)`  you get a ValueError `PurePosixPath('.') has an empty name`.

This PR adds `ValueError` to the except clause and also adds a debug log just so we have visibility into why things might have not worked.

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Tox - Green

# Checklist:

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -598,8 +598,7 @@ class TestCoreAsync(asynctest.TestCase):
         with TemporaryDirectory() as directory:
             await asyncio.gather(watch_dirs([directory]), modify_dir(directory))
 
-    # TODO: Test fails on mac only, needs investigating
-    @pytest.mark.xfail()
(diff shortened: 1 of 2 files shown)


tests/test_core.py
```diff
@@ -597,6 +598,7 @@ async def modify_dir(directory):
         with TemporaryDirectory() as directory:
             await asyncio.gather(watch_dirs([directory]), modify_dir(directory))
 
+    @pytest.mark.xfail(reason="Test seems to fail on Mac tests - should investigate")
```
FabioRosado, line 601: This test was xfailed in #1599 but we might have removed it, I couldn't find anything about this test being fixed in MacOS so I've just xfailed it again

PR #1953, opened 2022-11-07
Create mastodon connector

jacobtomlinson:
# Description

Started playing around with a Mastodon connector given all the folks heading there lately.

## Status
**READY**

## Type of change

New connector

# How Has This Been Tested?

Added tests and also tries it out [with a demo account](https://botsin.space/web/@opsdroiddev).

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1954, opened 2022-11-07
Resolve CI failures

jacobtomlinson:
CI seems to be failing in unrelated ways in #1953. Opening this PR to iterate and fix things.

- `jwt.encode` seems to return a string so removing the `.decode` call
- Removed version check because packaging shouldn't allow this to happen anyway in 99% of cases


FabioRosado (approved):
Nice thanks! Been meant to look into this for a while

PR #1957, opened 2022-11-30
First ephemeral

### LINKED ISSUE #1941: Feature Request: Support to respond with Ephemeral message for slack connector
# Description
Ability to send Ephemeral respond for slack connector. User is able to have option to send the respond as `in_channel` or `ephemeral` when sending responds, thanks!

sfrunza13:
# Description

The first ephemeral I got to send. Added another field to be configured in the configuration.yaml that will allow ephemerals to be sent instead of regular messages to the user that triggers the bot. 
Fixes #1941 

## Status
 **UNDER DEVELOPMENT** 

## Type of change

# How Has This Been Tested?

- So far I have set up an App I called mybot on slack, I used my local opsdroid exposing 8080 through ngrok.
image

# Checklist:

### DIFF
--- a/opsdroid/connector/slack/connector.py
+++ b/opsdroid/connector/slack/connector.py
@@ -427,7 +427,7 @@ class ConnectorSlack(Connector):
         data = self._generate_base_data(message)
         data["text"] = message.text
 
-        if self.response_type == "in_channel":
+        if self.response_type == "in-channel":
             return await self.slack_web_client.api_call(
                 "chat.postMessage",
                 data=data,
--- a/opsdroid/connector/slack/connector.py
+++ b/opsdroid/connector/slack/connector.py
@@ -60,6 +60,7 @@ class ConnectorSlack(Connector):
         self.name = config.get("name", "slack")
         self.bot_token = config["bot-token"]
         self.bot_name = config.get("bot-name", "opsdroid")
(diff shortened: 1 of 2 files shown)


sfrunza13:
I am interested to know how maintainers envision this feature being added. Is this idea of having another line in the yaml a good idea? Should the ephemeral be an option for things other than the regular message? Should both regular messages and ephemerals be a possibility for the same bot, not just one or the other?

FabioRosado (approved):
Hello @sfrunza13  thank you for working on this, apologies for the delay in reviewing it must have slipped by us.

Regarding your question, I think it would be good to add a configuration to the yaml file, but also allowing users to generate a Message with the ephemeral type so if a user triggers a specific skill, the bot could reply directly to that particular user.

So allowing both would be good an give more flexibility

ryanm101:
I actually have a use case for this..

User wants to query a DB
Bot executes query sent by user 
Bot gets DB results and encrypts them into a file
Bot responds to public chat with confirmation results as attachment
Bot sends key/passphrase to user via ephemeral chat (i know i could use private chat as well) but all comms on a single channel is nice. and I dont really want users to interact with BOT in private chat for audibility.

jacobtomlinson:
I'm going through and closing out stale PRs here. Sorry we never got this over the line.

Issue #1958, opened 2022-12-05
conversations.list getting caught with ratelimited error in v0.27/v0.28 (v0.26 is fine)


rodrigo-albuquerque:
# Description
Started opsdroid:
```shell
opsdroid_test@opsdroid_test-6ccd79897-7tds7:~$ opsdroid start -f configuration.yaml
Opsdroid has a new configuration format since version 0.17.0. Please read on how to migrate in the documentation at https://docs.opsdroid.dev/en/stable/configuration.html#migrate-to-new-configuration-layout.
[12/05/22 10:28:06] INFO     ========================================                                                                                                                                                                           logging.py:161
                    INFO     Started opsdroid v0.28.0.                                                                                                                                                                                          logging.py:162
                    WARNING  'welcome-message: true/false' is missing in configuration.yaml                                                                                                                                                       utils.py:152
                    WARNING  No databases in configuration. This will cause skills which store things in memory to lose data when opsdroid is restarted.                                                                                         loader.py:359
                    INFO     Opsdroid is now running, press ctrl+c to exit.                                                                                                                                                                        core.py:172
                    INFO     Connecting to Slack.                                                                                                                                                                                             connector.py:101
[12/05/22 10:28:07] INFO     Updating Channels from Slack API at Mon Dec  5 10:28:07 2022.                                                                                                                                                    connector.py:203
                    INFO     A new session (s_8745201999626) has been established                                                                                                                                                              __init__.py:360
                    INFO     Connected successfully with socket mode                                                                                                                                                                          connector.py:134
                    INFO     started web server on http://0.0.0.0:8080
```

Then, a couple of seconds later `conversations.list` gets caught by Slack API's `ratelimited` error :
```shell
 File "/home/opsdroid_test/.local/lib/python3.10/site-packages/opsdroid/connector/slack/connector.py", line 208, in _get_channels
    channels = await self.slack_web_client.conversations_list(
  File "/home/opsdroid_test/.local/lib/python3.10/site-packages/slack_sdk/web/async_client.py", line 2463, in conversations_list
    return await self.api_call("conversations.list", http_verb="GET", params=kwargs)
  File "/home/opsdroid_test/.local/lib/python3.10/site-packages/slack_sdk/web/async_base_client.py", line 161, in api_call
    return await self._send(
  File "/home/opsdroid_test/.local/lib/python3.10/site-packages/slack_sdk/web/async_base_client.py", line 201, in _send
[...]

Issue #1905, opened 2022-04-26
Event responses from a button end up outside of a thread even though start-thread = true


rodrigo-albuquerque:
# Description

When start-thread: true in configuration.yaml, replies using regular text or Blocks() construct end up inside of the thread, which is expected.

However, if the Blocks() construct has a button, the reply ends up outside of the thread in a new message in the channel.

# Steps to Reproduce:

0) set "start-thread: true" in configuration.yaml

1) Create a function that responds to a "hello" message using Blocks() with a button within a thread:

```python
class HelloCatcher(Skill):
    @match_regex(r'hello')
    async def my_first_reply(self, message):
        answer = \
            {
                "type": "section",
                "text": {
                    "type": "mrkdwn",
                    "text": f"Click me!"
                },
                "accessory": {
                    "type": "button",
                    "text": {
                        "type": "plain_text",
                        "text": "New reply",
                        "emoji": True
                    },
                    "style": "primary",
                    "value": "create",
                    "action_id": "new",
                },
            } 
        await message.respond(Blocks(answer))
```

2) Create a reply for when we click on above button

```python
@match_event(BlockActions, value="create")
    async def my_reply(self, event):
        await event.respond('This will unexpectedly appear outside of the thread.')
```

This is not expected. 

The expected behaviour is for the reply in 2) to be within the same thread.

# Related 

Previously, an issue where Opsdroid wouldn't reply in a thread when using blocks was fixed (for reference).

Previous issue raised: https://github.com/opsdroid/opsdroid/issues/1896
Previous fix: https://github.com/opsdroid/opsdroid/commit/33241f2d74940e061fada8eee8d8238230593295

rodrigo-albuquerque:
Hi Fabio, quick question. Is this somehow related to the other issue where Block replies ended up outside of a thread?

FabioRosado:
> Hi Fabio, quick question. Is this somehow related to the other issue where Block replies ended up outside of a thread?

It is somewhat related to that yet, but we are doing some 'odd' things with the event from the button click and when I had a quick look, I couldn't figure out what was happening there 😅

PR #1961, opened 2022-12-06
Added retry timer for Slack API rate limit threshold + ability to reply to other bots & within a thread using Blocks

rodrigo-albuquerque:
# Description

Issue 1  reported here: https://github.com/opsdroid/opsdroid/issues/1958
Issue 2 reported here: https://github.com/opsdroid/opsdroid/issues/1905

Issue 1:
The [_get_channels](https://github.com/opsdroid/opsdroid/blob/73da764ca72581a506f05117ec9690ca34b15400/opsdroid/connector/slack/connector.py#L197) method from Slack connector triggers a SlackAPIError exception when the rate limit threshold is reached in Slack API and throws an exception and Opsdroid stops its execution:

Here's the traceback thrown by Opsdroid currently:
```python
 File "/home/opsdroid_test/.local/lib/python3.10/site-packages/opsdroid/connector/slack/connector.py", line 208, in _get_channels
    channels = await self.slack_web_client.conversations_list(
  File "/home/opsdroid_test/.local/lib/python3.10/site-packages/slack_sdk/web/async_client.py", line 2463, in conversations_list
    return await self.api_call("conversations.list", http_verb="GET", params=kwargs)
  File "/home/opsdroid_test/.local/lib/python3.10/site-packages/slack_sdk/web/async_base_client.py", line 161, in api_call
    return await self._send(
  File "/home/opsdroid_test/.local/lib/python3.10/site-packages/slack_sdk/web/async_base_client.py", line 201, in _send
    return AsyncSlackResponse(**{**data, **res}).validate()
  File "/home/opsdroid_test/.local/lib/python3.10/site-packages/slack_sdk/web/async_slack_response.py", line 192, in validate
    raise e.SlackApiError(message=msg, response=self)
slack_sdk.errors.SlackApiError: The request to the Slack API failed. (url: https://www.slack.com/api/conversations.list)
The server responded with: {'ok': False, 'error': 'ratelimited'}
```

The solution for issue 1: The `ratelimited` error from Slack API gives us a retry timer back where instead of raising the exception and aborting the execution we can just back-off respecting the retry-timer which is exactly what I did.

Issue 2: When Opsdroid is replying to an InteractiveAction (within Blocks), it replies outside of a thread.

The solution for issue 2: I just added a small conditional to the payload that is sent back to the channel. If 'thread_ts' exists then Opsdroid should reply within the thread.

## Type of change

- Bug fix (for both issues)

I have caught the SlackAPIError exception, pulled the retry-timer from the error message and added a sleep function to wait for the specified amount of seconds advised by Slack API in the error message and continue afterwards.

# How Has This Been Tested?

After the code change, I left Opsdroid running for over an hour or so with no issues.

Here's a sample of the logs I see after the fix:

```shell
[12/06/22 22:37:00] INFO     Updating Channels from Slack API at Tue Dec  6 22:37:00 2022.                                                                                                                                                    connector.py:203
[12/06/22 22:37:15] WARNING  Rate limit threshold reached. Retrying after 30.0 seconds.                                                                                                                                                    connector.py:220
[12/06/22 22:37:49] WARNING  Rate limit threshold reached. Retrying after 30.0 seconds.                                                                                                                                                    connector.py:220
[12/06/22 22:38:25] WARNING  Rate limit threshold reached. Retrying after 30.0 seconds.                                                                                                                                                    connector.py:220
[12/06/22 22:38:56] WARNING  Rate limit threshold reached. Retrying after 30.0 seconds.                                                                                                                                                    connector.py:220
[12/06/22 22:39:28] WARNING  Rate limit threshold reached. Retrying after 30.0 seconds.                                                                                                                                                    connector.py:220
[12/06/22 22:40:00] WARNING  Rate limit threshold reached. Retrying after 30.0 seconds.                                                                                                                                                    connector.py:220
[12/06/22 22:40:35] WARNING  Rate limit threshold reached. Retrying after 30.0 seconds.                                                                                                                                                    connector.py:220
[12/06/22 22:41:08] WARNING  Rate limit threshold reached. Retrying after 30.0 seconds.                                                                                                                                                    connector.py:220
[12/06/22 22:41:38] INFO     Grabbed a total of 14427 channels from Slack
```

For issue 2: 

I tested by creating a hello world skill with a button for us to click on it so Opsdroid can respond:

```python
from opsdroid.matchers import match_event
from opsdroid.connector.slack.events import Blocks
from opsdroid.connector.slack.events import BlockActions

class HelloCatcher(Skill):
    @match_regex(r'hello man')
    async def my_first_reply(self, message):
        options = [
                {
                    "type": "section",
                    "text": {
                        "type": "mrkdwn",
                        "text": f"Please select your option:"
                    },
                },
            ]
        options.extend([
                    {
                        "type": "section",
                        "text": {
                            "type": "mrkdwn",
                            "text": f"Hey Rodrigo"
                        },
                        "accessory": {
                            "type": "button",
                            "text": {
                                "type": "plain_text",
                                "text": "Click here",
                                "emoji": True
                            },
                            "value": "test",
                            "action_id": "hoho"
                        },
                    },
                    {
                        "type": "divider"
                    },
                ])
        await message.respond(Blocks(options))

    @match_event(BlockActions, value="test")
    async def my_reply(self, event):
        await event.respond("Here's my reply.")
```

Without the fix, Opsdroid responds outside of a thread even when start_thread is set to False. 

With the fix added here, Opsdroid responds within a thread only when start_thread is set to False. 

Previously, it would take a couple of seconds to stop working.

### DIFF
--- a/opsdroid/connector/slack/connector.py
+++ b/opsdroid/connector/slack/connector.py
@@ -217,7 +217,7 @@ class ConnectorSlack(Connector):
                 except SlackApiError as error:
                     if "ratelimited" in str(error):
                         wait_time = float(error.response.headers["Retry-After"])
-                        _LOGGER.warning(_(f"Rate limiting threshold reached. Retrying after {wait_time} seconds."))
+                        _LOGGER.warning(_(f"Rate limit threshold reached. Retrying after {wait_time} seconds."))
                         await asyncio.sleep(wait_time)
                         continue
                     else:
--- a/opsdroid/connector/slack/connector.py
+++ b/opsdroid/connector/slack/connector.py
@@ -211,9 +211,11 @@ class ConnectorSlack(Connector):
                     )
                     self.known_channels.update({c["name"]: c for c in channels["channels"]})
                     cursor = channels["response_metadata"].get("next_cursor")
-
                     if not cursor:
                         break
+                    channel_count = len(self.known_channels.keys())
+                    _LOGGER.info("Grabbed a total of %s channels from Slack", channel_count)
+                    await asyncio.sleep(self.refresh_interval - arrow.now().time().second)
                 except SlackApiError as error:
                     if "ratelimited" in str(error):
                         wait_time = float(error.response.headers["Retry-After"])
@@ -222,10 +224,6 @@ class ConnectorSlack(Connector):
                         continue
                     else:
                         raise
-
-            channel_count = len(self.known_channels.keys())
-            _LOGGER.info("Grabbed a total of %s channels from Slack", channel_count)
-            await asyncio.sleep(self.refresh_interval - arrow.now().time().second)
 
     async def event_handler(self, payload):
         """Handle different payload types and parse the resulting events"""
--- a/opsdroid/connector/slack/connector.py
+++ b/opsdroid/connector/slack/connector.py
@@ -82,6 +82,7 @@ class ConnectorSlack(Connector):
         self.user_info = None
         self.bot_id = None
         self.known_users = {}
+        self.known_bots = {}
         self.known_channels = {}
 
         self._event_creator = SlackEventCreator(self)
@@ -406,19 +407,28 @@ class ConnectorSlack(Connector):
 
         return messages
 
-    async def lookup_username(self, userid):
+    async def lookup_username(self, userid, is_bot=False):
         """Lookup a username and cache it."""
-
         if userid in self.known_users:
-            user_info = self.known_users[userid]
+            return self.known_users[userid]
+        elif userid in self.known_bots:
+            return self.known_bots[userid]
         else:
-            response = await self.slack_web_client.users_info(user=userid)
-            user_info = response.data["user"]
-
-            if isinstance(user_info, dict):
-                self.known_users[userid] = user_info
-
-        return user_info
+            response = (
+                await self.slack_web_client.users_info(user=userid)
+                if not is_bot
+                else await self.slack_web_client.bots_info(bot=userid)
+            )
+            if "user" in response.data:
+                user_info = response.data["user"]
+                if isinstance(user_info, dict):
+                    self.known_users[userid] = user_info
+                return user_info
+            elif "bot" in response.data:
+                bot_info = response.data["bot"]
+                if isinstance(bot_info, dict):
+                    self.known_bots[userid] = bot_info
+                return bot_info
 
     async def replace_usernames(self, message):
         """Replace User ID with username in message text."""
--- a/opsdroid/connector/slack/connector.py
+++ b/opsdroid/connector/slack/connector.py
@@ -205,14 +205,23 @@ class ConnectorSlack(Connector):
             cursor = None
 
             while True:
-                channels = await self.slack_web_client.conversations_list(
-                    cursor=cursor, limit=self.channel_limit
-                )
-                self.known_channels.update({c["name"]: c for c in channels["channels"]})
-                cursor = channels["response_metadata"].get("next_cursor")
+                try:
+                    channels = await self.slack_web_client.conversations_list(
+                        cursor=cursor, limit=self.channel_limit
+                    )
+                    self.known_channels.update({c["name"]: c for c in channels["channels"]})
+                    cursor = channels["response_metadata"].get("next_cursor")
 
-                if not cursor:
-                    break
+                    if not cursor:
+                        break
+                except SlackApiError as error:
+                    if "ratelimited" in str(error):
+                        wait_time = float(error.response.headers["Retry-After"])
+                        _LOGGER.warning(_(f"Rate limiting threshold reached. Retrying after {wait_time} seconds."))
+                        await asyncio.sleep(wait_time)
+                        continue
+                    else:
+                        raise
 
             channel_count = len(self.known_channels.keys())
             _LOGGER.info("Grabbed a total of %s channels from Slack", channel_count)
@@ -245,7 +254,7 @@ class ConnectorSlack(Connector):
             )
 
     async def socket_event_handler(
-        self, client: SocketModeClient, req: SocketModeRequest
+            self, client: SocketModeClient, req: SocketModeRequest
     ):
         payload = {}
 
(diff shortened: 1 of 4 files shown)


opsdroid/connector/slack/connector.py
```diff
@@ -205,14 +205,23 @@ async def _get_channels(self):
...
+                try:
+                    channels = await self.slack_web_client.conversations_list(
+                        cursor=cursor, limit=self.channel_limit
+                    )
+                    self.known_channels.update({c["name"]: c for c in channels["channels"]})
+                    cursor = channels["response_metadata"].get("next_cursor")
+
+                    if not cursor:
+                        break
+                except SlackApiError as error:
```
FabioRosado, line 217: Just a note, we will need to test this change. I was looking at the existing tests and it seems that we didn't do a great job at testing this particular part before your changes 😞 

I haven't tested this, but I **think** you may be able to test the exception by doing something like:

```python

@pytest.mark.asyncio
@pytest.mark.add_response(SlackApiError(...))
async def test_search_history_rate_limit(
    connector, mock_api, caplog
):
    caplog.set_level(logging.INFO)
    await connector.search_history_messages(
        "C01N639ECTY", "1512085930.000000", "1512085980.000000", 1001
    )
    assert mock_api.called("/conversations.history")
    assert "Rate limit threshold reached." in caplog.text
```

You may need to pass things to the SlackApiError, I haven't looked into what they expect. But we can pair on this if you get stuck 😄

opsdroid/connector/slack/connector.py
```diff
@@ -205,14 +205,23 @@ async def _get_channels(self):
...
+                    if "ratelimited" in str(error):
+                        wait_time = float(error.response.headers["Retry-After"])
+                        _LOGGER.warning(_(f"Rate limit threshold reached. Retrying after {wait_time} seconds."))
+                        await asyncio.sleep(wait_time)
+                        continue
+                    else:
+                        raise
```
FabioRosado, line 224: We can use the same snippet above but throw a different exception and that should trigger this line

```python
@pytest.mark.asyncio
@pytest.mark.add_response(Exception)
async def test_search_history_messages_limit_more_than_1000(
    connector, mock_api, caplog
):
    caplog.set_level(logging.INFO)
    with pytest.raises(Exception):
        await connector.search_history_messages(
        "C01N639ECTY", "1512085930.000000", "1512085980.000000", 1001
    )
```

FabioRosado (commented):
Added some comments, it seems you have some linting issues as well.

You can follow [this article in the docs](https://docs.opsdroid.dev/en/stable/contributing/workflow.html#automatic-linting-with-black) to install pre-commit and black and get black to format your code automatically.

Once it's installed you can run `pre-commit run black .` and it will fix the linting issues for you 😄

opsdroid/connector/slack/connector.py
```diff
@@ -205,14 +205,23 @@ async def _get_channels(self):
...
+                        cursor=cursor, limit=self.channel_limit
+                    )
+                    self.known_channels.update({c["name"]: c for c in channels["channels"]})
+                    cursor = channels["response_metadata"].get("next_cursor")
+
+                    if not cursor:
+                        break
+                except SlackApiError as error:
+                    if "ratelimited" in str(error):
+                        wait_time = float(error.response.headers["Retry-After"])
```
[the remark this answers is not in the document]
rodrigo-albuquerque: when the exception is "ralimited" error, there is always a Retry-After header to tell us how long we should wait. Yes, it seems slack API always sends back 30s but the idea is to always respect such header? The other alternative would be to check if there is "Retry-After" in the header first, and if so, honour its value. Otherwise, use a hard-coded value of 30s? What do you think?

opsdroid/connector/slack/connector.py
rodrigo-albuquerque: The API response is this one:
```python
 File "/home/opsdroid_test/.local/lib/python3.10/site-packages/opsdroid/connector/slack/connector.py", line 208, in _get_channels
    channels = await self.slack_web_client.conversations_list(
  File "/home/opsdroid_test/.local/lib/python3.10/site-packages/slack_sdk/web/async_client.py", line 2463, in conversations_list
    return await self.api_call("conversations.list", http_verb="GET", params=kwargs)
  File "/home/opsdroid_test/.local/lib/python3.10/site-packages/slack_sdk/web/async_base_client.py", line 161, in api_call
    return await self._send(
  File "/home/opsdroid_test/.local/lib/python3.10/site-packages/slack_sdk/web/async_base_client.py", line 201, in _send
    return AsyncSlackResponse(**{**data, **res}).validate()
  File "/home/opsdroid_test/.local/lib/python3.10/site-packages/slack_sdk/web/async_slack_response.py", line 192, in validate
    raise e.SlackApiError(message=msg, response=self)
slack_sdk.errors.SlackApiError: The request to the Slack API failed. (url: https://www.slack.com/api/conversations.list)
The server responded with: {'ok': False, 'error': 'ratelimited'}
```

rodrigo-albuquerque: I was also able to print out "str(error)" by adding a print statement:
```python
str(error): 
The request to the Slack API failed. (url: https://www.slack.com/api/conversations.list)
The server responded with: {'ok': False, 'error': 'ratelimited'}
```

opsdroid/connector/slack/connector.py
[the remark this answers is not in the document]
FabioRosado: Yeah I think being defensive here is a good idea. I think we can do something like:

```suggestion
                        wait_time = float(error.response.headers.get("Retry-After", 30)
```

opsdroid/connector/slack/connector.py
FabioRosado: That's great then we can probably just create the `SlackApiError` with that (I'm assuming that the error is a wrapper around the request)

rodrigo-albuquerque:
hm, I think it's just 2 lines not covered by tests.

FabioRosado (approved):
I'm on the phone but it shows that all the CI passed and it's green so I'll merge it thanks! 😄

PR #1973, opened 2022-12-18
CI: fix Tox docker run

oleg-fiksel:
# Description

Seem that Tox has a problem running `docker run` directly form the commands section. Unfortunately I didn't find the reason why. So I replaced it with a `bash -c ..` command.
Also I reworked the section after the `docker run`. Now the logic would wait until container is running instead. This will compensate if the container starts longer than 10 seconds.

## Status
**READY** 

## Type of change

# How Has This Been Tested?

- CI pipline is 🟢 

# Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,53 +7,6 @@ on:
     - cron: "0 0 * * *"
 
 jobs:
-  test:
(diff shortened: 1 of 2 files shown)


FabioRosado (approved):
Awesome work!

PR #1980, opened 2023-01-22
Fix configuration sample entry

pwillis-eiq:
# Description

The documentation above (and the code https://github.com/opsdroid/opsdroid/blob/d7c754b285cf01d607fb5f0cf4456f51740c44cb/opsdroid/loader.py#L566-L572) say to use `no-dep`, but the example line shows `no-deps`.

Fixes example to say `no-dep` instead of `no-deps`

## Status
**READY** 

## Type of change

# How Has This Been Tested?

Tested using `no-deps` option; still installed deps failed. Switched to `no-dep`: no deps installed.

# Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -590,7 +590,7 @@ Set `no-dep` to true to skip the installation of dependencies on every start of 
 skills:
   myawesomeskill:
     no-cache: true
-    no-deps: true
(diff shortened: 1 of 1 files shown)


FabioRosado (approved):
Hello 👋 
Thank you so much for this fix

PR #1983, opened 2023-02-05
Report unhandled exceptions as warnings

xloem:
# Description

I was debugging some test issues and discovered there are a number of exceptions being silently absorbed when pytest runs. It looks like some of these exceptions are interrupting expected cleanups and socket closures here and there.

This change reports them as warnings, so pytest will summarize them.

## Status
**READY** | **ON HOLD**

## Type of change

# How Has This Been Tested?

- I ran pytest and casually observed new warnings output.

# Checklist:


FabioRosado (approved):
Awesome find thank you for opening this PR! 😄

PR #1984, opened 2023-02-23
add timeout on matrix connection and related errors handling

### LINKED ISSUE #1750: If the matrix server is unreachable on startup no error is logged
I need to look into this more, but my reverse proxy was offline for my matrix server and opsdroid started, but then didn't log anything at any level (my logs were on debug) to tell me it couldn't connect.

flefebvre03:
# Description

Fixes #1750

## Status
**READY**

## Type of change

# How Has This Been Tested?

- Attempted connection to non-existent matrix server, checked that an error is logged, and that no exception are raised

# Checklist:

### DIFF
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -222,7 +222,7 @@ class ConnectorMatrix(Connector):
                 login_response = await self.connection.login(
                     password=self.password, device_name=self.device_name
                 )
-            except:
+            except Exception:
                 _LOGGER.error("connexion error")
                 self.connection_failed = True
                 return
@@ -332,9 +332,7 @@ class ConnectorMatrix(Connector):
 
     async def listen(self):  # pragma: no cover
         """Listen for new messages from the chat service."""
-        if self.connection_failed is True:
-            return
-        while True:  # pylint: disable=R1702
+        while not self.connection_failed:  # pylint: disable=R1702
             response = await self.connection.sync(
                 timeout=int(60 * 1e3),  # 1m in ms
                 sync_filter=self.filter_id,
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -193,6 +193,8 @@ class ConnectorMatrix(Connector):
             encryption_enabled=self._allow_encryption,
             pickle_key="",
             store_name="opsdroid.db" if self._allow_encryption else "",
+            request_timeout = 60,
+            max_timeouts = 1,
         )
         self.connection = nio.AsyncClient(
             self.homeserver,
@@ -216,9 +218,14 @@ class ConnectorMatrix(Connector):
             self.connection.user_id = self.mxid
 
         elif self.mxid is not None and self.password is not None:
-            login_response = await self.connection.login(
+            try:
+                login_response = await self.connection.login(
                 password=self.password, device_name=self.device_name
-            )
+                )
+            except:
+                _LOGGER.error ("connexion error")
+                self.connection_failed = True
+                return
             if isinstance(login_response, nio.LoginError):
                 _LOGGER.error(
                     f"Error while connecting: {login_response.message} (status code {login_response.status_code})"
@@ -325,6 +332,8 @@ class ConnectorMatrix(Connector):
 
     async def listen(self):  # pragma: no cover
         """Listen for new messages from the chat service."""
+        if(self.connection_failed is True):
+            return
         while True:  # pylint: disable=R1702
             response = await self.connection.sync(
                 timeout=int(60 * 1e3),  # 1m in ms
--- a/opsdroid/connector/matrix/connector.py
+++ b/opsdroid/connector/matrix/connector.py
@@ -193,8 +193,8 @@ class ConnectorMatrix(Connector):
             encryption_enabled=self._allow_encryption,
             pickle_key="",
             store_name="opsdroid.db" if self._allow_encryption else "",
-            request_timeout = 60,
-            max_timeouts = 1,
+            request_timeout=60,
+            max_timeouts=1,
         )
         self.connection = nio.AsyncClient(
             self.homeserver,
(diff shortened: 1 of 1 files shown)


opsdroid/connector/matrix/connector.py
```diff
@@ -216,9 +218,14 @@ async def connect(self):
...
 
         elif self.mxid is not None and self.password is not None:
-            login_response = await self.connection.login(
-                password=self.password, device_name=self.device_name
-            )
+            try:
+                login_response = await self.connection.login(
+                    password=self.password, device_name=self.device_name
+                )
+            except:
```
FabioRosado, line 225: Could we do something like this instead?

```suggestion
            except Exception:
```

opsdroid/connector/matrix/connector.py
```diff
@@ -325,6 +332,8 @@ async def _parse_sync_response(self, response):
 
     async def listen(self):  # pragma: no cover
         """Listen for new messages from the chat service."""
+        if self.connection_failed is True:
+            return
         while True:  # pylint: disable=R1702
```
FabioRosado, line 337: Perhaps instead of having the above check we can do something like this what do you think?

```python
while not self.connection_failed:
```

FabioRosado (approved):
Hello thank you for working on this fix and opening the PR!

I've added some comments but they are non blocking for this PR so I'm approving it

flefebvre03:
I believe it's ready to be merged. 
Is there anything else I have to do?

opsdroid/connector/matrix/connector.py
```diff
@@ -193,6 +193,8 @@ async def connect(self):
             encryption_enabled=self._allow_encryption,
             pickle_key="",
             store_name="opsdroid.db" if self._allow_encryption else "",
+            request_timeout=60,
+            max_timeouts=1,
```
Cadair, line 197: Neither of these new args are used anywhere?

opsdroid/connector/matrix/connector.py
```diff
@@ -216,9 +218,14 @@ async def connect(self):
...
         elif self.mxid is not None and self.password is not None:
-            login_response = await self.connection.login(
-                password=self.password, device_name=self.device_name
-            )
+            try:
+                login_response = await self.connection.login(
+                    password=self.password, device_name=self.device_name
+                )
+            except Exception:
+                _LOGGER.error("connexion error")
```
Cadair, line 226:
```suggestion
                _LOGGER.error("Logging into matrix homeserver failed.")
```

Cadair (changes requested):
I think this is incomplete, you don't seem to use the timeout or the retry?

opsdroid/connector/matrix/connector.py
eguegain: These arguments are given to the AsyncClientConfig from nio. Developers do not have to "use" it, as it is used directly by nio. See the [documentation](https://matrix-nio.readthedocs.io/en/latest/_modules/nio/client/async_client.html?highlight=AsyncClientConfig#), ctrl+F max_timeouts. 
request_timeout=60 is unnecessary as it is the default value, but max_timeouts prevents infinite retries.

Cadair: Oh I am sorry, I read this as the constructor to the class.

I think it would be good to add these to the connector config, so people can customise them?

opsdroid/connector/matrix/connector.py
Cadair: I don't think the error case is handled properly here. If a request nio makes fails, it does not raise an excepiton it returns an [error class](https://github.com/poljar/matrix-nio/blob/main/nio/client/async_client.py#L960).

This also points out that there should be a test for this.

opsdroid/connector/matrix/connector.py
Cadair, line 197: I am not sure we want to have `max_timeouts` as 1. Quite often opsdroid is running somewhere out of sight, having it try to login once, throw an error and give up seems... unhelpful.

I think the main issue with the original issue I opened, was the lack of feedback about login failing. As long as we log what's happening I think we could keep the default number of retries as infinitiy.

opsdroid/connector/matrix/connector.py
Cadair, line 226: I think what this needs to be is something like this:
```suggestion
            
            login_response = await self.connection.login(
                password=self.password, device_name=self.device_name
            )
            if isinstance(login_response, LoginError):
                logger.error("Login to the matrix homeserver failed with error: %s", login_response)
```

opsdroid/connector/matrix/connector.py
flefebvre03: We can set an higher value for max_timeouts but we don't have any feedback about the number of attempts.
We have to wait for max_timeouts to be reached before we can log anything.
An other approach could be to set max_timeouts = 1 and manually retry when the exception is raised. 
But I 'm not sure that this is the way Opsdroid usually works.

opsdroid/connector/matrix/connector.py
Cadair: Good to know on the socket error. However, I think that we probably *also* want to catch non-socket errors at login which would be returned as a `LoginError` class? Perhaps a hybrid of both is the answer?

opsdroid/connector/matrix/connector.py
Cadair: > We have to wait for max_timeouts to be reached before we can log anything.

ah, ok that's interesting.

> An other approach could be to set max_timeouts = 1 and manually retry when the exception is raised.

This feels like a good answer, but I agree, I am not sure how the other connectors handle connection errors. It seems like [slack just errors](https://github.com/opsdroid/opsdroid/blob/master/opsdroid/connector/slack/connector.py#L115) and the core has no retry functionality. So perhaps we should just follow them and keep this as it is.

Cadair: given the comments below about how `max_timeouts` work, perhaps just having `request_timeout` be configurable?

opsdroid/connector/matrix/connector.py
flefebvre03: In fact, it's what we do:
We surrounded the login() method with a try / except, but we kept the if isinstance(login_response, LoginError) that was initially the code.

opsdroid/connector/matrix/connector.py
flefebvre03: I believe this kind of parameters should be shared between connectors, instead of a per-connector basis.
It is probably beyond the scope of this pull request.

jacobtomlinson:
Sorry we never got this over the line. Given the age of this PR I'm going to close it out.

PR #1986, opened 2023-03-04
Raise an exception in matrix event creator if request fails

Cadair:
# Description

I noticed an unreproducible error in the matrix event creator, I am going to guess it was some kinda server error. This PR just makes it so we raise an exception if this call fails.

## Status
**READY**

## Type of change

# How Has This Been Tested?

It hasn't I can't reproduce the error 🙈 

# Checklist:


jacobtomlinson (approved):
Gah this had green CI too! Sorry that we never got this merged. I've updated the branch so hopefully things are good to go now.

Issue #983, opened 2019-06-07
Add arm64 docker images


octagonprogramming:
# Description
Please include a summary of the issue. 
```
When trying to run ops in docker this error message is generated:
          standard_init_linux.go:211: exec user process caused "exec format error"
 
Not sure if this is an error on my end or what
```

## Steps to Reproduce
Please also include relevant information and steps to reproduce the bug/issue.
### Requirments
```
Be running docker on ubuntu server
     -specifically running ubuntu server on raspberry pi
```

```
# Pull the container image
$ docker pull opsdroid/opsdroid:latest

# Run the container
$ docker run --rm -it -v /path/to/config_folder:/root/.config/opsdroid opsdroid/opsdroid:latest
```

## Expected Functionality
Explain what should happen.
```
A container running opsdroid should be created and run
```

## Experienced Functionality
Explain what happened instead(Please include the debug log).
```
$ docker run --rm -it -v /path/to/config_folder:/root/.config/opsdroid opsdroid/opsdroid:latest
standard_init_linux.go:211: exec user process caused "exec format error"
```
## Versions
- **Opsdroid version:** 
```
image: opsdroid/opsdroid:latest
```

- **Python version:** 
```
3.6.7
```

- **OS/Docker version:**
```
Linux version 4.15.0-1036-raspi2 (buildd@bos02-arm64-045) (gcc version 7.3.0 (Ubuntu/Linaro 7.3.0-16ubuntu3)) #38-Ubuntu SMP PREEMPT Wed May 8 09:46:19 UTC 2019
```

Docker version 18.09.5, build e8ff056

## Configuration File
Please include your version of the configuration file bellow.

```yaml
# Your code goes here.

```

## Additional Details
```
Any other details you wish to include such as screenshots, console messages, etc.
A temporary solution for this was I created a container from the image ubuntu:latest and installed opsdroid there and it works so far but haven't done any configuration.
```

octagonprogramming:
Update: I cloned the repository and rebuild the container and ran it with the provided command in the README (with a different image name) and the container ran without the same error.

octagonprogramming:
On further investigation, the opsdroid image from the official repository is for an amd64 system so it does not work on my system which is arm64.

jacobtomlinson:
Thanks for raising this. We are currently only building images for x86/64 architectures. 

It would be great to add support for ARM processors, we should look into how we can do this.

octagonprogramming:
I would love to work on this!

[...]

PR #1988, opened 2023-04-20
Adding ARM build suppport

ryanm101:
Adding ARM build support is really slow but that is due to QEMU as its not native ARM builds.

One improvement might be to create an ARM dedicated workflow, but not sure how that would impact pushes to the same tag. 

Fixes: https://github.com/opsdroid/opsdroid/issues/983


jacobtomlinson (approved):
I'm so sorry this took so long to review. Merging!

PR #1989, opened 2023-04-22
Fix the currently broken facebook connector

plantroon:
# Description

I wanted to use the facebook connector and found that it was broken. Sending a message from facebook side would result in opsdroid receiving it but then proceeding to use user_id as message's text, due to how the arguments were positioned.

I bumped the API version as well to the newest one.

I did not report an issue prior to this.

## Status
**READY**

## Type of change

# How Has This Been Tested?

Simply by following instructions on setting up the facebook connector in documentation. I tested it with multiple skills and all work fine. Hello.py skill would however use the facebook's user_id instead of his actual name, but that is how this connector must have worked originally, as there doesn't seem to be a straight-forward way to get this name. If this is a problem I can look around the facebook API docs and implement it, though barely any skills use it.

# Checklist:


jacobtomlinson (approved):
I'm so sorry it took so long to get around to reviewing this. Things have been slow here lately. 

Overall I think this looks great. I pushed a commit to fix some linting issues, but otherwise I think this should be good to merge.

Issue #1899, opened 2022-04-12
Matrix connector: redact (remove) messages or events


oleg-fiksel:
# Description

Opsdroid Matrix connector doesn't support redacting/removing/deleting events/messages.

## Steps to Reproduce

* [ ] TODO

## Expected Functionality

* [ ] TODO

## Experienced Functionality

* [ ] TODO

## Versions
- **Opsdroid version: v0.25.0**

lucastronova:
I have a PR open addressing this (#1993)

oleg-fiksel:
Thanks @lucastronova !

Unfortunately I'm AFK for at least 3 weeks, so will not be able to review your PR soon.

Maybe someone from the maintainers will have time to take a look.

PR #1993, opened 2023-05-23
Allow removal of Matrix messages

lucastronova:
# Description

Added remove method to the Message event class, and function in the Matrix connector.py using this method.
Example use:

    class UnpongSkill(Skill):

    @match_regex(r"pong")
    async def unpong(self, event):
        await event.remove()

is a skill that automatically redacts any 'pong' message sent in the room.
For this to work the bot must have the privileges to remove the requested message.

Addresses #1899

## Status
**READY** 

## Type of change

- This change requires a documentation update(?)

# How Has This Been Tested?

- manually tested in private Matrix room
- ran tox; only tests that failed were unrelated to changes.

# Checklist:

### DIFF
--- a/opsdroid/events.py
+++ b/opsdroid/events.py
@@ -284,6 +284,9 @@ class Message(Event):
 
         return await super().respond(response)
 
+    async def remove(self, reason):
+        return await super().respond(DeleteMessage(reason, linked_event=self))
+
 
 class EditedMessage(Message):
     """A  `opsdroid.events.Message` which has been edited.
@@ -546,6 +549,10 @@ class UnpinMessage(Event):
 
 class DeleteMessage(Event):
     """Event to represent deleting a message or other event."""
+
+    def __init__(self, reason, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.reason = reason
 
 
 class BanUser(Event):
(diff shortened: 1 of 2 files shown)


opsdroid/events.py
```diff
@@ -284,6 +284,9 @@ async def respond(self, response_event):
 
         return await super().respond(response)
 
+    async def remove(self, reason=None):
+        return await super().respond(DeleteMessage(reason, linked_event=self))
+
```
Cadair, line 289: This is the only bit of this PR that I am unsure on. I don't know if we want to add this helper method or not.

We don't have a helper method for `react` or any other specific type of response. I shall ask the other maintainers what their thoughts are.

opsdroid/events.py
```diff
@@ -284,6 +284,9 @@ async def respond(self, response_event):
 
         return await super().respond(response)
 
+    async def remove(self, reason=None):
+        return await super().respond(DeleteMessage(reason, linked_event=self))
```
Cadair, line 288:
```suggestion
        return await self.respond(DeleteMessage(reason, linked_event=self))
```

Sleuth56:
I can confirm that this code as written does delete matrix messages when it has the permission to do so. When it does not have permission to remove messages the following exception is thrown and opsdroid keeps running. Looks good to me. `opsdroid.connector.matrix.connector.MatrixException: RoomRedactError: M_FORBIDDEN You don't have permission to redact events`

PR #2015, opened 2023-11-29
Fix mongo database error

st01cs:
# Description

response = await self.database[self.collection].find_one() 

[response] will be None when cant find document in collection, response.key() will throw exception. 

## Status
**READY** 

## Type of change

# How Has This Been Tested?

Find an document which does not exist in collection, Opsdroid will not be broken.

# Checklist:

- [ *] I have performed a self-review of my own code
- [ *] I have made corresponding changes to the documentation (if applicable)
- [ *] I have added tests that prove my fix is effective or that my feature works
- [ *] New and existing unit tests pass locally with my changes


jacobtomlinson (approved):
Sorry for the delay in reviewing here. Thanks!

PR #2021, opened 2024-03-27
EditedMessage should use 'username' as 'user' is not present in slack API response

rodrigo-albuquerque:
# Description

When a bot such as Opsdroid itself edits a slack message, then [user_id=event["message"]["user"] ](https://github.com/opsdroid/opsdroid/blob/e2ffb0ce1d859d7962de3af14cd83c36c45f2850/opsdroid/connector/slack/create_events.py#L103) throws an exception as Slack API returns `["message"]["username"]` rather than `["message"]["user"]` :

```python
/Users/rodrigo/github.com/opsdroid_test/.venv/lib/python3.11/site-packages/opsdroid/connector/slack/create_events.py:103 in edit_message                         
                             │                    
                             │                                                                                                                                                                  │                    
                             │   100 │   │   return events.EditedMessage(                                                                                                                       │                    
                             │   101 │   │   │   text,                                                                                                                                          │                    
                             │   102 │   │   │   user=user_name,                                                                                                                                │                    
                             │ ❱ 103 │   │   │   user_id=event["message"]["user"],                                                                                                              │                    
                             │   104 │   │   │   target=event["channel"],                                                                                                                       │                    
                             │   105 │   │   │   connector=self.connector,                                                                                                                      │                    
                             │   106 │   │   │   event_id=event["ts"],                                                                                                                          │                    
                             ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯                    
                             KeyError: 'user'
```

We should also take into account that Opsdroid may edit the message and not throw the above exception.

## Status
**READY**

## Type of change

# How Has This Been Tested?

I added `print(f'event[message]: {event["message"]}')` in [here](https://github.com/opsdroid/opsdroid/blob/e2ffb0ce1d859d7962de3af14cd83c36c45f2850/opsdroid/connector/slack/create_events.py#L99) so we can print the Slack API response just before we edited the message.

Then, in my test skill I added this:

```python
updated_message_blocks = [
            {
                "type": "section",
                "text": {
                    "type": "mrkdwn",
                    "text": f"Editing an existing block to this."
                }
            },
        ]

        # Create an EditedBlocks event
        edited_blocks_event = EditedBlocks(
            blocks=updated_message_blocks,
            user=BOT_ID,
            target=channel_id,
            connector=self.opsdroid.get_connector("slack"),
            linked_event=message_ts
        )

        # Send the EditedBlocks event to Opsdroid for processing
        await event.respond(edited_blocks_event)
```

The result of the above code being executed is seen here:

```python
event[message]: {'subtype': 'bot_message', 'text': 'Editing an existing block to this.', 'username': 'opsdroiddev1', 'icons': {'emoji': ':smile:'}, 'type': 'message', 'edited': {'user': 'B15006Y7Y5R', 'ts': '1711390208.000000'}, 'bot_id': 'B15006Y7Y5R', 'app_id': 'A42US8BQPBR', 'thread_ts': '1211390201.014429', 'parent_user_id': 'U42V150AH8B', 'blocks': [{'type': 'section', 'block_id': 'kdKvc', 'text': {'type': 'mrkdwn', 'text': 'Editing an existing block to this.', 'verbatim': False}}], 'ts': '1711390203.443818', 'source_team': 'EACV1ZA7J', 'user_team': 'EACV1ZA7J'}
```

Note that Slack API returns `["message"]["username"]` rather than`["message"]["user"]`, so when this block was executed without my changes I got the exception displayed in the Description section of this PR. When I changed it to `["message"]["username"]` then the Exception was gone because of course it matched Slack API's response.

# Checklist:


jacobtomlinson (approved):
Thanks!

PR #1999, opened 2023-06-22
Added 3.11 job to tox and github actions config 

Alex-Wengg:
# Description
First iteration to help migrate python3.8 -> latest python3.11

PR #2028, opened 2024-04-25
Add Python 3.11 to CI

jacobtomlinson:
Sorry that we didn't review #2001 @Maverick073. Picking things up here to try and get it merged now that CI is green.


jacobtomlinson:
Hmm there is more to this as `asynctest` does not support Python 3.11 and needs removing.

jacobtomlinson:
This is already partially done in #1999 so let's stick with that PR.

PR #2050, opened 2025-02-18
Matrix Connector Returns List of RoomIDs Method

### LINKED ISSUE #1732: Add a method on the matrix connector which lists all rooms the account is joined to

BradenAzis:
# Description

Fixes #1732 

No new dependencies necessary. SImply uses the joined_rooms() member of ConnectorMatrix.connection, which is an nio.AsyncClient object.  Also fetches error./failure to fetch rooms.

## Status
**READY**

## Type of change

# How Has This Been Tested?

Tests written in opsdroid/connector/matrix/tests/test_connector_connect.py. Three separate test cases, one where it handles proper use, a failure to retrieve, and unexpected input/behavior. These tests use a Mock API call and assertion statements that the function returns the proper list of room IDs.

# Checklist:

- I have performed a self-review of my own code
- I have added tests that prove my fix is effective or that my feature works
- New and existing unit tests pass locally with my changes


jacobtomlinson (changes requested):
Thanks @BradenAzis. We just merged #2054 which gets the CI back in a good state, but it also bumps the versions of a lot of things, so this PR needs updating to fix the failures.

Also it looks like you checked in `opsdroid-env` with your python environment in, can you remove that again?

PR #2064, opened 2025-04-29
Added proper filters for websocket events in Mattermost connector

### LINKED ISSUE #2063: Mattermost messages should be properly filtered before they are even pared by OpsDroid
# Description
Mattermost send the full event stream to OpsDroid, which in turn will parse all events it can handle. At the skill level we have written a custom constraint that will then filter out the following events:

1. Emoji Reactions where the emoji was not the first; OpsDroid should only parse on the first reaction of the configured emoji. If people later add the same reaction, OpsDroid should not spam additional responses
2. If the channel is not a direct message channel, OpsDroid should ignore all messages except for those where it was specifically tagged to not spam its responses and annoy everyone; this needs to be configurable, maybe someone has a use case in the future where they want all messages to be processed

At that level the messages have already been process by Rasa NLU for example, and thus consumed unnecessary CPU cycles. In addition it's not very intuitive to new users that they have to implement custom constraints to get the most common use cases running.

This issue aims at adding sensible defaults to cover the common use cases out of the box, but allow them to be disabled for advanced users.

## Additional Details
You can assign this issue to me 😃

aboukhal:
# Description

Added proper filters so the bot picks up only messages that were directed to it.
Added unit tests for all the different paths and restructured the file paths for the json responses so it's easier to understand how the responses relate to each other.

Fixes #2063 

## Status
**READY**

## Type of change

# How Has This Been Tested?

Running in our test environment. We removed our constraint decorator and with the new filters, it behaves exactly as before.

# Checklist:


jacobtomlinson (approved):

jacobtomlinson:
Looks like there is a Windows CI failure that needs addressing before this can be merged.

Lichtblitz:
One of the windows builds worked fine. I believe this is because of the flakey test suite. I assume there is an issue with not releasing the port for the mock webserver quickly enough before the next test starts. I'll look into that at some point to make tests more reliable.
Could you just try rerunning the failed tests?

jacobtomlinson:
Sure I've rerun them. However we might need to add `pytest-rerunfailures` to do this automatically because we don't want flaky tests on `main`.

Lichtblitz:
The tests have been flakey since I started work, even before the refresh. I'll check and see if I can fix the issue a bit more fine grained before going with the nuclear option.

aboukhal:
This is super weird. `aiohttp` raises some strange exceptions during teardown on windows+python 3.11 and windows+python 3.12 only. Timing the websocket reconnect less aggressive (1 second wait instead of adding it back to the queue without delay) has seemingly fixed it.

I've also added some debug logs to be able to analyze the elusive flakey tests related to port binding errors in the future.

I've also fixed the broken github actions in forks - those will no longer try and fail to deploy to docker hub - only to ghcr.

It's all in a separate commit for easier review.

.github/workflows/deploy-docker-dev.yml
```diff
@@ -7,7 +7,39 @@ on:
...
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+    - name: Build and push
+      uses: docker/build-push-action@v3
+      with:
+        push: true
+        tags: ${{ steps.docker_meta.outputs.tags }}
+        platforms: linux/amd64
+        labels: ${{ steps.docker_meta.outputs.labels }}
```
jacobtomlinson, line 39: Do we really need forks to publish dev containers? The intent of this action is just to run on `main` when a PR is merged so that folks can pull down the latest tip of trunk changes. I don't feel like this is something forks needs to care about.

aboukhal: I use this actively to pull my work state into our test server before opening a pull request. Docker containers are much more objective for these tests, than running directly from code.

jacobtomlinson: Sure, but this repo has hundreds of forks, which could result in vast amounts of storage being used. This feels like something that is concenient for you but is unnecessarily wasteful.

Perhaps I could suggest adding an extra check to the `if: github.repository!='opsdroid/opsdroid'` that looks for [repo environment variable](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-configuration-variables-for-multiple-workflows). That way you can opt into this on your fork without setting it for everyone.

aboukhal: Great idea, I'll look into that!

aboukhal: I've added a new commit that introduces an action variable and adds documentation for it as well.

jacobtomlinson (approved):
Sweet thanks for jumping on that!

PR #2070, opened 2026-01-16
Added Password Generator Skill

askit10:
# Description

I have implemented a new skill called `PasswordGeneratorSkill` that automatically generates a secure password upon the `OpsdroidStarted` event. 

Additionally, I updated the default configuration to use the `websocket` connector instead of `shell`. This resolves the "RuntimeError: Event loop is closed" issue commonly faced by Windows users when using the shell connector.

Fixes
#Fixes Windows compatibility issues

## Status
**READY**

## Type of change

[x]- Bug fix (non-breaking change which fixes an issue)
[x]- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- Manual testing on Windows 10 environment.
- Verified that running `opsdroid start` launches the bot successfully without crashing and prints the generated password in the console.

# Checklist:

### DIFF
--- a/configuration.yaml
+++ b/configuration.yaml
@@ -1,0 +1,7 @@
+
+connectors:
+  - name: websocket
+
+skills:
+  - name: password
+    path: password_skill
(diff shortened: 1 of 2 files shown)


jacobtomlinson (commented):
Nice work! I don't think this is something we are going to merge but nice to se eyou're giving opsdroid a go. You should look into sending the message back through the bot instead of printing the output.

